PR #1, opened 2020-05-10
Setup remix ide

ioedeveloper:


GrandSchtroumpf:
You didn't create the app with the nx cli, did you ? 
The nx.json file has no reference of the remix-ide, which might leads to some un-optimize operation, especially on affected build/tests

ioedeveloper:
> You didn't create the app with the nx cli, did you ?
@GrandSchtroumpf I migrated remix-ide to the monorepo as it is and tried maintaining git history alongside. This operation cannot be handled by nx cli. We can leverage the affected command when we scaffold a new app for remix-ide (react/theia).

GrandSchtroumpf:
Maybe we could update the nx.json by hand

ioedeveloper:
Yes, i'm working on that now.

PR #5, opened 2020-06-02
Libs setup

### LINKED ISSUE #6: Build libs using @nrwl/node:package
Currently there are some issues with such build, we are moving with an alternative in https://github.com/ethereum/remix-project/pull/5

This issue is opened on nx repo https://github.com/nrwl/nx/issues/3142

Aniket-Engg:
- remove unnecessary directories

fixes #6


workspace.json
```diff
@@ -66,6 +66,166 @@
...
 
       }
+    },
+    "remix-analyzer": {
+      "root": "libs/remix-analyzer",
+      "sourceRoot": "libs/remix-analyzer/",
+      "projectType": "library",
+      "schematics": {},
+      "architect": {
+        "lint": {},
```
ioedeveloper, line 76: Does remix-libs have linting configured?

ioedeveloper:
`nx dep-graph` displays a dependecy graph showing dependencies between different apps and libs. I think it will be good if you added config for libs that depend on each other (that is if we have any). I think the config should go in nx.json

PR #18, opened 2020-07-20
CI updated for libs and ide build

Aniket-Engg:


.circleci/config.yml
```diff
@@ -4,6 +4,29 @@
...
+      # CircleCI maintains a library of pre-built images
+      # documented at https://circleci.com/docs/2.0/circleci-images/
+    resource_class: xlarge
+      # - image: circleci/mongo:3.4.4
+    environment:
+      - COMMIT_AUTHOR_EMAIL: "yann@ethereum.org"
+      - COMMIT_AUTHOR: "Circle CI"
+    working_directory: ~/remix-project
+
+    parallelism: 20
```
ioedeveloper, line 22: This step is not needed since we are not making use of parallelism for running tests.

Issue #30, opened 2020-07-14
Integrate the ability of running provider plugins.


yann300:
A first naive implementation:
 - appManager listen on plugin activated
 - If it is of `provider` plugin, appManager call runtab.addProvider for adding it to the env list.
 - when it's deactivated, entry is removed.

LianaHus:
is there a 'provider' type for plugins?

yann300:
in the branch  I was working yes but it's not quite ready for review (but soon)

PR #178, opened 2020-07-28
Add plugin provider

yann300:
This add the ability to reference new web3 provider.
Each "provider" plugin will be listed in the environment section of the run tab.

fix https://github.com/ethereum/remix-project/issues/30

Current state:

 - [x] with metamask, rejecting a transaction doesn't resolve the call in the client.
 - [x] sometimes `gasEstimate` returns a JSON RPC error (probably when the transaction fails), then force sending the transaction doesn't pop up anything in the mobile side.
   EDIT: https://github.com/ethereum/remix-project/pull/178/files#diff-b41d6048d9047202a90628384dd7d690R181

 - [x] ~~can take more than 10 seconds for a call to resolve.~~
we'll probably need if the problem persist to have our own bridge.

 - [x] ~~call to a plugin can't be done asynchronously, so when a call doesn't resolve, all the other call to web3 are lined up (This probably can't be solved right now).~~ this can't be solved. Plugin calls are synchronous (cc @GrandSchtroumpf fyi)


apps/remix-ide/src/app/tabs/network-module.js
```diff
@@ -59,8 +60,12 @@ export class NetworkModule extends Plugin {
   }
 
   /** Add a custom network to the list of available networks */
-  addNetwork (customNetwork) {
-    this.blockchain.addProvider(customNetwork)
+  addNetwork (network) { // { name, url }
+    if (network.url === 'ipc') {
```
iurimatias, line 64: suggestion:

```
const provider = network.url === 'ipc' ? new Web3.providers.IpcProvider() : new Web3.providers.HttpProvider(network.url)
this.blockchain.addProvider({ name: network.name, provider })
```
```

File: apps/remix-ide/src/app/udapp/run-tab.js [hunk 2]
```diff
@@ -201,6 +201,31 @@ export class RunTab extends LibraryPlugin {
...
+                  callback(null, result)
+                })
+                .catch(e => {
+                  console.log('wallet connect', e)
+                  callback(e)
+                })
+            }
+          }
+          this.blockchain.addProvider({ name: profile.displayName, provider: web3Provider })
+        })(profile, this)
```
iurimatias, line 226: is this really necessary (() => {})() it feels like it could be removed, and use this instead of app

apps/remix-ide/src/app/udapp/run-tab.js
```diff
@@ -201,6 +201,31 @@ export class RunTab extends LibraryPlugin {
     this.renderDropdown(this.udappUI, this.fileManager, this.compilersArtefacts, this.config, this.editor, this.logCallback)
     this.renderRecorder(this.udappUI, this.fileManager, this.config, this.logCallback)
     this.renderRecorderCard()
+
+    this.on('manager', 'pluginDeactivated', profile => {
+      if (profile.kind === 'provider') this.blockchain.removeProvider(profile.name)
+    })
+    this.on('manager', 'pluginActivated', profile => {
```
iurimatias, line 208: suggestion: `this.on('manager', 'pluginActivated', this.addProfileProvider.bind(this)` and then move the code to that function

apps/remix-ide/src/app/udapp/run-tab.js
```diff
@@ -201,6 +201,31 @@ export class RunTab extends LibraryPlugin {
...
+      if (profile.kind === 'provider') {
+        ((profile, app) => {
+          const web3Provider = {
+            sendAsync (payload, callback) {
```
iurimatias, line 212: suggestion, something like:

```javascript
async sendAsync (payload, callback) {
              console.log('wallet connect', payload)
              try {
                const result = await app.call(profile.name, 'sendAsync', payload)
                console.log('wallet connect', result)
                callback(null, result)
              } catch (e) {
                  console.log('wallet connect', e)
                  callback(e)
                })
}
```

apps/remix-ide/src/app/udapp/run-tab.js
yann300: `this` can be removed yes. 
`(() => {})()`  is to make sure the creation of the provider is encapsulated. (that the local var ` profile` doesn't get overridden in case the event callback is triggered multiple time.
But there might be better way to handle that.

yann300: actually I had to keep `this` ...

yann300:
@iurimatias just pushed some updates

ryestew (approved):

PR #343, opened 2020-08-11
Moved E2E Tests To Typescript

ioedeveloper:


apps/remix-ide-e2e/src/commands/addAtAddressInstance.ts
```diff
@@ -1,8 +1,9 @@
-const EventEmitter = require('events')
+import { NightwatchBrowser } from 'nightwatch'
+import EventEmitter from "events"
 
 class addAtAddressInstance extends EventEmitter {
-  command (address, isValidFormat, isValidChecksum) {
-    this.api.perform((done) => {
+  command (this: NightwatchBrowser, address: string, isValidFormat: boolean, isValidChecksum: boolean): NightwatchBrowser {
```
yann300, line 5: sounds like risky to name a variable `this`, could be addressed in an another PR..

ioedeveloper: It is not a variable. `this` is the context which the command function uses. It is normally passed by the caller of the `command` function from nightwatch. Also, passing parameters into the `command` function without specifying `this` works perfectly. It just uses the default context for `this` which is the `addAtAddressInstance` class for this example.

apps/remix-ide-e2e/src/reports/CHROME_84.0.4147.89_Mac_OS_X_sample.test.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
yann300, line 1: can we remove all the reports files or is this necessary to keep them?

ioedeveloper: Yeah, we don't need the report files. But git ignore the directories during commit because they are empty. We will need something in it at least to get git to commit the directory.

yann300 (approved):

Issue #54, opened 2019-11-22
Make communication link more reachable


Aniket-Engg:
At present, we have our gitter channel link inside 'settings' tab as “Have a question”. This doesn’t seem a setting. We should bring it out at some significant place and probably try to generalise it a bit more like “Talk with Remix” or “Ask Remix” or “Remix Forum”. A user shouldn’t only come to gitter when he/she has some question but if he/she has a suggestion, feedback or just curious to say something about Remix can use our gitter channel.

Issue #196, opened 2020-04-14
add a website link to Home page


LianaHus:

LianaHus:
the link already there but we should consider to make it more visible after the UI inprovments of website
@ryestew

PR #351, opened 2020-08-12
Homepage UI refactoring

LianaHus:
fix https://github.com/ethereum/remix-project/issues/54
fix https://github.com/ethereum/remix-project/issues/73
 https://github.com/ethereum/remix-project/issues/196 the link should be added after website update https://github.com/ethereum/remix-project/issues/358

### DIFF
--- a/apps/remix-ide/src/app/ui/landing-page/landing-page.js
+++ b/apps/remix-ide/src/app/ui/landing-page/landing-page.js
@@ -61,9 +61,6 @@ let css = csjs`
     height: 150px;
     opacity: 0.7;
   }
-  .enviroments {
-    display: flex;
-  }
   .envLogo {
     height: 16px;
   }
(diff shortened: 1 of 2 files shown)


LianaHus:
1. `New Features` should come instead of learn more.
2. yeah. but what we can do, we want to have it in home, no?
3. we discussed this. we merged those two sections
4. this section become icons together with resources. it was planned in issue

we should fix Sourcify icon

in general, I agree that is is noisier now, i tried to remove some not necessary stuff but if we put Titter and Medium there is not much we can do.

ryestew:
OK - now I'm almost approving... but when I'm on the light theme and move the terminal up and down a bit I see this:
image

So there are double scroll bars and the gray is invading the white area.

ioedeveloper (approved):

PR #357, opened 2020-08-14
expected vs returned in remix-tests result

### LINKED ISSUE #321: remix-tests could export result with difference in expected vs returned values
It would be nice if remix-tests can export not just that an assertion failed but also the difference in expected vs returned values

Aniket-Engg:
fixes #321

### DIFF
--- a/libs/remix-tests/sol/tests.sol.ts
+++ b/libs/remix-tests/sol/tests.sol.ts
@@ -15,6 +15,13 @@ library Assert {
     uint256 expected
   );
 
+  event AssertionEventInt(
+    bool passed,
+    string message,
+    int256 returned,
+    int256 expected
+  );
+
   function ok(bool a, string memory message) public returns (bool result) {
     result = a;
     emit AssertionEvent(result, message);
@@ -25,9 +32,9 @@ library Assert {
     emit AssertionEventUint(result, message, a, b);
   }
 
-  function equal(int a, int b, string memory message) public returns (bool result) {
+  function equal(int256 a, int256 b, string memory message) public returns (bool result) {
     result = (a == b);
-    emit AssertionEvent(result, message);
+    emit AssertionEventInt(result, message, a, b);
   }
 
   function equal(bool a, bool b, string memory message) public returns (bool result) {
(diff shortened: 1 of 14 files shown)


libs/remix-tests/sol/tests.sol.ts
```diff
@@ -8,14 +8,21 @@ library Assert {
...
+    uint256 returned,
+    uint256 expected
+  );
+
   function ok(bool a, string memory message) public returns (bool result) {
     result = a;
     emit AssertionEvent(result, message);
   }
 
-  function equal(uint a, uint b, string memory message) public returns (bool result) {
```
yann300, line 16: why did you change the type to `uint256`?

libs/remix-tests/src/testRunner.ts
```diff
@@ -228,20 +229,26 @@ export function runTest (testName: string, testObject: any, contractDetails: Com
...
                     for (const i in receipt.events) {
                         const event = receipt.events[i]
-                        if (event.raw.topics.indexOf(topic) >= 0) {
-                            const testEvent = web3.eth.abi.decodeParameters(['bool', 'string'], event.raw.data)
+                        const eIndex = assertionEventHashes.indexOf(event.raw.topics[0]) // event name topic will always be at index 0
+                        if (eIndex >= 0) {
+                            const testEvent = web3.eth.abi.decodeParameters(assertionEvents[eIndex].params, event.raw.data)
                             if (!testEvent[0]) {
+                                if(eIndex === 0) { // for 'Assert.ok' method
+                                    testEvent[2] = 'false'
```
yann300, line 241: how can you be sure at this point that the returned value is `false`

libs/remix-tests/sol/tests.sol.ts
Aniket-Engg: `uint` is alias of `uint256`. `uint256` is required in event details here: https://github.com/ethereum/remix-project/blob/expectVsReturn/libs/remix-tests/src/assertionEvents.ts#L8 while decoding the event data, so I put it everywhere

libs/remix-tests/src/testRunner.ts
Aniket-Engg: Because for `Assert.ok` it has only two params. `if (!testEvent[0])` this condition will be satisfied when we will get `false` while expecting `true`. So I explicitly put those parameters in the object to access it similarly afterwards.

yann300 (approved):

PR #367, opened 2020-08-19
best-practices from remix-ide repo

yann300:


best-practices.md
```diff
@@ -0,0 +1,153 @@
+# Current "Best Practice" Conventions
+
+
+- Please use [JS Standard Style](https://standardjs.com/) as a coding style guide.
```
Aniket-Engg, line 4: I think we don;t use standard now

yann300: what is the spec of the new linter?

Aniket-Engg (approved):

Issue #70, opened 2020-06-23
Update UI of Settings module 


ryestew:
Update to Ethworks Figma:
https://www.figma.com/file/giWj3yC998FcVDOqX8y3ia/Remix-Final?node-id=0%3A1

yann300:
linked to https://github.com/ethereum/remix-project/issues/214

LianaHus:
th input field for token pass needs to have a proper background color

ryestew:
I believe this has been fixed.

PR #368, opened 2020-08-20
Settings Tab UI Update

ioedeveloper:
Closes https://github.com/ethereum/remix-project/issues/70 and closes https://github.com/ethereum/remix-project/issues/369
Preview: https://ipfsgw.komputing.org/ipfs/QmQ6JmkpTmtwpYzfZ3jx2rT4RujbHWKr5BmskdeyHfhQLY

### DIFF
--- a/apps/remix-ide-e2e/src/commands/renameFile.ts
+++ b/apps/remix-ide-e2e/src/commands/renameFile.ts
@@ -42,7 +42,7 @@ function renameFile (browser: NightwatchBrowser, path: string, newFileName: stri
         })
       })
       .click('body') // blur
-      .waitForElementVisible('#modal-footer-ok', 10000)
+      .waitForElementVisible('#modal-footer-ok', 100000)
       .pause(2000)
       .click('#modal-footer-ok')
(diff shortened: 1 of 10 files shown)


ryestew:
I was going to say that there is a problem with the text of the labels - because it is text-secondary because it is a bit hard to read because it is so light.  Then I clicked on it to check the box -and the text got darker and legible!  So let's go with your fancy UI!  It works nicely once you know what is going on.  

There is a problem with the github access token on Chrome on a mac in the DARK theme- ( this is not an issue in firefox)

image

I looked briefly and I don't see why your code has a issue.  If you look on the run & deploy tab - you'll see that the background color of the text boxes are a dark color.  But it didn't narrow it down to figure out the problem with your settings tab.

ioedeveloper:
@ryestew Funny because i have a dark color for github access token input box. Maybe you need to clear your cache.
Screenshot 2020-08-21 at 08 20 47

ryestew:
I checked on a different mac and it does not have the issue - so I assume you are right and it was something that got stuck in my cache.

PR #422, opened 2020-09-14
changed the way media panels are shown

LianaHus:
Second PR for https://github.com/ethereum/remix-project/issues/73


ryestew (commented):
* I think the media badges could be a bit smaller - like 1.2em or so.  It seems that they are too large in relation to Remi. 
* The shadow of the iframe could start from a lighter color of gray and could be a bit wider too.

ryestew (approved):

PR #427, opened 2020-09-16
Tree-View React Library

ioedeveloper:


tsconfig.json
```diff
@@ -23,7 +23,10 @@
       "@remix-project/remix-simulator": ["dist/libs/remix-simulator/index.js"],
       "@remix-project/remix-solidity": ["dist/libs/remix-solidity/index.js"],
       "@remix-project/remix-tests": ["dist/libs/remix-tests/src/index.js"],
-      "@remix-project/remix-url-resolver": ["dist/libs/remix-url-resolver/index.js"]
+      "@remix-project/remix-url-resolver": [
+        "dist/libs/remix-url-resolver/index.js"
+      ],
+      "@remix-ui/tree-view": ["libs/remix-ui/tree-view/src/index.ts"]
```
Aniket-Engg, line 29: should `dist` folder be referred here?

Aniket-Engg: Not sure if name should be `remix-ui-tree-view` as it is like that everywhere

PR #437, opened 2020-09-22
call plugin using URL params  after page load

yann300:
`pluginCall` can be used as an url parameter to call a plugin after page loads.

This will be mostly used now for loading a learneth tutorial from a link.

### DIFF
--- a/apps/remix-ide/src/app.js
+++ b/apps/remix-ide/src/app.js
@@ -403,12 +403,22 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   const params = queryParams.get()
 
   // Set workspace after initial activation
-  if (Array.isArray(workspace)) {
-    try {
-      await appManager.activatePlugin(workspace)
-    } catch (e) {
-      console.error(e)
-    }
+  if (Array.isArray(workspace)) {    
+    appManager.activatePlugin(workspace).then(() => {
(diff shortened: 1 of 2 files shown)


ioedeveloper (approved):

apps/remix-ide/src/app.js
```diff
@@ -405,7 +405,21 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   // Set workspace after initial activation
   if (Array.isArray(workspace)) {
     try {
-      await appManager.activatePlugin(workspace)
+      appManager.activatePlugin(workspace).then(() => {
+        // If plugins are loaded from the URL params, we focus on the last one.
+        if (pluginLoader.current === 'queryParams' && Array.isArray(workspace) && workspace.length > 0) menuicons.select(workspace[workspace.length - 1])
```
Aniket-Engg, line 410: `Array.isArray(workspace)` is already checked above with `if`

apps/remix-ide/src/app.js
```diff
@@ -405,7 +405,21 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
   // Set workspace after initial activation
   if (Array.isArray(workspace)) {
     try {
```
Aniket-Engg, line 407: Not sure, if try-catch is still required as we don't have any `await` statement now

Issue #376, opened 2020-08-25
Highlight code in unit testing plugin for failed tests


Aniket-Engg:
Investigate and if feasible, pass location with failed tests and highlight code

- [x] add `location` in `remix-tests` result
- [ ] show in UI of SUT Plugin

PR #448, opened 2020-09-24
location for failed tests

Aniket-Engg:
Related to #376

### DIFF
--- a/libs/remix-tests/src/testRunner.ts
+++ b/libs/remix-tests/src/testRunner.ts
@@ -39,6 +39,36 @@ function isPayable(funcABI: FunctionDescription): boolean {
 }
 
 /**
+ * @dev Check node name
+ * @param node AST node
+ * @param name name
+ */
+
+function isNodeName(node: AstNode, name: string): boolean {
+    return node.name === name
+}
+
+/**
+ * @dev Check node type
+ * @param node AST node
+ * @param type type
+ */
+
+function isNodeType(node: AstNode, type: string): boolean {
+    return node.nodeType === type
+}
+
+/**
+ * @dev Check if node type is from the typesList
+ * @param node AST node to be checked
+ * @param typesList list of types
+ */
+
+function isNodeTypeIn(node: AstNode, typesList: string[]): boolean {
+    return typesList.includes(node.nodeType)
+}
+
+/**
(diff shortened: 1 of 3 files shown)


libs/remix-tests/src/testRunner.ts
```diff
@@ -84,6 +84,30 @@ function getAvailableFunctions (fileAST: AstNode, testContractName: string): str
     return funcList;
 }
 
+function getAssertMethodLocation (fileAST: AstNode, testContractName: string, functionName: string, assertMethod: string): string {
+    let location
+    if(fileAST.nodes && fileAST.nodes.length > 0) {
+        const contractAST: AstNode = fileAST.nodes.find(node => node.name === testContractName && node.nodeType === 'ContractDefinition')
+        if(contractAST && contractAST.nodes) {
```
GrandSchtroumpf, line 91: Same as above: `if(contractAST?.nodes?.length)`

libs/remix-tests/src/testRunner.ts
```diff
@@ -84,6 +84,30 @@ function getAvailableFunctions (fileAST: AstNode, testContractName: string): str
...
+            const funcNode: AstNode = contractAST.nodes.find(node => (node.name === functionName && node.nodeType === "FunctionDefinition"))
+            // Check if statement nodeType is 'ExpressionStatement', for example:
+            // Assert.equal(foo.get(), 100, "initial value is not correct");
+            // Check if statement nodeType is 'Return', for example:
+            // return Assert.equal(foo.get(), 100, "initial value is not correct");
+            const expressions = funcNode.body.statements.filter(s =>
```
GrandSchtroumpf, line 97: This is super difficult to read. It looks to me that you're doing a lot of filtering operations. If I were you I would create a bunch of reusable guard functions with nice naming: 
```typescript
function isStatement(node: AstNode): node is AstStatement {
  return node.nodeType === 'ExpressionStatement';
}
```
The example above might not make sense in your case. It's just to showcase the concept.

GrandSchtroumpf (approved):

libs/remix-tests/src/testRunner.ts
Aniket-Engg: Not sure if type guards are suitable here. Added helpers.

Issue #431, opened 2020-09-18
Refactor execution context & Remix simulator provider


yann300:
currently remix-simulator is using directly execution-context for making calls to the VM. This should be decoupled.

PR #451, opened 2020-09-26
misc refactors remix-debug / remix-lib

iurimatias:


libs/remix-debug/src/cmdline/contextManager.js
```diff
@@ -0,0 +1,57 @@
+const remixLib = require('@remix-project/remix-lib')
```
yann300, line 1: I think this file can be removed (it was already removed in an another PR)

libs/remix-debug/src/source/astWalker.js
```diff
@@ -0,0 +1,55 @@
+'use strict'
```
yann300, line 1: I think this file was also removed from a precedent PR

apps/remix-ide/src/app/udapp/make-udapp.js
```diff
@@ -2,7 +2,6 @@ var registry = require('../../global/registry')
 var remixLib = require('@remix-project/remix-lib')
 var yo = require('yo-yo')
 var EventsDecoder = remixLib.execution.EventsDecoder
-var TransactionReceiptResolver = require('../../lib/transactionReceiptResolver')
```
yann300, line 5: `transactionReceiptResolver.js` the file seems not removed, do you still need it?

libs/remix-lib/src/execution/txListener.js
```diff
@@ -9,6 +9,16 @@ const defaultExecutionContext = require('./execution-context')
 const txFormat = require('./txFormat')
 const txHelper = require('./txHelper')
 
+function addExecutionCosts(txResult, tx) {
```
yann300, line 12: This function is already defined in this file, not sure which one you want to keep

yann300:
fix https://github.com/ethereum/remix-project/issues/431

yann300 (approved):

Issue #320, opened 2019-06-20
optimization runs


ayrat555:
A new version of remix doesn't allow to specify the number of optimization runs when compiling a contract

Aniket-Engg:
- [ ] Update `remix-solidity`
- [ ] Update UI

LianaHus:
for UI. we should add the input field with tooltip for the number but it should be disabled when optimisation is off.

PR #452, opened 2020-09-28
added runs for compiler

Aniket-Engg:
Related to #320

### DIFF
--- a/libs/remix-solidity/src/compiler/types.ts
+++ b/libs/remix-solidity/src/compiler/types.ts
@@ -143,6 +143,7 @@ export interface Source {
 
 export interface CompilerInputOptions {
     optimize: boolean | number,
+    runs: number,
     libraries?:  {
         [fileName: string]: Record<string, string>
     },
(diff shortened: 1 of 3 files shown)


libs/remix-solidity/src/compiler/types.ts
```diff
@@ -143,6 +143,7 @@ export interface Source {
 
 export interface CompilerInputOptions {
     optimize: boolean | number,
+    runs: number,
```
GrandSchtroumpf, line 146: Can it be any number or a subset of number ?
Let's take an example, if you want to display the status of a request you might want to use number literal :
```typescript
status: 200 | 400 | 500 | ...;
```
I'm not sure if it's the same in your case though.

GrandSchtroumpf (approved):

libs/remix-solidity/src/compiler/types.ts
Aniket-Engg: Usually It is 200. Further user can define it from 1 to bigger value

GrandSchtroumpf: ok, then keep it at number

LianaHus: do you know the range if any? could it be 10000000? @Aniket-Engg

PR #453, opened 2020-09-29
Debugger UI Refactor (React Library)

ioedeveloper:
Closes https://github.com/ethereum/remix-project/issues/399


libs/remix-ui/debugger-ui/src/lib/slider/slider.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useState, useEffect } from 'react'
+
+export const Slider = ({ stepManager, sliderValue }) => {
+  const [state, setState] = useState({
+    currentValue: ''
+  })
+
+  useEffect(() => {
```
yann300, line 8: Each time the slider is moved, line 9 get hit 1 time and line 13 2 times.
even if `if (value === state.currentValue) return` helps, that may be useful to check if that may cause performance issue in the stack.

libs/remix-ui/debugger-ui/src/lib/vm-debugger/dropdown-panel.tsx
```diff
@@ -0,0 +1,166 @@
+import React, { useState, useRef, useEffect } from 'react'
+import { TreeView, TreeViewItem } from '@remix-ui/tree-view'
```
yann300, line 2: An error is thrown while unfolding an item from the treeview. 
Also the way it is formmatted doesn't seem to be accurate
Screenshot from 2020-10-08 11-47-00

yann300: ^ that seems to be fixed

yann300:
- [ ] debugging `test.set(12)` (see bottom) gives a wrong state when being inside the `test2.set(12)` the value p should be 34 but it displays 8


```
contract test {
    uint p;
    test2 t;
    constructor () public {
        t = new test2();
        t.set(34);
    }
    
    function set(uint _p) public {
        p = _p;
        t.set(12);
    }
}

contract test2 {
    uint p;
    function set(uint _p) public {
        p = _p;
    }
}
```

yann300:
^ please scratch that... the issue exists in remix-alpha already

libs/remix-ui/utils/src/lib/should-render.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
```
yann300, line 1: This seems not use in this PR, is this still needed to keep it?

yann300 (approved):

PR #458, opened 2020-10-01
AST walk method updated

### LINKED ISSUE #415: Update remix-astwalker for latest AST
As latest AST is not quite generic for walking/traversing, spend time to analyse the latest AST and update the `remix-astwalker`

Aniket-Engg:
fixes #415


libs/remix-astwalker/src/astWalker.ts
```diff
@@ -58,62 +58,131 @@ export class AstWalker extends EventEmitter {
...
+    const normalizedNodes = []
+    objNodes.forEach((element) => {
+      const firstIndex = normalizedNodes.findIndex(e => e.id === element.id)
+      if(firstIndex == -1) normalizedNodes.push(element)
+    })
+    return normalizedNodes
+  }
+
+  getASTNodeChildren(ast: AstNode): AstNode[] {
+    const nodes = ast.nodes           // for ContractDefinition
```
yann300, line 83: What would be the best way to share this with the solidity team in order to keep this up to date? or is this stable enough so we don't really need to be worried about potential changes.

Aniket-Engg: I will think about it.

yann300 (approved):

PR #479, opened 2020-10-13
Fix Debugger Slider Bug

ioedeveloper:
Fixed https://github.com/ethereum/remix-project/issues/253

### DIFF
--- a/apps/remix-ide-e2e/src/tests/debugger.test.ts
+++ b/apps/remix-ide-e2e/src/tests/debugger.test.ts
@@ -129,6 +129,25 @@ module.exports = {
     .goToVMTraceStep(717)
     .pause(5000)
     .checkVariableDebug('soliditylocals', localVariable_step717_ABIEncoder) // all locals should be initiaed
+  },
+
+  'Should load more solidity locals array': function (browser: NightwatchBrowser) {
+    browser.addFile('locals.sol', sources[3]['browser/locals.sol'])
+    .clickLaunchIcon('udapp')
+    .createContract('')
+    .clickInstance(3)
+    .clickFunction('t - transact (not payable)')
+    .pause(2000)
+    .debugTransaction(6)
+    .waitForElementVisible('*[data-id="slider"]')
+    .click('*[data-id="slider"]')
+    .setValue('*[data-id="slider"]', '5000')
+    .waitForElementPresent('*[data-id="treeViewTogglearray"]')
+    .click('*[data-id="treeViewTogglearray"]')
+    .waitForElementPresent('*[data-id="treeViewLoadMore"]')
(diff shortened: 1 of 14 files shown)


libs/remix-debug/src/solidity-decoder/types/ArrayType.js
```diff
@@ -76,23 +76,35 @@ class ArrayType extends RefType {
...
       offset = offset + 32
     }
-    for (var k = 0; k < length; k++) {
+    if (isNaN(length)) {
+      return {
+        value: '<decoding failed - length is NaN>',
+        type: this.typeName
+      }
+    }
+    let limit = length
```
yann300, line 93: shouldn't `limit` be `length - skip` ?

yann300:
I tried with
```
contract test {
    function t () public {
        uint[] memory array = new uint[](150);
        for (uint k = 0; k < 150; k++) {
            array[k] = k;
        }
    }
}
```
`load more` works but it has loaded until `249` although the actual size is `150`

yann300:
From UX perspektives, that might be good to open the array in a modal... (for a next PR anyway)

yann300:
might even be fine to only loads the first 10 or 20 items from the array.. although 100 is fine (just a remark)
Also we should also probably do this for state array.

yann300:
> Also we should also probably do this for state array.

I suppose not in this PR.

yann300:
Should we have e2e for this?
The contract above could fit.

apps/remix-ide-e2e/src/tests/debugger.test.ts
```diff
@@ -94,6 +93,29 @@ module.exports = {
...
     'current displayed content is not from the ERC20 source code')
     })
+  },
+
+  'Should load more solidity locals array': function (browser: NightwatchBrowser) {
+    browser.addFile('locals.sol', sources[2]['browser/locals.sol'])
+    .clickLaunchIcon('udapp')
+    .waitForElementPresent('*[data-id="deployAndRunClearInstances"]')
+    .click('*[data-id="deployAndRunClearInstances"]')
+    .waitForElementPresent('*[title="Deploy - transact (not payable)"]')
```
yann300, line 103: if possible that'd better to use the commands. i.e something like
```
selectContract('test')
    .createContract('')
    .clickInstance(2)
    .clickFunction('test1 - transact (not payable)', {types: 'bytes userData', values: '0x000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000015b38da6a701c568545dcfcb03fcb875f56beddc4'})
    
```

yann300 (approved):

yann300 (approved):

PR #526, opened 2020-10-28
Hide Highlights of plugin if not in focused

LianaHus:


apps/remix-ide/src/app/editor/SourceHighlighters.js
```diff
@@ -28,13 +25,40 @@ class SourceHighlighters {
     }
   }
 
+  highlightAllFrom (from) {
```
yann300, line 28: could you put some doc to this function?

apps/remix-ide/src/app/editor/SourceHighlighters.js
```diff
@@ -28,13 +25,40 @@ class SourceHighlighters {
...
+    try {
+      if (!this.highlighters[from]) return
+      const sourceHighlight = new SourceHighlighter()
```
yann300, line 31: From what I see you don't need to re-instanciate each time a `sourceHighlight = new SourceHighlighter()` but only if it doesn't already exist

LianaHus: they keep an ad for each marker and if I don't do it the new marker has different id

apps/remix-ide/src/app/editor/sourceHighlighter.js
```diff
@@ -39,6 +39,7 @@ class SourceHighlighter {
     this.source = null
     if (lineColumnPos) {
       this.source = filePath
+      if (!this.source) this.source = this._deps.fileManager.currentFile()
```
yann300, line 42: Why do you need to do that? i'd say if `filePath` is not defined the function can't anyway process and should return an error or just return.

apps/remix-ide/src/app/editor/editor.js
```diff
@@ -189,14 +189,17 @@ class Editor extends Plugin {
...
       this.event.trigger('sessionSwitched', [])
-
       this.editor.getSession().on('change', () => {
         this._onChange()
         this.event.trigger('contentChanged', [])
       })
     })
   }
 
+  onActivation () {
```
yann300, line 199: might need to do something like `onDeactivation` : `this.off('sidePanel', 'focusChanged')` .
Please check the syntax.

yann300:
That should probably need a little e2e test. 
At least to test that highlight is removed when switched to another view

apps/remix-ide/src/app/editor/SourceHighlighters.js
yann300: > ad ?

apps/remix-ide/src/app/editor/editor.js
LianaHus: right

apps/remix-ide/src/app/editor/SourceHighlighters.js
LianaHus: sorry. "an id"

apps/remix-ide/src/app/editor/SourceHighlighters.js
```diff
@@ -28,13 +25,40 @@ class SourceHighlighters {
...
+      for (const index in this.highlighters[from]) {
+        sourceHighlight.currentSourceLocationFromfileName(
+          this.highlighters[from][index].position,
+          this.highlighters[from][index].filePath,
```
yann300, line 35: in which case do you have `filePath` here ? I didn't see `filePath` but `source` is present. Also `hexColor` seems not

yann300: ^ might be the reason why you have `filePath` undefined and that you needed to to `if (!this.source) this.source = this._deps.fileManager.currentFile()`

apps/remix-ide/src/app/editor/contextualListener.js
```diff
@@ -123,14 +123,15 @@ class ContextualListener extends Plugin {
   }
 
   _highlightInternal (position, node) {
+    if (node.nodeType == 'Block') return
```
ioedeveloper, line 126: Can you change this to strict equality check `===` not `==`

ioedeveloper (changes requested):

apps/remix-ide/src/app/editor/contextualListener.js
LianaHus: :) of course, thanks. you are checking the standart manually :)

ioedeveloper (approved):

Issue #519, opened 2020-10-22
The 'Drak' theme is being load after reload


LianaHus:
the saved one should be loaded from localstorage

PR #544, opened 2020-11-02
save the theme for reload

LianaHus:
fix https://github.com/ethereum/remix-project/issues/519

### DIFF
--- a/apps/remix-ide/src/app/tabs/theme-module.js
+++ b/apps/remix-ide/src/app/tabs/theme-module.js
@@ -77,7 +77,7 @@ export class ThemeModule extends Plugin {
     }
     const next = themeName || this.active   // Name
     const nextTheme = this.themes[next] // Theme
-    if (!this.forced) this._deps.config.set('settings/theme', next)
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/tabs/theme-module.js
```diff
@@ -77,7 +77,7 @@ export class ThemeModule extends Plugin {
     }
     const next = themeName || this.active   // Name
     const nextTheme = this.themes[next] // Theme
-    if (!this.forced) this._deps.config.set('settings/theme', next)
```
yann300, line 80: we can't remove this test I think.
`this.forced` is set when the theme is set from the url parameter, in that case we don't want to persist that.
the issue could be that the value of `this.forced` is not correctly setup

LianaHus: right. when we set it we compare null with undefined

LianaHus: should be !!queryTheme instead

Aniket-Engg (approved):

PR #545, opened 2020-11-02
Fixed tab title hidden display

ioedeveloper:


LianaHus (approved):
This is temporary solution of course bit we can go with it for now.

Issue #515, opened 2020-10-20
Publish to Docker sometimes fails


yann300:
with 
```
#!/bin/bash -eo pipefail
./apps/remix-ide/ci/build_and_publish_docker_images.sh
CONTRIBUTING.md    dist			package.json
Dockerfile	   docker-compose.yaml	release-process.md
Dockerfile.dev	   gulpfile.js		team-best-practices.md
LICENSE		   jest.config.js	temp_publish_docker
README.md	   lerna.json		tmp
apps		   libs			tsconfig.json
babel.config.js    node_modules		workspace.json
best-practices.md  nx.json		yarn.lock
build.yaml	   package-lock.json
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
WARNING! Your password will be stored unencrypted in /home/circleci/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
Building remixide
ERROR: Couldn't connect to Docker daemon at https://35.231.88.155:2376 - is it running?

If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.

Exited with code exit status 1
CircleCI received exit code 1
```

yann300:
cc @edisinovcic

Jakic007:
The reason for this is that docker image that is to be built is tagged with the branch name which in this case has illegal character "#".
Error is misleading but if you check again, on all branches named "fix#xxx" docker-compose build fails while on the others succeeds.
There is already an issue about that on docker-compose repository.
https://github.com/docker/compose/issues/6050

Proposed solution:
Replace every "#" with for example "-" in the CI script.

edisinovcic:
@Jakic007 good catch missed that one. Feel free to change all the "#" to "-". Tnx

yann300:
doh! many thanks @Jakic007 @edisinovcic

PR #555, opened 2020-11-05
docker-compose build fix

Jakic007:
This small change fixes those failed docker-compose builds that have happened for branches that included character "#" in their name which is illegal character when tagging docker image.

This change will replace every "#" character with dash character and solve the issue.
https://github.com/ethereum/remix-project/issues/515


edisinovcic (approved):

edisinovcic:
@yann300 this is ready to be merged, should we target some other branch?

yann300 (approved):

PR #590, opened 2020-11-17
added release from remix_beta section

LianaHus:


yann300:
`## Remix IDE release` should be removed I suppose

LianaHus:
> `## Remix IDE release` should be removed I suppose

I would keep this for now if you don't mind. we have all kinds of release instructions here.

Aniket-Engg (approved):

PR #617, opened 2020-11-23
Convert single quoted strings to byte32 arguments

QaDeS:
Just starting out, and not being able to test the Ballot contract easily irritated me a lot. With this change, you're able to just pass in "['foo','bar']" when instantiating the contract.

Error handling is clunky even though it works, and I m not 100% sure it does not break anything. Open for suggestions on how to improve :)

### DIFF
--- a/apps/remix-ide/src/app/ui/multiParamManager.js
+++ b/apps/remix-ide/src/app/ui/multiParamManager.js
@@ -6,6 +6,25 @@ var copyToClipboard = require('./copy-to-clipboard')
 var remixLib = require('@remix-project/remix-lib')
 var txFormat = remixLib.execution.txFormat
 
+function stringToBytes32(str) {
+  const bytes = new TextEncoder().encode(str)
+  if(bytes.length > 32) {
+    // TODO find a better way to report the error
+    alert(`String too long '${str}'`)
+    return "'${str}'"
+  }
+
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/ui/multiParamManager.js
```diff
@@ -6,6 +6,25 @@ var copyToClipboard = require('./copy-to-clipboard')
 var remixLib = require('@remix-project/remix-lib')
 var txFormat = remixLib.execution.txFormat
 
+function stringToBytes32(str) {
+  const bytes = new TextEncoder().encode(str)
+  if(bytes.length > 32) {
+    // TODO find a better way to report the error
```
LianaHus, line 12: Please add an issue and mention the number here if you can't avoid todos
@todo(#number) - description

Aniket-Engg:
As per discussion, it may not be good to add this custom solution to automatically convert string to Bytes32.

We can have a separate plugin for all possible conversion. What do you think @ryestew @ioedeveloper ?

For now, string can be converted to Bytes32 from Remix terminal using `ethers.utils.formatBytes32String("remix")`

Aniket-Engg:
As per discussion with @ryestew during UX standup, a plugin is being planned for conversions.

PR #622, opened 2020-11-24
updated the tooltip of "Runs"

LianaHus:


Aniket-Engg (approved):
I think this can be a bit short but it seems required to clear the concept of `runs`

Issue #619, opened 2020-11-23
Expose handleImportCall


defiboy:
At the moment the[ remix-solidity compiler ](https://github.com/ethereum/remix-project/blob/master/libs/remix-solidity/src/compiler/compiler.ts#L20) requires a handleImportCall function as parameter in the constructor and one of the consumers of this method is the [CompileTab](https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/tabs/compileTab/compileTab.js#L15), that internally defines this function.

In order to allow plugins to reuse this logic we should expose this functionality. Probably via a plugin? or any other mechanism?

PR #624, opened 2020-11-25
Extends contentImport with resolveAndSave

yann300:
fix https://github.com/ethereum/remix-project/issues/619


yann300:
`resolve` resolve only external URL and doesn't save the result in the fileexplorer
`resolveAndSave` resolve internal and external URL, if external URL it saves the result in the fileexplorer.

Please would like to have input about using other function name.

Aniket-Engg (approved):

yann300:
last call before merging, anyone has feedback for the naming convention we should use here?

yann300:
cc @ryestew @bunsenstraat

PR #623, opened 2020-11-25
compiler configuration from URL e2e tests

Aniket-Engg:
fixes #533

PR #625, opened 2020-11-25
updated runs from URL

Aniket-Engg:
updated runs from URL on compiler plugin UI

### DIFF
--- a/apps/remix-ide/src/app/tabs/compileTab/compilerContainer.js
+++ b/apps/remix-ide/src/app/tabs/compileTab/compilerContainer.js
@@ -197,12 +197,14 @@ class CompilerContainer {
       class="custom-select ml-2 ${css.runs}"
       id="runs"
       placeholder="200"
+      value="200"
       type="number"
       title="Estimated number of times each opcode of the deployed code will be executed across the life-time of the contract."
       onchange=${() => this.onchangeRuns()}
     >`
-    if (this.compileTabLogic.optimize) this._view.runs.removeAttribute('disabled')
-    else {
+    if (this.compileTabLogic.optimize) {
+      this._view.runs.value = this.compileTabLogic.runs
+    } else {
       this._view.runs.setAttribute('disabled', '')
     }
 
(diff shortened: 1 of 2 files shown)


apps/remix-ide/src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -201,8 +201,9 @@ class CompilerContainer {
       title="Estimated number of times each opcode of the deployed code will be executed across the life-time of the contract."
       onchange=${() => this.onchangeRuns()}
     >`
-    if (this.compileTabLogic.optimize) this._view.runs.removeAttribute('disabled')
-    else {
+    if (this.compileTabLogic.optimize) {
+      this._view.runs.value = this.compileTabLogic.runs
```
LianaHus, line 205: i think you should still remove disable

Aniket-Engg: by checking it locally, that doesn't seem required. Because by default it is enabled and disabled only when `this.compileTabLogic.optimize === false
`

Aniket-Engg: probably because disable is removed here https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/tabs/compileTab/compilerContainer.js#L347

LianaHus: by default, it should be disabled. I would prefer to leave it as it is setting the initial state.

yann300 (approved):

LianaHus:
this is not working for me. i put 300 and after refresh, it is still 200

LianaHus (changes requested):
Please also add e2e test

LianaHus:
also sometimes I have "NaN" in URL

apps/remix-ide/src/app/tabs/compileTab/compilerContainer.js
Aniket-Engg: > Please also add e2e test

will do once https://github.com/ethereum/remix-project/pull/623 is merged

LianaHus (approved):

PR #638, opened 2020-11-26
Add git service in remixd & use it from the terminal

yann300:
fix https://github.com/ethereum/remix-project/issues/592

based on https://github.com/ethereum/remixd/pull/99

### DIFF
--- a/apps/remix-ide/src/app.js
+++ b/apps/remix-ide/src/app.js
@@ -390,7 +390,8 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     debug,
     analysis,
     test,
-    filePanel.remixdHandle
+    filePanel.remixdHandle,
+    filePanel.gitHandle
   ])
 
   try {
--- a/apps/remix-ide/src/app/files/git-handle.js
+++ b/apps/remix-ide/src/app/files/git-handle.js
@@ -16,12 +16,4 @@ export class GitHandle extends WebsocketPlugin {
   constructor () {
     super(profile)
   }
-
-  deactivate () {
-    super.deactivate()
-  }
-
-  activate () {
-    super.activate()
-  }
 }
--- a/apps/remix-ide/src/app/files/git-handle.js
+++ b/apps/remix-ide/src/app/files/git-handle.js
@@ -5,8 +5,7 @@ const profile = {
   name: 'git',
   displayName: 'Git',
   url: 'ws://127.0.0.1:65521',
-  methods: ['command'],
-  events: [],
+  methods: ['execute'],
   description: 'Using Remixd daemon, allow to access git API',
   kind: 'other',
   version: packageJson.version
--- a/apps/remix-ide/src/app/files/git-handle.js
+++ b/apps/remix-ide/src/app/files/git-handle.js
@@ -1,0 +1,27 @@
+import { WebsocketPlugin } from '@remixproject/engine-web'
+import * as packageJson from '../../../../../package.json'
+
+const profile = {
+  name: 'git',
+  displayName: 'Git',
+  url: 'ws://127.0.0.1:65521',
+  methods: ['command'],
+  events: [],
+  description: 'Using Remixd daemon, allow to access git API',
+  kind: 'other',
+  version: packageJson.version
+}
+
+export class GitHandle extends WebsocketPlugin {
+  constructor () {
+    super(profile)
+  }
+
+  deactivate () {
+    super.deactivate()
+  }
+
+  activate () {
+    super.activate()
+  }
+}
--- a/apps/remix-ide/src/app/files/remixd-handle.js
+++ b/apps/remix-ide/src/app/files/remixd-handle.js
@@ -84,7 +84,7 @@ export class RemixdHandle extends WebsocketPlugin {
           }
         }, 3000)
         this.locahostProvider.init(_ => this.fileSystemExplorer.ensureRoot())
-        this.call('manager', 'activatePlugin', 'git')    
+        this.call('manager', 'activatePlugin', 'git')
       }
     }
     if (this.locahostProvider.isConnected()) {
--- a/apps/remix-ide/src/app/files/remixd-handle.js
+++ b/apps/remix-ide/src/app/files/remixd-handle.js
@@ -40,7 +40,7 @@ export class RemixdHandle extends WebsocketPlugin {
   deactivate () {
     this.fileSystemExplorer.hide()
     if (super.socket) super.deactivate()
-    this.appManager.ensureDeactivated('git')
+    this.call('manager', 'deactivatePlugin', 'git')
     this.locahostProvider.close((error) => {
       if (error) console.log(error)
     })
@@ -52,8 +52,8 @@ export class RemixdHandle extends WebsocketPlugin {
   }
 
   async canceled () {
-    this.appManager.ensureDeactivated('remixd')
-    this.appManager.ensureDeactivated('git')
+    this.call('manager', 'deactivatePlugin', 'remixd')
+    this.call('manager', 'deactivatePlugin', 'git')
   }
 
   /**
@@ -84,7 +84,7 @@ export class RemixdHandle extends WebsocketPlugin {
           }
         }, 3000)
         this.locahostProvider.init(_ => this.fileSystemExplorer.ensureRoot())
-        this.appManager.ensureActivated('git')
+        this.call('manager', 'activatePlugin', 'git')    
       }
     }
     if (this.locahostProvider.isConnected()) {
--- a/apps/remix-ide/src/app/files/remixd-handle.js
+++ b/apps/remix-ide/src/app/files/remixd-handle.js
@@ -40,6 +40,7 @@ export class RemixdHandle extends WebsocketPlugin {
   deactivate () {
     this.fileSystemExplorer.hide()
     if (super.socket) super.deactivate()
+    this.appManager.ensureDeactivated('git')
     this.locahostProvider.close((error) => {
       if (error) console.log(error)
     })
@@ -52,6 +53,7 @@ export class RemixdHandle extends WebsocketPlugin {
 
   async canceled () {
     this.appManager.ensureDeactivated('remixd')
+    this.appManager.ensureDeactivated('git')
   }
 
   /**
@@ -82,6 +84,7 @@ export class RemixdHandle extends WebsocketPlugin {
           }
         }, 3000)
         this.locahostProvider.init(_ => this.fileSystemExplorer.ensureRoot())
+        this.appManager.ensureActivated('git')
       }
     }
     if (this.locahostProvider.isConnected()) {
--- a/apps/remix-ide/src/app/panels/file-panel.js
+++ b/apps/remix-ide/src/app/panels/file-panel.js
@@ -5,6 +5,7 @@ var yo = require('yo-yo')
 var EventManager = require('../../lib/events')
 var FileExplorer = require('../files/file-explorer')
 var { RemixdHandle } = require('../files/remixd-handle.js')
+var { GitHandle } = require('../files/git-handle.js')
 var globalRegistry = require('../../global/registry')
 var css = require('./styles/file-panel-styles')
 
@@ -60,6 +61,7 @@ module.exports = class Filepanel extends ViewPlugin {
     var fileSystemExplorer = createProvider('localhost')
 
     self.remixdHandle = new RemixdHandle(fileSystemExplorer, self._deps.fileProviders.localhost, appManager)
+    self.gitHandle = new GitHandle()
 
     const explorers = yo`
       <div>
--- a/apps/remix-ide/src/app/panels/terminal.js
+++ b/apps/remix-ide/src/app/panels/terminal.js
@@ -103,7 +103,7 @@ class Terminal extends Plugin {
     })
     this.on('scriptRunner', 'error', (msg) => {
       this.commands.error.apply(this.commands, msg.data)
-    })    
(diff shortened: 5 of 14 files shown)


apps/remix-ide/src/app/files/git-handle.js
```diff
@@ -0,0 +1,27 @@
+import { WebsocketPlugin } from '@remixproject/engine-web'
+import * as packageJson from '../../../../../package.json'
+
+const profile = {
+  name: 'git',
+  displayName: 'Git',
+  url: 'ws://127.0.0.1:65521',
+  methods: ['command'],
+  events: [],
```
GrandSchtroumpf, line 9: events is deprecated. You can remove it

apps/remix-ide/src/app/panels/terminal.js
```diff
@@ -485,7 +485,7 @@ class Terminal extends Plugin {
     return self._view.el
 
     function wrapScript (script) {
-      if (script.startsWith('remix.')) return script
+      if (script.trim().startsWith('remix.') || script.trim().startsWith('git')) return script
```
GrandSchtroumpf, line 488: Create a variable to make it easier to read: 
```javascript
const isRemix = script.trim().startsWith('remix.')
const isGit = script.trim().startsWith('git')
if (isRemix || isGit) return script;
```
or 
```javascript
const isKnownScript = ['remix.', 'git'].some(prefix => script.trim().startWith(prefix))
if (isKnownScript) return script
```

apps/remix-ide/src/app/panels/terminal.js
```diff
@@ -746,10 +746,15 @@ class Terminal extends Plugin {
       }
     }
     try {
-      await this.call('scriptRunner', 'execute', script)
+      if (script.trim().indexOf('git') === 0) {
```
GrandSchtroumpf, line 749: Keep the same logic everywhere. Use startWith instead of indexOf() (except if you need the index for later, which is not the case here).

apps/remix-ide/src/app/files/git-handle.js
```diff
@@ -0,0 +1,27 @@
+import { WebsocketPlugin } from '@remixproject/engine-web'
+import * as packageJson from '../../../../../package.json'
+
+const profile = {
+  name: 'git',
+  displayName: 'Git',
+  url: 'ws://127.0.0.1:65521',
+  methods: ['command'],
```
GrandSchtroumpf, line 8: Why `command` while the scriptRunner method is `execute` ? I think it makes it difficult for developers to follow up with the naming.
I would recommend to define a kind "terminal" or whatever with the right interface.

apps/remix-ide/src/app/panels/terminal.js
```diff
@@ -746,10 +746,15 @@ class Terminal extends Plugin {
...
+        const result = await this.call('git', 'command', script)
+        self.commands.html(yo`<pre>${result}</pre>`)
+      } else {
+        await this.call('scriptRunner', 'execute', script)
```
GrandSchtroumpf, line 753: Why not displaying the result here ? I guess it's because scriptRunner emit an event instead of returning a value.
It's quite difficult to follow and I think you should take the opportunity of this PR to homogenize the process. Ideally we should have : 
```javascript
const pluginName = {
  remix: 'scriptRunner',
  git: 'git'
}
const prefix = script.trim.split('.').unshift()
const plugin = pluginName[prefix]
const result = await this.call(plugin, 'command', script)
```

libs/remixd/src/bin/remixd.ts
```diff
@@ -38,15 +38,24 @@ import * as program from 'commander'
...
-      websocketHandler.start((ws: WS) => {
+      websocketHandler.start((ws: WS, sharedFolderClient: servicesList.Sharedfolder) => {
         sharedFolderClient.setWebSocket(ws)
         sharedFolderClient.setupNotifications(program.sharedFolder)
         sharedFolderClient.sharedFolder(program.sharedFolder, program.readOnly || false)
       })
       killCallBack.push(websocketHandler.close.bind(websocketHandler))
+
+      // git
+      const websocketHandlerForGit = new WebSocket(65521, { remixIdeUrl: program.remixIde }, () => new servicesList.GitClient())
```
GrandSchtroumpf, line 52: Look like code duplication. If remixd is supposed to manage multiple port build it up for that : 
```javascript
const servicesList = {
  git: GitClient,
  folder: Sharedfolder
}
...
const ports = {
  git: 65521,
  folder: 65520
}
function startService<S extends 'git' | 'folder'>(service: S, client: servicesList[T]) {
  const socket = new Websocket(port[service], new servicesList[service]());
  socket.start((...) => ...)
}
```

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,62 @@
+import * as WS from 'ws' // eslint-disable-line
+import { PluginClient } from '@remixproject/plugin'
+const { spawn } = require('child_process')
+const gitRegex = '^git\\s[^&|;]*$'
+
+export class GitClient extends PluginClient {
+  methods: ['command']
```
GrandSchtroumpf, line 7: Same as above

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  sharedFolder (currentSharedFolder: string, readOnly: boolean): void {
+    this.currentSharedFolder = currentSharedFolder
+    this.readOnly = readOnly
+  }
+
+  command (cmd: string) {
+    return new Promise((resolve, reject) => {
+      try {
+        try {
```
GrandSchtroumpf, line 24: The double try / catch is redundant You can remove it as if validateCommand fails it will be catch in the first try and reject the promise

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,62 @@
...
+          validateCommand(cmd, gitRegex)
```
GrandSchtroumpf, line 25: Looks like an [assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions). Maybe better to rename it as it.

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,62 @@
...
+          result += data.toString()
+        })
+        child.stderr.on('data', (err) => {
+          error += err.toString()
+        })
+        child.on('close', (exitCode) => {
+          if (exitCode !== 0) {
+            reject(error)
+          } else {
+            resolve(result + error)
```
GrandSchtroumpf, line 43: It doesn't looks like a good idea to concat result + error & return it like that in the resolve.

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.websocket = websocket
+  }
+
+  sharedFolder (currentSharedFolder: string, readOnly: boolean): void {
+    this.currentSharedFolder = currentSharedFolder
+    this.readOnly = readOnly
+  }
+
+  command (cmd: string) {
+    return new Promise((resolve, reject) => {
```
GrandSchtroumpf, line 22: I would recommend to put the Promise closer to its used : 
```typescript
assertCommand(cmd, gitRegex)
const options = { cwd: this.currentSharedFolder, shell: true }
const child = spawn(cmd, options)
let result = ''
let error = ''
  child.stdout.on('data', (data) => result += data.toString())
  child.stderr.on('data', (err) => error += data.toString())
  child.on('close', (exitCode) => exitCode !== 0 ? reject(error) : resolve(result))
})
```
You don't need to try / catch here everything that throw will be handled by the caller

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,62 @@
...
+    })
+  }
+}
+
+/**
+ * Validate that command can be run by service
+ * @param cmd
+ * @param regex
+ */
+function validateCommand (cmd, regex) {
```
GrandSchtroumpf, line 58: rename `assertCommand` and remove the regex parameter, except if you see a reason where it could be different than the one at the top of the file.

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,62 @@
+import * as WS from 'ws' // eslint-disable-line
+import { PluginClient } from '@remixproject/plugin'
+const { spawn } = require('child_process')
+const gitRegex = '^git\\s[^&|;]*$'
```
GrandSchtroumpf, line 4: Put the regex in the assertion function directly.

libs/remixd/src/websocket.ts
```diff
@@ -1,24 +1,24 @@
 import * as WS from 'ws'
 import * as http from 'http'
-import { WebsocketOpt, SharedFolderClient } from './types' // eslint-disable-line
+import { WebsocketOpt, ServiceClient } from './types' // eslint-disable-line
```
GrandSchtroumpf, line 3: any reason why you need to disable line ? I'm not sure you need to do that anymore.

libs/remixd/src/websocket.ts
```diff
@@ -1,24 +1,24 @@
...
       console.log((new Date()) + ' Received request for ' + request.url)
       response.writeHead(404)
       response.end()
     })
     const loopback = '127.0.0.1'
 
-    this.server.listen(this.port, loopback, function () {
-      console.log((new Date()) + ' remixd is listening on ' + loopback + ':65520')
+    this.server.listen(this.port, loopback, () => {
+      console.log((new Date()) + ' remixd is listening on ' + loopback + ':' + this.port + '')
```
GrandSchtroumpf, line 21: I would looks a little bit cleaner with a string template I think :
```typescript
console.log(`${new Date()} remixd is listening on ${loopback}:${this.port}`)
```

GrandSchtroumpf (changes requested):

ioedeveloper (approved):

ioedeveloper:
E2E looks good.

apps/remix-ide/src/app/panels/terminal.js
yann300: that gonna be hard to make `scriptRunner` to return a value, as we don't really know when a script finishes (it can `console.log` at any time). will try to update git to emit event instead.

apps/remix-ide/src/app/files/remixd-handle.js
```diff
@@ -82,6 +84,7 @@ export class RemixdHandle extends WebsocketPlugin {
           }
         }, 3000)
         this.locahostProvider.init(_ => this.fileSystemExplorer.ensureRoot())
+        this.appManager.ensureActivated('git')
```
GrandSchtroumpf, line 87: This has been removed in #651. Why not running `this.call('manager', 'activatePlugin', 'git')` ?
Or even let the engine lazily activate the git plugin on first call

apps/remix-ide/src/app/files/remixd-handle.js
```diff
@@ -40,6 +40,7 @@ export class RemixdHandle extends WebsocketPlugin {
   deactivate () {
     this.fileSystemExplorer.hide()
     if (super.socket) super.deactivate()
+    this.appManager.ensureDeactivated('git')
```
GrandSchtroumpf, line 43: Any reason why not using `this.call('manager', 'deactivatePlugin', 'git')` ?
Use call through the engine when possible instead of direct call to a plugin.

apps/remix-ide/src/app.js
```diff
@@ -387,7 +387,8 @@ Please make a backup of your contracts and start using http://remix.ethereum.org
     debug,
     analysis,
     test,
-    filePanel.remixdHandle
+    filePanel.remixdHandle,
+    filePanel.gitHandle
```
GrandSchtroumpf, line 391: as a Plugin I would recommend putting Git at the root of the project and access its methods only through remix-plugin API

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -60,6 +61,7 @@ module.exports = class Filepanel extends ViewPlugin {
     var fileSystemExplorer = createProvider('localhost')
 
     self.remixdHandle = new RemixdHandle(fileSystemExplorer, self._deps.fileProviders.localhost, appManager)
+    self.gitHandle = new GitHandle()
```
GrandSchtroumpf, line 64: I would recommend to put GitHandle at the root of the project and not in the FilePanel property

apps/remix-ide/src/app/panels/terminal.js
GrandSchtroumpf: You can increase the timeout of the scriptRunner plugin and return the value. It doesn't prevent you to listen on log events if you need

libs/remixd/src/services/gitClient.ts
```diff
@@ -0,0 +1,48 @@
...
+    let error = ''
+    child.stdout.on('data', (data) => {
+      result += data.toString()
+    })
+    child.stderr.on('data', (err) => {
+      error += err.toString()
+    })
+    child.on('close', () => {
+      if (error !== '') this.emit('error', error)
+      else this.emit('log', result)
```
GrandSchtroumpf, line 34: I think returning a value was a better way to handle it.

GrandSchtroumpf (approved):

apps/remix-ide/src/app.js
yann300: will plan this changes later on. This will anyway go in during the react refactoring.

PR #641, opened 2020-11-26
at address is disabled when no contract compiled

LianaHus:


apps/remix-ide/src/app/tabs/runTab/contractDropdown.js
```diff
@@ -29,12 +30,14 @@ class ContractDropdownUI {
       var contractNames = document.querySelector(`.${css.contractNames.classNames[0]}`)
       contractNames.innerHTML = ''
       if (success) {
-        this.selectContractNames.removeAttribute('disabled')
+        this.enableContractNames(true)
+        this.atAddress.removeAttribute('disabled')
```
yann300, line 34: can you use `enableAtAddress` here (and below)?

apps/remix-ide/src/app/tabs/runTab/contractDropdown.js
```diff
@@ -145,24 +175,32 @@ class ContractDropdownUI {
...
     if (!document.querySelector(`.${css.contractNames}`)) return
-    var contractNames = document.querySelector(`.${css.contractNames.classNames[0]}`)
     if (/.(.abi)$/.exec(currentFile)) {
       this.createPanel.style.display = 'none'
       this.orLabel.style.display = 'none'
       this.compFails.style.display = 'none'
-      contractNames.appendChild(yo`<option>(abi)</option>`)
-      this.selectContractNames.setAttribute('disabled', true)
+      this.loadType = 'abi'
+      this.enableContractNames(false)
```
yann300, line 196: should you `enableAtAddress(true)` here ?

apps/remix-ide/src/app/tabs/runTab/contractDropdown.js
LianaHus: yes, I refactor that later and fergot, thanks

apps/remix-ide/src/app/tabs/runTab/contractDropdown.js
```diff
@@ -21,21 +21,22 @@ class ContractDropdownUI {
...
       if (success) {
-        this.selectContractNames.removeAttribute('disabled')
+        this.enableContractNames(true)
         this.dropdownLogic.getCompiledContracts(compiler, compilerFullName).forEach((contract) => {
-          contractNames.appendChild(yo`<option value="${contract.name}" compiler="${compilerFullName}">${contract.name} - ${contract.file}</option>`)
+          this.selectContractNames.appendChild(yo`<option value="${contract.name}" compiler="${compilerFullName}">${contract.name} - ${contract.file}</option>`)
         })
       } else {
-        this.selectContractNames.setAttribute('disabled', true)
+        this.enableContractNames(false, 'sol')
```
yann300, line 37: the second parameter is never used

yann300:
a little inconsistency:
 - compile a contract
 - contract is shown in the run tab
 - switch to a abi file
 - switch back to the solidity file
 - => the contract previously displayed is lost and one has to recompile

yann300:
when using the at Address function , it shows
`<AT ADDRESS> AT 0X5B3...EDDC4 (MEMORY)`
it should perhaps display if possible the name of the file instead of `<AT ADDRESS>`

apps/remix-ide/src/app/tabs/runTab/contractDropdown.js
```diff
@@ -80,14 +81,42 @@ class ContractDropdownUI {
...
   }
 
+  enableContractNames (enable) {
+    if (enable) {
+      this.selectContractNames.removeAttribute('disabled')
+      this.selectContractNames.setAttribute('title', 'Select contract for Deploy or At Address.')
+    } else {
+      this.selectContractNames.setAttribute('disabled', true)
+      if (this.loadType === 'sol') {
+        this.selectContractNames.setAttribute('title', '⚠ Select and compile *.sol file to deploy or access a contract.')
```
yann300, line 91: `⚠ Select and compile *.sol file to deploy or access a contract.` if the user already compiled but switch to abi and then back to the solidity fine, that sounds weird to show him this msg as he actually already compiled.

yann300:
regarding the previous comment:
> => the contract previously displayed is lost and one has to recompile

when switching back to the sol file, the list of compiled contracts should be re-shown. and there 's no need to display `Untitled.abi` in the dropdown list because there is anyway only one entry

LianaHus:
> a little inconsistency:
> 
> * compile a contract
> * contract is shown in the run tab
> * switch to a abi file
> * switch back to the solidity file
> * => the contract previously displayed is lost and one has to recompile

yes, we discussed this behavior. now you want to keep it?

LianaHus:
> regarding the previous comment:
> 
> > => the contract previously displayed is lost and one has to recompile
> 
> when switching back to the sol file, the list of compiled contracts should be re-shown. and there 's no need to display `Untitled.abi` in the dropdown list because there is anyway only one entry

we discussed this as well. in a call.

apps/remix-ide/src/app/tabs/runTab/contractDropdown.js
LianaHus: but he switched. the current file is defining the state everywhere. it wouldn't be reset if autocompile is on

yann300 (approved):

Issue #346, opened 2020-08-11
Migrate libs to typescript


Aniket-Engg:
- [x] remix-lib 
- [x] remix-debug 
- [x] remix-simulator

Aniket-Engg:
All libs are in ts now.

PR #648, opened 2020-11-30
Remix lib ts

Aniket-Engg:
related to #346


libs/remix-astwalker/tests/newTests.ts
```diff
@@ -130,12 +130,12 @@ tape("New ASTWalker", (t: tape.Test) => {
 	 However, for non-typescript uses, we add this test which casts
 	 to an AST to check that there is a run-time check in walkFull.
       */
-      astWalker.walkFull(<AstNode>legacyNode, (node: AstNode) => {
+      astWalker.walkFull(<any>legacyNode, (node: AstNode) => {
```
yann300, line 133: is that fine to have this `<any>` here?

libs/remix-astwalker/tests/newTests.ts
```diff
@@ -130,12 +130,12 @@ tape("New ASTWalker", (t: tape.Test) => {
...
         count += 1;
       });
     }
-    t.throws(badCall, /first argument should be an ast/,
-      "passing legacyAST fails");
+    // t.throws(badCall, /first argument should be an ast/,
```
yann300, line 137: any good reason to remove these tests?

libs/remix-lib/src/execution/eventsDecoder.ts
```diff
@@ -36,29 +38,29 @@ class EventsDecoder {
     this._decodeEvents(tx, receipt.logs, contract, contracts, cb)
   }
 
-  _eventABI (contract) {
-    const eventABI = {}
+  _eventABI (contract): Record<string, unknown> {
```
yann300, line 41: we can have a minimum type here `{ event, inputs, object, abi }` instead of `unknown`

libs/remix-astwalker/tests/newTests.ts
Aniket-Engg: Because when legacyAST is passed we don't show an error to keep it compatible to old compilers

libs/remix-astwalker/tests/newTests.ts
Aniket-Engg: Because a there is type definition mismatch for legacyAST node as we don't support that now

libs/remix-lib/src/execution/eventsDecoder.ts
Aniket-Engg: ok

libs/remix-lib/src/execution/execution-context.ts
```diff
@@ -268,26 +277,26 @@ function ExecutionContext () {
...
-  this.listenOnLastBlock = () => {
+  listenOnLastBlock () {
     this.listenOnLastBlockId = setInterval(() => {
       this._updateBlockGasLimit()
     }, 15000)
   }
 
   // TODO: remove this when this function is moved
-  const self = this
+  // const self = this
```
yann300, line 287: I guess that can be just removed

Aniket-Engg: yes

yann300 (approved):

PR #652, opened 2020-12-02
Listen to the optimism compiler

yann300:
Allow to listen on the optimism compiler plugin , which should be added soon to the directory https://github.com/ethereum/remix-plugins-directory/pull/183
So we can display the contract and deploy it from the run tab


yann300:
@bunsenstraat this allows remix ide to listen on `compilationFinished` from the optimism compiler plugin (so we can display contracts in the run tab and deploy them)

ioedeveloper (approved):

Issue #403, opened 2020-09-07
Remix Desktop: Use offline solidity compilers, plugins resources using for example IPFS


LianaHus:
Use offline solidity compilers, plugins resources using for example IPFS

yann300:
- theme
 - local icons instead of font-awesome (?)
 - solidity compiler
 - connect to IPFS node for plugins

yann300:
depend on https://github.com/ethereum/remix-desktop/pull/25

PR #653, opened 2020-12-02
package asset (theme files)

yann300:
fix https://github.com/ethereum/remix-project/issues/402
partially_fix https://github.com/ethereum/remix-project/issues/403 with "offline compiler"


apps/remix-ide/src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -209,6 +209,7 @@ class CompilerContainer {
     this._view.versionSelector = yo`
       <select onchange="${this.onchangeLoadVersion.bind(this)}" class="custom-select" id="versionSelector" disabled>
         <option disabled selected>${this.data.defaultVersion}</option>
+        <option disabled>builtin</option>
```
LianaHus, line 212: can we put a version here? Is this used for offline one?

apps/remix-ide/src/app/tabs/compileTab/compilerContainer.js
```diff
@@ -499,18 +502,20 @@ class CompilerContainer {
 
   // fetching both normal and wasm builds and creating a [version, baseUrl] map
   async fetchAllVersion (callback) {
-    let allVersions, selectedVersion, allVersionsWasm, isURL
+    let selectedVersion, allVersionsWasm, isURL
+    let allVersions = [{ path: 'builtin', longVersion: 'latest local version' }]
```
LianaHus, line 506: Can we put a exact version here? "Latest local (v0.10.12)"

yann300: we don't know before hand which version will be loaded cause it is the one located in the asset folder, we could hardcode it, we'll need to change it each time we put a new version in the asset

LianaHus (approved):

PR #664, opened 2020-12-08
landing page update

LianaHus:


apps/remix-ide/src/app/ui/landing-page/landing-page.js
```diff
@@ -193,7 +194,7 @@ export class LandingPage extends ViewPlugin {
           href="https://twitter.com/EthereumRemix"
         >
         </a>
-        <script src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
+        <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
```
yann300, line 197: why do you need to load the same script twice?

LianaHus: because I couln't change data-theme

LianaHus: ah you are right

LianaHus: or not. I use update of yo-yo

yann300 (approved):

PR #668, opened 2020-12-09
File explorer

ioedeveloper:


ryestew:
When I try to connect with remixd, with the absolute path to a folder, in the file explorer, I get the **localhost** folder at the root of my computer.

ryestew:
the 4th icon - 
image
is confusing me - I think the text instead of:
Upload Local files to the Browser Storage Explorer

should be:
Load a local file into Remix's browser folder

ryestew:
So its probably beyond the scope of what you are doing for this 1st move to React - but it would be nice if a user could drag a file into and out of a folder.  Its a currently a bit hard to clean up the folders & files ( in master and in this branch).

ryestew:
Creating a file and folder in localhost is great!

ryestew (commented):
the folder icons are too dark in the light theme.  I put filter: invert(0.5); and it looks like this now:

image

ioedeveloper:
> When I try to connect with remixd, with the absolute path to a folder, in the file explorer, I get the **localhost** folder at the root of my computer.

I also experienced this bug after I rebased to master. It is somehow related to @yann300 PR https://github.com/ethereum/remix-project/pull/689/files.

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -44,63 +47,86 @@ const profile = {
...
-    self._deps.fileProviders.localhost.event.register('connecting', (event) => {
+    this.event = new EventManager()
+    // fileExplorer.ensureRoot()
+    this._deps.fileProviders.localhost.event.register('connecting', (event) => {
     })
 
-    self._deps.fileProviders.localhost.event.register('connected', (event) => {
-      fileSystemExplorer.show()
-    })
+    // this._deps.fileProviders.localhost.event.register('connected', (event) => {
```
yann300, line 81: This is handled somewhere else or you don't want anymore these event to apply?

ioedeveloper: They are still valid events. I will update my PR.

bunsenstraat:
If I right click on the icons, not on the names, I don't get a context menu, is that by design?

yann300 (approved):

ioedeveloper:
> If I right click on the icons, not on the names, I don't get a context menu, is that by design?

Yes it is. I had trouble retreiving filename if icon is clicked so I limited it to the text field.

PR #678, opened 2020-12-11
remix-debug: js to ts

Aniket-Engg:
Related to #346


apps/remix-ide/src/app/editor/contextView.js
```diff
@@ -28,7 +27,7 @@ class ContextView {
     this._view = null
     this._nodes = null
     this._current = null
-    this.sourceMappingDecoder = new SourceMappingDecoder()
+    this.sourceMappingDecoder = sourceMappingDecoder
```
yann300, line 30: This is not an instantiation anymore?

Aniket-Engg: no. `SourceMappingDecoder` is not a class anymore

yann300 (approved):

PR #683, opened 2020-12-17
updating images with theme

LianaHus:


ryestew (approved):
First a small point - I think that the Sourcify plugin is too bright.  But if we make a rule that these have a filter: invert(1) in the dark theme of .8 then MythX will be not be contrasty enough.  So we should probably keep it as you have it.  But this is not the focus of this PR - because this filter: invert was happening before.

The reason for this PR is addressing when we use **images** on the homepage other than the images in the Featured plugin buttons  - (the featured plugin buttons already inverted depending on the theme).

This PR does help the remix logo for "Featuring Website" ( a name we should probably change)(note to self... ).   Ideally the color or the lightness of the images would match the text color.  This is the case with the font awesome icons above and below it (see the image below) - and that is because they are icons which match the font color and the remix image is an image.
image

But this PR does improve the situation.  So lets merge it.  
Sorry to be so long winded.

PR #689, opened 2020-12-21
Fix removePrefix from fileProvider

yann300:


GrandSchtroumpf (approved):
Looks like a lot of check for "/" at different places. I fear to might be difficult to follow if an issue occurs. I would advocate for putting all the formatting in one reusable function outside of the class if possible. But I don't have the big picture here, so I'm not sure it's possible...

PR #692, opened 2020-12-22
added icons to the file manager

LianaHus:


LianaHus:
Screenshot from 2020-12-22 11-24-26
Screenshot from 2020-12-22 11-24-11

ryestew:
Its great!  My only question is for plugins like Remixd & GIT- that have no icon - they have nothing.  Which makes sense - but they do look a little bit like a mistake ( albeit a small mistake).  Maybe we should merge this as is.  

Or here are 2 ideas for an alteration :
- Put  NA  ( for not appropriate - or not applicable) if they have no icon.  
- remove the space.

Maybe these types of plugins with no UI need some kind of indication that they have no UI.  So this idea should be in another issue.

LianaHus:
> Its great! My only question is for plugins like Remixd & GIT- that have no icon - they have nothing. Which makes sense - but they do look a little bit like a mistake ( albeit a small mistake). Maybe we should merge this as is.
> 
> Or here are 2 ideas for an alteration :
> 
> * Put  NA  ( for not appropriate - or not applicable) if they have no icon.
> * remove the space.
> 
> Maybe these types of plugins with no UI need some kind of indication that they have no UI. So this idea should be in another issue.

yeah I know. but it's kind of right so user will not search for it. we can add later a special icon though

ryestew (approved):

PR #714, opened 2021-01-05
canActivate option

bunsenstraat:
I've added an option so that plugins can activate other plugins by adding an extra optional line to the profiles
canActivate: ['plugin1', 'plugin2', ... ]
This will allow that plugin to activate other plugins. But we can curate this when the profiles are made.
This optional. If the other plugin does not exist or the caller doesn't have canActivate in its profile it gets caught.


apps/remix-ide/src/remixAppManager.js
```diff
@@ -15,8 +15,8 @@ export function isNative (name) {
   return nativePlugins.includes(name) || requiredModules.includes(name)
 }
 
-export function canActivate (name) {
-  return ['ethdoc'].includes(name) || isNative(name)
+export function canActivate (from, to) {
+  return ['ethdoc'].includes(from.name) || isNative(from.name) || (typeof to !== 'undefined' ? typeof from.canActivate !== 'undefined' ? from.canActivate.includes(to.name) : false : false)
```
yann300, line 19: Could the new test be put as a new line and with maybe a comment. 
Just matter of better readability

bunsenstraat: sure

bunsenstraat: > is `typeof to !== 'undefined' ` really needed?
> same for `typeof from.canActivate`

if you provide a plugin name to call that does not exist to is undefined
canactivate is optional, so it should be checked

bunsenstraat:
> Ok that will give us better flexibility, but this might need to be declared in remix-plugin as well. (I guess at least for defining `canActivate` as an optional parameter)

I agree. I Will work on it.

Aniket-Engg (approved):

Aniket-Engg:
May be you should rebase it. Not sure why changes from https://github.com/ethereum/remix-project/pull/711 are appearing in this PR too.

PR #722, opened 2021-01-11
fix publish to docker

yann300:


Jakic007:
Hi @yann300 
Can you with this: `sed 's/[#\/]/-/g' <<< $CIRCLE_BRANCH`
That should replace both slash and hash char with dash char.

Aniket-Engg (approved):

PR #725, opened 2021-01-11
changed focus methods

bunsenstraat:
the focus method is exposed through the plugin, so I added the super method so we can actually use it. Before that, on the side panel it was only used to emit an event that triggered highlighting being cleared or something. Let's discuss.


yann300:
what action does the `super.focus`?

bunsenstraat:
> what action does the `super.focus`?

```
  showContent (name) {
    if (!this.contents[name]) throw new Error(`Plugin ${name} is not yet activated`)
    // hiding the current view and display the `moduleName`
    if (this.active) {
      this.contents[this.active].style.display = 'none'
    }
    this.contents[name].style.display = 'flex'
    this.active = name
  }

  focus (name) {
    this.showContent(name)
  }
  ```
  
  on panel

yann300 (approved):

ioedeveloper (approved):

Issue #535, opened 2020-10-30
Debugger: show error in UI when tx number is invalid


LianaHus:
we have js error now in the console

Aniket-Engg:
Also, if in `Deploy and Run`, Environment is Javascript VM and you put a testnet transaction hash, it doesn't show anything on UI, neither starts the debugging. In console, there is an error.

May be in general, error message should also contain the text like `make sure you have selected right environment`

PR #738, opened 2021-01-14
error management in debugger

LianaHus:
fix https://github.com/ethereum/remix-project/issues/535


apps/remix-ide/src/app/tabs/debugger-tab.js
```diff
@@ -61,6 +62,10 @@ export class DebuggerTab extends DebuggerApiMixin(ViewPlugin) {
     return this.el
   }
 
+  showMessage (title, body) {
+    modalDialogCustom.alert(title, body)
```
yann300, line 66: This need to be sanitize, to disallow injecting script.

apps/remix-ide-e2e/src/tests/debugger.test.ts
```diff
@@ -183,6 +183,7 @@ module.exports = {
     .getEditorValue((content) => {
       browser.assert.ok(content.indexOf('if slt(sub(dataEnd, headStart), 32) { revert(0, 0) }') != -1, 'current displayed content is not a generated source')
     })
+    .modalFooterOKClick()
```
yann300, line 186: why do you need to do that?

LianaHus: I had to because of error of fetching. now I'll remove it

libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
```diff
@@ -8,6 +8,7 @@ import { DebuggerUIProps } from './idebugger-api' // eslint-disable-line
 import { Toaster } from '@remix-ui/toaster' // eslint-disable-line
 /* eslint-disable-next-line */
 import './debugger-ui.css'
+const helper = require('../../../../../apps/remix-ide/src/lib/helper')
```
ioedeveloper, line 11: We can make this a standalone lib that can be imported and made use of by any app/lib.

ioedeveloper (approved):

PR #740, opened 2021-01-14
Migrate runtab to viewplugin, method getSettings

bunsenstraat:
This is to expose methods to the plugin system. In this case getSettings and probably in the future more events and other data.


apps/remix-ide/src/app/udapp/run-tab.js
```diff
@@ -44,6 +45,31 @@ export class RunTab extends LibraryPlugin {
     this.compilersArtefacts = compilersArtefacts
     this.networkModule = networkModule
     this.fileProvider = fileProvider
+    this.createVMAccount = this.pluginUDapp.createVMAccount
```
yann300, line 48: if PluginUdapp is not used elsewhere, I think just removing it would be better

bunsenstraat: Are you saying createVMAccount and the other methods are not used?

bunsenstraat: Are you saying createVMAccount and the others methods are not used? or just that the class is not needed.

yann300: no, just the `PluginUdapp` (the class is not needed)

apps/remix-ide/src/app/udapp/run-tab.js
```diff
@@ -44,6 +44,43 @@ export class RunTab extends LibraryPlugin {
...
+      this.emit('newTransaction', tx, receipt)
+    })
+  }
+
+  getSettings () {
+    return new Promise((resolve, reject) => {
+      if (!this.container) reject(new Error('UI not ready'))
+      else {
+        resolve({
+          selectedAccount: this.container.querySelector('#txorigin').selectedOptions[0].value,
```
yann300, line 61: could you replace `.container.querySelector('#txorigin')` by call to `this.settingsUI`

yann300: same for the others settings

PR #743, opened 2021-01-15
Plugin API: support terminal plugin interface

dev-ion:
add new terminal api to **plugin-api** ([https://github.com/ethereum/remix-plugin/pull/313](url)).

```
export interface ITerminal {
  events: {} & StatusEvents;
  methods: {
    logTx(tx: Transaction): void;
    logHtml(msg: string): void;
  };
}
```

스크린샷 2021-01-15 오후 5 50 24
< example use of Klaytn Network (not Ethereum, but forked from Ethereum) ><br><br>
They can be used in plugins with custom network. (ex. Quorum Plugin)

remix-plugin PR here: [https://github.com/ethereum/remix-plugin/pull/313](url)


GrandSchtroumpf:
As explain in the other PR I would rather have a more generic method `log(message)` that would be easier to extends.
Some ideas: 
- `log(message, type)`
- `log(message, params)`: params would support "type" and other things for later updates
- `log({ message, type })`
- `log(message)`: The ide would look for the format of the message (mandatory field) and decide how to show it.

dev-ion:
> As explain in the other PR I would rather have a more generic method `log(message)` that would be easier to extends.
> Some ideas:
> 
> * `log(message, type)`
> * `log(message, params)`: params would support "type" and other things for later updates
> * `log({ message, type })`
> * `log(message)`: The ide would look for the format of the message (mandatory field) and decide how to show it.

Thanks!
I'll specify those methods and mention you again later.

bunsenstraat:
I think we can convert this one to draft then? While we await a new proposal?

bunsenstraat:
A general terminal API is now provided
https://github.com/ethereum/remix-plugin/pull/371

PR #748, opened 2021-01-18
Update the size of shadowbox

LianaHus:


ryestew (commented):
When I run the branch - in the background I get:

<div class="modal-content modalContent_3hqkfp undefined"> 
so .remixModalContent isn't showing up.

LianaHus:
> When I run the branch - in the background I get:
> 
> so .remixModalContent isn't showing up.

are you sure it's related? i just changed one value, can it be from new FE?

ioedeveloper (approved):

PR #752, opened 2021-01-19
Context Menu For Plugins

ioedeveloper:


apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -34,7 +34,7 @@ var canUpload = window.File || window.FileReader || window.FileList || window.Bl
 const profile = {
   name: 'fileExplorers',
```
yann300, line 35: this new API will be public, so we should take the new very carefully. cc @bunsenstraat

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -96,6 +97,20 @@ module.exports = class Filepanel extends ViewPlugin {
...
   }
 
+  /**
+   *
+   * @param item { name: string, type?: string[], path?: string[], extension?: string[], pattern?: string[] }
+   * @param callback (...args) => void
+   */
+  registerContextMenuItem (item, callback) {
+    if (!item.name || !callback) return console.error('menu name and callback is mandatory')
+    if (!item.type && !item.path && !item.extension && !item.pattern) return console.error('invalid file matching criteria provided')
```
yann300, line 107: you could also throw an error here, so it bounces back to the caller

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -96,6 +97,20 @@ module.exports = class Filepanel extends ViewPlugin {
     return this.el
   }
 
+  /**
+   *
+   * @param item { name: string, type?: string[], path?: string[], extension?: string[], pattern?: string[] }
+   * @param callback (...args) => void
+   */
+  registerContextMenuItem (item, callback) {
```
yann300, line 105: Passing a callback might not work in the context of an iframe... (e.g I don't think that `function` is a valid parameter for a remix api
cc @bunsenstraat

yann300 (approved):

PR #763, opened 2021-01-20
Update fileManager.js

bunsenstraat:
first discard the file, then call fileRemoved which causes the unselectCurrentFile to resave the file.


ioedeveloper (approved):

Aniket-Engg (commented):
It is deleting the file fine but while deleting the directory after deletion, it says `No such file or directory` in the toaster

Aniket-Engg:
Also I find something strange with `tests` directory. When I delete it, it got deleted with all of its files. Further, when I reload IDE, `tests` directory is back with no files in it.

Issue #816, opened 2021-01-30
Debugging a complex input - console errors


ryestew:
I was testing this PR about complex inputs.
in this contract:
```
// SPDX-License-Identifier: MIT

pragma solidity >=0.6.0 <0.8.0;
pragma abicoder v2;


contract Biz {
    
    uint256 nummy;
    
    struct Todo {
        string text;
        uint numb;
    }
    
    function complexIn(Todo memory _todoNow ) public {
        nummy = _todoNow.numb;
        
    }
}
```
After deploying I go to the **complexin** function and input this:
["Fabulous",3]

When I drag the slider a bit I get this error in the console repeated a lot:

```
main.js:620 Uncaught TypeError: Cannot read property 'lastStep' of undefined
    at InternalCallTree.findScope (main.js:620)
    at InternalCallTree.retrieveFunctionsStack (main.js:624)
    at VmDebuggerLogic.<anonymous> (main.js:450)
    at EventManager.trigger (main.js:549)
    at Debugger.<anonymous> (main.js:478)
    at EventManager.trigger (main.js:549)
    at DebuggerStepManager.jumpTo (main.js:526)
    at setValue (main.js:2049)
    at handleChange (main.js:2049)
    at HTMLUnknownElement.callCallback (vendor.js:211592)
```

and if I put 
["Fabulous","3"]
it updates a bit differently - still errors though.

PR #820, opened 2021-02-01
Fix getting parentScope while debugging

yann300:
fix https://github.com/ethereum/remix-project/issues/816


ryestew (commented):
This fixes the problem - but the source hightlighting is not updating correctly.

ryestew (approved):
Source highlighting works now.

PR #858, opened 2021-02-09
reset staticAnalysis after deactivation

### LINKED ISSUE #803: Static analysis is not receiving compilationfinished after reactivation
steps to reproduce: 

activate SA and Compiler.
open a file. add 2 vars with similar names(number and number) for example
compile. 
see the warning
now deactivate and activate again.
compile.

no warning

Aniket-Engg:
fixes #803


apps/remix-ide/src/app/tabs/analysis-tab.js
```diff
@@ -28,6 +28,9 @@ class AnalysisTab extends ViewPlugin {
   }
 
   render () {
+    const listOfPlugins = localStorage.getItem('workspace')
```
yann300, line 31: is it the only way to do it? accessing the localstorage directly here sounds risky

yann300: and the name `solidityStaticAnalysis` is not really tied to the view itself

yann300 (approved):

Issue #687, opened 2020-12-19
Annotations disappear when changing file


EdsonAlcala:
Hello,

I'm working in a plugin that adds annotations to the editor and when changing files the annotations disappear.

![remix-issue](https://user-images.githubusercontent.com/3077635/102688380-4cb6c380-41ee-11eb-8a9f-449974f5720a.gif)

EdsonAlcala:
By the way this happens when the auto compile option is enabled. I guess the compiler does a clear annotations that removes all annotations. Would it be possible to not remove annotations created by another plugin?

Thanks

PR #876, opened 2021-02-15
Improve annotations

yann300:
fix https://github.com/ethereum/remix-project/issues/687
fix https://github.com/ethereum/remix-project/issues/787


bunsenstraat (approved):
tested it with testing plugins to see if annotation stay when switching between plugins and tab. clearing annotations also works fine. anything else I need to test @yann300 ?

yann300:
> clearing annotations also works fine. anything else I need to test 

making sure decoration are removed when the plugin is deactivated, but I think you already did.

PR #906, opened 2021-02-18
Display icons based on file types.

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/198

### DIFF
--- a/apps/remix-ide/src/lib/helper.js
+++ b/apps/remix-ide/src/lib/helper.js
@@ -68,7 +68,14 @@ module.exports = {
     const hexValue = hash.slice(2, hash.length)
     return this.is0XPrefixed(hash) && /^[0-9a-fA-F]{64}$/.test(hexValue)
   },
-  find: find
+  find: find,
+  getPathIcon (path) {
+    return path.endsWith('.txt')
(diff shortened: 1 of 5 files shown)


apps/remix-ide/src/lib/helper.js
```diff
@@ -68,7 +68,16 @@ module.exports = {
     const hexValue = hash.slice(2, hash.length)
     return this.is0XPrefixed(hash) && /^[0-9a-fA-F]{64}$/.test(hexValue)
   },
-  find: find
+  find: find,
+  getPathIcon (path) {
+    return path.endsWith('.txt')
```
yann300, line 73: that would be perhaps more readable with a `switch` `case`

yann300 (approved):

Aniket-Engg (approved):
Change set looks fine. It would be great if some tests can be added for this.

apps/remix-ide/src/lib/helper.js
ioedeveloper: A switch statement will not work so well for `path.endsWith()` because the argument for switch will need to match a case, which we don't have in this case.

PR #918, opened 2021-02-22
fixed saving matomo choice in localstorage

LianaHus:


apps/remix-ide/src/app/tabs/settings-tab.js
```diff
@@ -199,29 +205,35 @@ module.exports = class SettingsTab extends ViewPlugin {
...
       elementStateChanged(self._view.personalLabel, isChecked)
     }
+
     function onchangeMatomoAnalytics (event) {
-      elementStateChanged(self._view.useMatomoAnalyticsLabel, event.target.checked)
-      const _paq = window._paq = window._paq || []
-      if (this.checked) {
+      const isChecked = self.config.get('settings/matomo-analytics')
+
+      self.config.set('settings/matomo-analytics', !isChecked)
```
yann300, line 234: out of curiosity, any reason you need to save `!isChecked` instead of `isChecked`

LianaHus: to be honest, the other way around would be probably better but it was done like that for other choices I decided to not brake consistency

yann300 (approved):

PR #923, opened 2021-02-24
Resolver engine is removed

LianaHus:


yann300 (approved):

yann300:
it should probably be removed from the github import modal dialog

PR #933, opened 2021-03-02
Throw if no workspace

yann300:
This throws an exception while using file manager and file provider if no workspace has been set.

### DIFF
--- a/apps/remix-ide/src/app/ui/modal-dialog-custom.js
+++ b/apps/remix-ide/src/app/ui/modal-dialog-custom.js
@@ -5,7 +5,7 @@ var css = require('./styles/modal-dialog-custom-styles')
 module.exports = {
   alert: function (title, text) {
     if (text) return modal(title, yo`<div>${text}</div>`, null, { label: null })
-    return modal('alert', yo`<div>${title}</div>`, null, { label: null })
+    return modal('Alert', yo`<div>${title}</div>`, null, { label: null })
   },
   prompt: function (title, text, inputValue, ok, cancel, focus) {
     return prompt(title, text, false, inputValue, ok, cancel, focus)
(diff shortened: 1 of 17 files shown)


apps/remix-ide/src/lib/gist-handler.js
```diff
@@ -60,6 +60,8 @@ function GistHandler (_window) {
           if (!errorLoadingFile) {
             const provider = fileManager.getProvider('workspace')
             provider.lastLoadedGistId = gistId
+          } else {
+            modalDialogCustom.alert(errorLoadingFile.message || errorLoadingFile)
```
LianaHus, line 64: Please add a title to the modal, looks kind of unfinished without.

yann300: I can update those, but most of the custom alert doesn't set any title..

LianaHus: I know. I'm putting everywhere now. Looks better

apps/remix-ide/src/app/ui/modal-dialog-custom.js
```diff
@@ -5,7 +5,7 @@ var css = require('./styles/modal-dialog-custom-styles')
 module.exports = {
   alert: function (title, text) {
     if (text) return modal(title, yo`<div>${text}</div>`, null, { label: null })
-    return modal('', yo`<div>${title}</div>`, null, { label: null })
+    return modal('alert', yo`<div>${title}</div>`, null, { label: null })
```
LianaHus, line 8: capital?

Aniket-Engg (approved):

ioedeveloper (approved):

PR #949, opened 2021-03-10
Added loader for localhost

ioedeveloper:


apps/remix-ide/src/app/files/remixd-handle.js
```diff
@@ -95,6 +95,7 @@ export class RemixdHandle extends WebsocketPlugin {
           label: 'Connect',
           fn: () => {
             try {
+              this.locahostProvider.event.trigger('loading')
```
yann300, line 98: normally `event` should be triggered by the class, not from an exernal one.
At least could you have a `preInit()` function in `locahostProvider` and call that one?

ioedeveloper: Okay

yann300 (approved):

Issue #956, opened 2021-03-11
modal for pubishing gist still refers to browser folder


ryestew:
<img width="563" alt="image" src="https://user-images.githubusercontent.com/6050001/110826699-56871a80-8263-11eb-8d61-130d0bcf74f5.png">

PR #968, opened 2021-03-12
Update publish to gist modal with workspace name.

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/956

### DIFF
--- a/libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
+++ b/libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
@@ -559,7 +559,7 @@ export const FileExplorer = (props: FileExplorerProps) => {
   }
 
   const publishToGist = () => {
-    modal('Create a public gist', 'Are you sure you want to publish all your files in browser directory anonymously as a public gist on github.com? Note: this will not include directories.', {
+    modal('Create a public gist', `Are you sure you want to publish all your files in ${name} directory anonymously as a public gist on github.com? Note: this will not include directories.`, {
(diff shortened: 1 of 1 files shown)


libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
```diff
@@ -557,7 +557,7 @@ export const FileExplorer = (props: FileExplorerProps) => {
   }
 
   const publishToGist = () => {
-    modal('Create a public gist', 'Are you sure you want to publish all your files in browser directory anonymously as a public gist on github.com? Note: this will not include directories.', {
+    modal('Create a public gist', `Are you sure you want to publish all your files in ${name} directory anonymously as a public gist on github.com? Note: this will not include directories.`, {
```
ryestew, line 560: I think the word **directory** is wrong and should be replaced with 'workspace'

so currently if my workspace is named Fred it reads: 
Are you sure you want to publish all your files in Fred directory anonymously as a public gist on github.com?

but as a user I don't see the workspace as a directory.  So maybe this is better - (also I moved the word anonymously)

Are you sure you want to anonymously publish all your files in the Fred workspace as a public gist on github.com?

So it would be more like this:
Are you sure you want to anonymously publish all your files in the ${name} workspace as a public gist on github.com? Note: this will not include directories.

ryestew (changes requested):

Aniket-Engg (approved):

ryestew (approved):

Issue #967, opened 2021-03-12
Blank name for files with the same name displayed in editor tabs.


ioedeveloper:
<img width="654" alt="Screen Shot 2021-03-12 at 13 33 17" src="https://user-images.githubusercontent.com/23282059/110940868-dbdbff00-8337-11eb-9acd-aeb39ff88578.png">

Aniket-Engg:
How is a blank file even created?

bunsenstraat:
It's not blank files. It has to do with the tab titles.

PR #974, opened 2021-03-15
Add workspace name to tabs

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/967

### DIFF
--- a/apps/remix-ide/src/app/files/fileManager.js
+++ b/apps/remix-ide/src/app/files/fileManager.js
@@ -612,8 +612,7 @@ class FileManager extends Plugin {
   }
 
   currentWorkspace () {
-    const file = this.currentFile() || ''
-    const provider = this.fileProviderOf(file)
+    const provider = self._deps.filesProviders.workspace
 
     return provider.workspace
   }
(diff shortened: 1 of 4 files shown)


apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -610,6 +610,13 @@ class FileManager extends Plugin {
       if (callback) callback(error)
     })
   }
+
+  getCurrentWorkspace () {
```
LianaHus, line 614: Maybe currentWorkspace() ? Just a suggestion

apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -610,6 +610,13 @@ class FileManager extends Plugin {
       if (callback) callback(error)
     })
   }
+
+  currentWorkspace () {
+    const file = this.currentFile() || ''
+    const provider = this.fileProviderOf(file)
```
yann300, line 616: Here you could just do
`const provider = self._deps.filesProviders.workspace`
because anyway, the localhost provider doesn't have the `workspace` property

ryestew (commented):
Running locally, I got remixd running and was on localhost in the workspaces selectbox.  Then I switched to another workspace (named fred) and I got this error:
Screen Shot 2021-03-16 at 8 14 17 AM

yann300 (approved):

PR #975, opened 2021-03-15
Compilerconfig

bunsenstraat:
* Get the compiler configuration
   * This function is used by remix-plugin compiler API.
   * @return {object} settings {version, evmVersion, optimize, runs, version, language}


apps/remix-ide/src/app/tabs/compile-tab.js
```diff
@@ -269,6 +269,22 @@ class CompileTab extends ViewPlugin {
     })
   }
 
+  /**
+   * Get the compiler configuration
+   * This function is used by remix-plugin compiler API.
+   * @return {object} settings {version, evmVersion, optimize, runs, version, language}
+   */
+
+  getCompilerConfig () {
```
yann300, line 278: isn't there a duplication with `getCurrentCompilerConfig`

yann300: the function is already existing, perhaps we can just expose it..

yann300:
I think this can be closed, let me know otherwise

PR #988, opened 2021-03-18
Solidity Unit testing plugin improvements

Aniket-Engg:


LianaHus:
disable of 'create' is not working correctly when one change the name but then redo changes
tests -> tests/a -> tests

yann300 (approved):

yann300:
here some results for manual testing:

 1)
  - input `to` in the test directory and click `generate`, the file is generated correclty. 
  - input `to/` will create the path `default_workspace/to//newFile_test.sol`. which seems wrong.
  - switch to the file explorer and a folder with a blank folder has been created. reload remix and the folder name seems to be displayed correctly.
 
 2) the `Create` button was in my case always disabled..

PR #990, opened 2021-03-20
dgit update

bunsenstraat:
test experimental branch for dgit workspaces
manual activation from the plugin manager is needed for testing:
DECENTRALIZED GIT
then add the plugin from https://dgit2-remix.web.app


apps/remix-ide/src/remixAppManager.js
```diff
@@ -111,6 +111,17 @@ export class RemixAppManager extends PluginManager {
     try {
       const res = await fetch(this.pluginsDirectory)
       plugins = await res.json()
+      plugins.push({
```
yann300, line 114: it is a native plugin, would it be possible to  register and activate it the same way we do for native plugin (see app.js for some code examples)

yann300: And maybe you'd like to add it to `requiredPlugins` of remixAppManager.js cause it shouldn't be seen from the plugin manager (or should it?)

bunsenstraat: This was just to make it easy to test. No reason to make it native. It doesn’t actually do a lot. It’s the dgitprovider which is native and does all the work

apps/remix-ide/src/app/files/dgitProvider.js
```diff
@@ -0,0 +1,398 @@
...
+      protocol: 'https',
+      ipfsurl: 'https://ipfs.io/ipfs/'
+    }
+  }
+
+  async getGitConfig () {
+    const workspace = await this.call('filePanel', 'getCurrentWorkspace')
+    return {
+      fs: window.remixFileSystem,
+      dir: `.workspaces/${workspace.name}`
```
yann300, line 48: we could do without hardcoding the value. see https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/files/workspaceFileProvider.js#L10

yann300 (approved):

apps/remix-ide/src/app/files/dgitProvider.js
bunsenstraat: ok I handled it.

apps/remix-ide/src/remixAppManager.js
bunsenstraat: it's gone

LianaHus (approved):

Issue #1022, opened 2021-03-24
0.11.0 Release breaks gists


alexroan:
Opening gists in remix via a URL (like this: [`https://remix.ethereum.org/#version=soljson-v0.6.7+commit.b8d736ae.js&optimize=false&evmVersion=null&gist=0c5928a00094810d2ba01fd8d1083581`](https://remix.ethereum.org/#version=soljson-v0.6.7+commit.b8d736ae.js&optimize=false&evmVersion=null&gist=0c5928a00094810d2ba01fd8d1083581)) is broken as of the 0.11.0 release.

----------------

Error message:

Gist load error
No workspace has been opened.

----------------

The issue is resolved once the default workspace is generated, but that must be done by navigating to remix.ethereum.org (without params).

This is a UX problem for us as we use gists to direct users from our [docs](https://docs.chain.link/docs/get-the-latest-price#solidity) and will be a frustrating stumbling block for beginners.

dmitrydao:
Cannot agree more on this one. Cleaning the cache or going to the Remix website without gists attached in the link sets up a default workspace but a lot of people will find it frustrating that it doesn't work for them if they've used Remix already.

It would be great if a default workspace is created automatically despite the extra arguments in the link.
Thanks in advance.

pappas999:
Agreed. Creating a default workspace if one doesn't exist when opening these gists would solve the issue, and improve user experience for first time users of remix

dmitrydao:
Thank you so much! It works.
Appreciate you understood the importance of this one.

PR #1023, opened 2021-03-24
Keep gist in a separate workspace at load

yann300:
fix https://github.com/ethereum/remix-project/issues/1022


ryestew (approved):
When I ran this and the file explorer came up it was clear I had to click on the file to see the file in the editor.  It would be a bit better if the 1st GIST file ( I mean if it is a gist with many files in it ) come up in the editor.  But it was clear enough when I was checking with a gist loading from the URL.

Issue #117, opened 2019-07-22
Remove all "this.events.emit()" in plugin by "this.emit()"


GrandSchtroumpf:
To avoid breaking changes we duplicate `this.events.emit()` and `this.emit()` in plugins.
We should only keep `this.emit()` in the end.

PR #1037, opened 2021-03-25
removed old way of eventlistener usage

LianaHus:
fix https://github.com/ethereum/remix-project/issues/117


bunsenstraat:
if a plugin calls on, the first param is the plugin emitting the event
```
    this.on("fileManager", "fileRemoved", function (x) {
      console.log("REMOVE", x);
      me.log(x);
    });
```
so if you do

```
    fileManager.on('filesAllClosed', () => {
      this.call('manager', 'activatePlugin', 'home')
      this._view.filetabs.active = 'home'
    })
```
You are telling filemanager to listen to an event on the plugin 'filesAllClosed'. So this won't work.

apps/remix-ide/src/app/panels/main-view.js
```diff
@@ -65,20 +65,20 @@ export class MainView {
       self._view.mainPanel.style.display = 'none'
       self._components.contextView.show()
     })
-    self.tabProxy.event.on('openFile', (file) => {
```
bunsenstraat, line 68: we can't call ON here, because the tabproxy plugin is not registered yet. plugin won't allow that.... what to do? same for all other on calls, can only be made after the plugin is loaded into the engine.

LianaHus:
freezing this for now

PR #1041, opened 2021-03-26
Optimise e2e

ioedeveloper:


ioedeveloper:
We previously disabled tests for goerli testnet and mainnet https://github.com/ethereum/remix-project/blob/master/apps/remix-ide-e2e/src/tests/runAndDeploy.ts#L76. Should i enable them back in this PR?

yann300:
we could also run `publish (to ipfs and docker)` in parrallel with the e2e tests.

yann300:
> Should i enable them back in this PR?

if they are passing now yes, if not the issue I think was the metamask version  we use, so maybe we can tackle that another time

yann300 (approved):

PR #1072, opened 2021-04-02
alert user when full while migrating

bunsenstraat:
test it by doing this a couple of times in your console
```
  let r = await fetch('https://raw.githubusercontent.com/ethereum/remix-project/179c9634fb7469d9799a93de6b7dfa4aabc982ed/package-lock.json')
  window.remixFileSystem.writeFileSync(`${Date.now()}.json`, await r.text())
```

it will fill up the storage, after 2 times or so. then run the migration.

this will give you how much is used, after 9mb it starts to fail

```
var _lsTotal=0,_xLen,_x;for(_x in localStorage){ if(!localStorage.hasOwnProperty(_x)){continue;} _xLen= ((localStorage[_x].length + _x.length)* 2);_lsTotal+=_xLen; console.log(_x.substr(0,50)+" = "+ (_xLen/1024).toFixed(2)+" KB")};console.log("Total = " + (_lsTotal / 1024).toFixed(2) + " KB");

```

### DIFF
--- a/apps/remix-ide/src/migrateFileSystem.js
+++ b/apps/remix-ide/src/migrateFileSystem.js
@@ -1,6 +1,6 @@
 import { Storage } from '@remix-project/remix-lib'
 import { joinPath } from './lib/helper'
-
+var modalDialogCustom = require('./app/ui/modal-dialog-custom')
 /*
   Migrating the files to the BrowserFS storage instead or raw localstorage
 */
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/migrateFileSystem.js
```diff
@@ -53,7 +53,11 @@ const populateWorkspace = async (workspace, json, browserProvider) => {
       browserProvider.createDir(joinPath(workspace, item))
       await populateWorkspace(workspace, json[item].children, browserProvider)
     } else {
-      await browserProvider.set(joinPath(workspace, item), json[item].content)
+      await browserProvider.set(joinPath(workspace, item), json[item].content, (err) => {
```
yann300, line 56: what is the typical message that is returned when the local storage is full?

bunsenstraat: Screenshot 2021-04-06 at 12 13 48

yann300: is ti possible to stop the migration if that happens?

bunsenstraat: it's a recursive script... the old files just stay there right? It can capture any error, it could also be the storage is not full when migration starts. If storage is full files get created but they will be empty if the result would surpass the storage limit.

bunsenstraat: So? Is it good enough or not? @yann300

yann300 (approved):

PR #1073, opened 2021-04-02
fix FE not rendering and display alert

bunsenstraat:


yann300 (approved):

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -13,7 +13,7 @@ var globalRegistry = require('../../global/registry')
 var examples = require('../editor/examples')
 var GistHandler = require('../../lib/gist-handler')
 var QueryParams = require('../../lib/query-params')
-
+var modalDialogCustom = require('../ui/modal-dialog-custom')
```
LianaHus, line 16: could you please use const instead?

bunsenstraat: ok

yann300:
step for testing the feature:
 - copy 0.10.10 assets to the asset folder of the platform being tested.
 - browse 0.10.10, remove the `.workspace` folder, keep some folders and files at the root.
 - browse 0.11.3, the modal should appear.
 - migrate and reload 0.11.3, files should have been migrated.

Issue #186, opened 2020-04-24
Static Analysis: add a checkbox to checkAll to each categorie


LianaHus:
add a checkbox near the header of modules

PR #1104, opened 2021-04-10
refactor: static analyser plugin from yo-yo to react

tizah:
**[ Definition of Done ]** :

- [x] The static analysis plugin, currently built with yo-yo framework should be re-written in Reactjs.
### Additional Improvements
- [x] Change default checkbox to custom checkbox used in settings tab and terminal.
- [x] Make title of category content bold and increase space below description.
- [x] Make panels separately collapsible and remove accordion
- [x] Add select all for each category in static analysis tab. 
### Related issues
https://github.com/ethereum/remix-project/issues/186
https://github.com/ethereum/remix-project/issues/212 

https://www.loom.com/share/fff9abaa363f4f509ac6294b2a837d86


ioedeveloper:
We no longer need the old static analyser module. Remove it.

apps/remix-ide/src/app/tabs/analysis-tab.js
```diff
@@ -1,9 +1,11 @@
+import React from 'react' // eslint-disable-line
 import { ViewPlugin } from '@remixproject/engine-web'
+import ReactDOM from 'react-dom'
 import { EventEmitter } from 'events'
+import {RemixUiStaticAnalyser} from '@remix-ui/static-analyser' // eslint-disable-line
 import * as packageJson from '../../../../../package.json'
+var Renderer = require('../ui/renderer')
```
ioedeveloper, line 7: Is `Renderer` still needed?

libs/remix-ui/static-analyser/src/lib/ErrorRenderer.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useState } from 'react' //eslint-disable-line
+
+interface ErrorRendererProps {
+  message: any;
+  opt: any,
+  warningErrors: any
+}
+
+const ErrorRenderer = ({ message, opt }: ErrorRendererProps) => {
+  const [, setError] = useState(
```
ioedeveloper, line 10: I don't think this is needed since we don't access the state variables.

libs/remix-ui/static-analyser/src/lib/ErrorRenderer.tsx
```diff
@@ -0,0 +1,76 @@
...
+    {
+      row: null,
+      column: null,
+      text: null,
+      type: null,
+      errFile: null
+    }
+  )
+  const getPositionDetails = (msg: any) => {
+    const result = { } as any
```
ioedeveloper, line 20: We can give result a better type. (Record<string, int | string>)

libs/remix-ui/static-analyser/src/lib/Checkbox/StaticAnalyserCheckedBox.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react' //eslint-disable-line
+
+interface StaticAnalyserCheckBoxProps {
```
ioedeveloper, line 3: We can extract this checkbox as a custom checkbox reusable component.

ioedeveloper (changes requested):

libs/remix-ui/static-analyser/src/lib/Button/StaticAnalyserButton.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'  //eslint-disable-line
+
+interface StaticAnalyserButtonProps {
+  onClick: (event) => void
+  buttonText: string,
+  disabled?: boolean
+}
+
+const StaticAnalyserButton = ({
```
ioedeveloper, line 9: Let's also make this button a reusable component outside static analyser component.

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
```diff
@@ -0,0 +1,383 @@
...
+    } else {
+      setAutoRun(true)
+    }
+
+    return () => { }
+  }, [autoRun, categoryIndex])
+
+  const run = (lastCompilationResult, lastCompilationSource, currentFile) => {
+    // const highlightLocation = async (location, fileName) => {
+    //   await props.analysisModule.call('editor', 'discardHighlight')
```
ioedeveloper, line 114: Editor highlighting is missing in your changes. Please include it and add test for highlighting.

ioedeveloper (changes requested):

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
```diff
@@ -0,0 +1,383 @@
...
+
+          }
+        )
+      }
+    } else {
+      setAutoRun(true)
+    }
+
+    return () => { }
+  }, [autoRun, categoryIndex])
```
ioedeveloper, line 110: Do you want to activate autoRun if categoryIndex changes?

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
```diff
@@ -0,0 +1,383 @@
...
+                lastCompilationSource.sources,
+                lastCompilationResult.sources
+              )
+              row = location.start.line
+              column = location.start.column
+              locationString = row + 1 + ':' + column + ':'
+              fileName = Object.keys(lastCompilationResult.contracts)[file]
+            }
+            warningCount++
+            const msg = `
```
ioedeveloper, line 160: Can we use a different implementation for this and not use string to represent JSX?

libs/remix-ui/static-analyser/src/lib/ErrorRenderer.tsx
tizah: will be removed, but in future when we need to handle `errors` and not `warning` it will be useful.

libs/remix-ui/static-analyser/src/lib/ErrorRenderer.tsx
tizah: `type` added

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
tizah: no. have removed it

LianaHus:
- [ ] `Select all` of the sections are not aligned with labels (removing pt-1 from div should fix it)
- [ ] The top-level Select all doesn't work for me. I unselected sections Select all and tried to click on top-level one but it is not working (we need to check if we have e2e tests)
- [ ] Autorun doesn't work as well (we need to check if we have e2e tests)
- [ ] Please remove the right border of the leftmost triangle to unfold/fold the section
- [ ] Please make `Select all`, `Autorun` and `Run` buttons same width
- [ ] We need to remove extra space, everything looks zoomed in. For example bottom margins from elements inside each check(staticanalysismodule*, H6, label...). also mx-4 from div to move it a bit left (we can have a call)
- [ ] onload all sections{or all but one) should be closed/folded

LianaHus (changes requested):
UI Review:
- [x] `Select all` of the sections are not aligned with labels (removing pt-1 from div should fix it)
- [x] The top-level Select all doesn't work for me. I unselected sections Select all and tried to click on top-level one but it is not working (we need to check if we have e2e tests)
- [x] Autorun doesn't work as well (we need to check if we have e2e tests)
- [x] Please remove the right border of the leftmost triangle to unfold/fold the section
- [x] Please make `Select all`, `Autorun` and `Run` buttons have same width
- [x] We need to remove extra space, everything looks zoomed in. For example bottom margins from elements inside each check(staticanalysismodule*, H6, label...). also mx-4 from div to move it a bit left (we can have a call)
- [x] onload all sections{or all but one) should be closed/folded

apps/remix-ide/src/app/tabs/analysis-tab.js
tizah: In some we still need the render, because once the section is commented out or removed the plugin throws an error, i am yet to understand the cause of this behavior.

libs/remix-ui/static-analyser/src/lib/Checkbox/StaticAnalyserCheckedBox.tsx
tizah: checkbox is now a reusable component

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
tizah: Editor highlighting included

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
tizah: Extracted string into a function

yann300 (approved):

ioedeveloper (approved):

ryestew (changes requested):
I noticed that in the results section there is the line:
**last result for:** 
But the file that is getting analyzed does not get filled in. 
image

LianaHus:
> I noticed that in the results section there is the line:
> **last result for:**
> But the file that is getting analyzed does not get filled in.
> image

yes there should be the file name. good catch

tizah:
Added the name of file at result section 
image

ryestew (approved):

LianaHus:
"Please reload Remix after having saved the token." is still there. is this problem solved?

LianaHus (approved):

LianaHus:
> "Please reload Remix after having saved the token." is still there. is this problem solved?

Sorry wrong PR

Issue #1107, opened 2021-04-12
call to deactivate plugin doesn't seem to work


yann300:
calling 
`this.call('manager', 'deactivatePlugin', 'home')` 
doesn't actually deactivate the plugin, although `canDeactivatePlugin` from remixAppManager.je returns `true` for the plugin caller

bunsenstraat:
Where are you calling it from, who is 'this'?

yann300:
`this` was the `tabs` plugin  and it's listed in the required modules

bunsenstraat:
tried it in tabs, works fine. 
so where exactly did you put it?

yann300:
in `tab-proxy.js` 
I updated https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/panels/tab-proxy.js#L131
this get called when a tab which contain a plugin is closed

bunsenstraat:
seems fine to me. can you explain the steps to reproduce the bug?

bunsenstraat:
or make a branch with the bug

yann300:
- activate a mainPanel plugin from the plugin manager
 - close the tab (which contains the plugin)
 - the plugin is still marked as activated in the pluginManager. ( if I use `this.call('manager', 'deactivatePlugin', name)` instead of `this.appManager.deactivatePlugin(name)`

bunsenstraat:
ah so it's not 'home' you were talking about

bunsenstraat:
not this.call('manager', 'deactivatePlugin', 'home')

yann300:
`this.call('manager', 'activatePlugin', 'home')` seems to work

bunsenstraat:
Yes I know, it depends on the plugin. 
Basically you have to see it like this, if you call appmanager directly it executes the method in manager.js, and then the FROM caller is 'manager' the toggleActive happens always. https://github.com/ethereum/remix-plugin/blob/8e494ee10f8539e74cd7ba30b6bddf1cb2ccd025/packages/engine/core/src/lib/manager.ts#L187

If you call this.call('manager'...), it's not the manager that deactivates, it is the plugin calling which is FROM, in this case 'tabs', manager then then calls 'canDeactivate' on the plugin 'name' itself through postmessage.  https://github.com/ethereum/remix-plugin/blob/8e494ee10f8539e74cd7ba30b6bddf1cb2ccd025/packages/engine/core/src/lib/manager.ts#L195

[...]

PR #1110, opened 2021-04-13
Fixmanagerdeact

bunsenstraat:
explanation:

https://github.com/ethereum/remix-project/issues/1107


LianaHus:
if we don't have tests for that it worth to add some

bunsenstraat:
> if we don't have tests for that it worth to add some

well, the test is before this didn't work. it would not deactivate plugins when called through this.call('manager'

bunsenstraat:
> if we don't have tests for that it worth to add some

you mean like real tests?

yann300 (approved):

PR #1111, opened 2021-04-13
createing default workspace

### LINKED ISSUE #1028: handle error on file/folder creation in internal/external plugins
Currently, when internal/external plugins are trying to write/read in File Explorer, when no workspace available, the FileManager is not creating the file/folder but throwing an error.
Our plugins are not handling this error but either doing nothing or printing an error in the console. No feedback to the user.
I see 3 options:

1. handle that it FileManager - simply create a workspace on demand
2. handle that it FileManager - FilManager is auto-creating a new workspace when the last one is deleted
3. update all internal/external plugins that writing a file/folder in File Manager and handle the error case by case

LianaHus:
fix #1028

### DIFF
--- a/apps/remix-ide/src/app/files/fileManager.js
+++ b/apps/remix-ide/src/app/files/fileManager.js
@@ -563,7 +563,6 @@ class FileManager extends Plugin {
     if (file.startsWith('browser')) {
       return this._deps.filesProviders.browser
     }
-    const provider = this._deps.filesProviders.workspace
     return this._deps.filesProviders.workspace
   }
 
--- a/apps/remix-ide/src/app/files/fileManager.js
+++ b/apps/remix-ide/src/app/files/fileManager.js
@@ -564,7 +564,6 @@ class FileManager extends Plugin {
       return this._deps.filesProviders.browser
     }
     const provider = this._deps.filesProviders.workspace
-    if (!provider.isReady()) throw createError({ code: 'ECONNRESET', message: 'No workspace has been opened.' })
     return this._deps.filesProviders.workspace
   }
 
--- a/apps/remix-ide/src/app/files/workspaceFileProvider.js
+++ b/apps/remix-ide/src/app/files/workspaceFileProvider.js
@@ -30,7 +30,7 @@ class WorkspaceFileProvider extends FileProvider {
   }
 
   removePrefix (path) {
-    if (!this.workspace) this.createDefaultWorkspace()
+    if (!this.workspace) this.createWorkspace()
     path = path.replace(/^\/|\/$/g, '') // remove first and last slash
     if (path.startsWith(this.workspacesPath + '/' + this.workspace)) return path
     if (path.startsWith(this.workspace)) return this.workspacesPath + '/' + this.workspace
@@ -51,7 +51,7 @@ class WorkspaceFileProvider extends FileProvider {
   }
 
   resolveDirectory (path, callback) {
-    if (!this.workspace) this.createDefaultWorkspace()
+    if (!this.workspace) this.createWorkspace()
     super.resolveDirectory(path, (error, files) => {
       if (error) return callback(error)
       const unscoped = {}
@@ -76,13 +76,13 @@ class WorkspaceFileProvider extends FileProvider {
   }
 
   _normalizePath (path) {
-    if (!this.workspace) this.createDefaultWorkspace()
+    if (!this.workspace) this.createWorkspace()
     return path.replace(this.workspacesPath + '/' + this.workspace + '/', '')
   }
 
-  createDefaultWorkspace () {
-    this.workspace = 'generated_workspace'
-    this.event.trigger('create_workspace_default', [this.workspace])
+  createWorkspace (name) {
+    if (!name) name = 'default_workspace'
+    this.event.trigger('create_workspace', [name])
   }
 }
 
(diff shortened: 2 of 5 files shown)


Aniket-Engg (approved):

apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -563,8 +563,6 @@ class FileManager extends Plugin {
     if (file.startsWith('browser')) {
       return this._deps.filesProviders.browser
     }
-    const provider = this._deps.filesProviders.workspace
```
yann300, line 566: do you need to remove this?

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -109,6 +117,18 @@ export const Workspace = (props: WorkspaceProps) => {
     }
   }, [])
 
+  const createNewWorkspace = async (workspaceName) => {
```
yann300, line 120: maybe it would make sense to refactor this with `onFinishCreateWorkspace` because they do the same thing.

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -202,14 +202,17 @@ module.exports = class Filepanel extends ViewPlugin {
...
-    const browserProvider = this._deps.fileProviders.browser
-    const workspacesPath = this._deps.fileProviders.workspace.workspacesPath
-    await this.processCreateWorkspace(workspaceName)
-    for (const file in examples) {
-      try {
-        await browserProvider.set('browser/' + workspacesPath + '/' + workspaceName + '/' + examples[file].name, examples[file].content)
-      } catch (error) {
-        console.error(error)
+    else {
+      this._deps.fileProviders.workspace.setWorkspace(workspaceName)
```
yann300, line 206: I don't think you need the `else` here

apps/remix-ide/src/app/files/workspaceFileProvider.js
```diff
@@ -8,6 +9,7 @@ class WorkspaceFileProvider extends FileProvider {
     super('')
     this.workspacesPath = '.workspaces'
     this.workspace = null
+    this.event = new EventManager()
```
yann300, line 12: instead of using `require('../../lib/events')` you could also use `const EventEmitter = require('events')` which is the oob solution. The previous one is no longer used for new features.

apps/remix-ide/src/app/files/workspaceFileProvider.js
```diff
@@ -74,9 +76,14 @@ class WorkspaceFileProvider extends FileProvider {
...
 
   _normalizePath (path) {
-    if (!this.workspace) throw new Error('No workspace has been opened.')
+    if (!this.workspace) this.createWorkspace()
     return path.replace(this.workspacesPath + '/' + this.workspace + '/', '')
   }
+
+  createWorkspace (name) {
+    if (!name) name = 'default_workspace'
+    this.event.trigger('create_workspace', [name])
```
yann300, line 85: name of the events are usually camel case I think `createWorkspace`

apps/remix-ide/src/app/files/fileManager.js
LianaHus: yes we don't need to show that error anymore

apps/remix-ide/src/app/panels/file-panel.js
LianaHus: it is changing nothing just increasing readability of the code

apps/remix-ide/src/app/files/workspaceFileProvider.js
LianaHus: ok

apps/remix-ide/src/app/files/workspaceFileProvider.js
LianaHus: ok

apps/remix-ide/src/app/panels/file-panel.js
yann300, line 206: when no workspace is selected and a plugin try to write to remix, did you make sure the app switch to the new created workspace? cause I am not sure doing `fileProviders.workspace.setWorkspace(workspaceName)` is enough

yann300 (approved):

apps/remix-ide/src/app/files/workspaceFileProvider.js
LianaHus: it is not working with a change.

apps/remix-ide/src/app/panels/file-panel.js
LianaHus: is selected or available?

apps/remix-ide/src/app/files/workspaceFileProvider.js
yann300: you might need to update the function, it should work

yann300: just check where it is used

apps/remix-ide/src/app/panels/file-panel.js
yann300: I meant "selected".
I am not sure doing `this._deps.fileProviders.workspace.setWorkspace(workspaceName)` is the best way to handle that, but that seem to be ok for the moment

apps/remix-ide/src/app/files/workspaceFileProvider.js
yann300: you can find example where it's being used there https://github.com/ethereum/remix-project/search?q=EventEmitter

PR #1128, opened 2021-04-20
Require loading Solidity 0.3.6 with worker

### LINKED ISSUE #1127: Support Solidity 0.3.6
I tried to compile some legacy contracts in Remix and noticed that compiler versions >=0.4.0 worked, but 0.3.6 did not. The loader just kept spinning.

The logs reveal the following:
Screenshot

I think the problem is that since 0.3.6 loads the wasm version it should also use the worker. Probably remix uses the worker for >=0.4.0 compilers, and not for older ones. I suggest to change that rule to >=0.3.6.

axic:
Fixes #1127.

i think this will fix the issue.

### DIFF
--- a/apps/remix-ide/src/app/compiler/compiler-utils.js
+++ b/apps/remix-ide/src/app/compiler/compiler-utils.js
@@ -25,8 +25,10 @@ export function canUseWorker (selectedVersion) {
   const version = semver.coerce(selectedVersion)
   const isNightly = selectedVersion.includes('nightly')
   return browserSupportWorker() && (
+    // All compiler versions (including nightlies) after 0.6.3 are wasm compiled
     semver.gt(version, '0.6.3') ||
-    (semver.gt(version, '0.3.6') && !isNightly)
+    // Only releases are wasm compiled starting with 0.3.6
+    (semver.gte(version, '0.3.6') && !isNightly)
   )
 }
 


apps/remix-ide/src/app/compiler/compiler-utils.js
```diff
@@ -25,8 +25,10 @@ export function canUseWorker (selectedVersion) {
   const version = semver.coerce(selectedVersion)
   const isNightly = selectedVersion.includes('nightly')
   return browserSupportWorker() && (
+    // All compiler versions (including nightlies) after 0.6.3 are wasm compiled
     semver.gt(version, '0.6.3') ||
-    (semver.gt(version, '0.3.6') && !isNightly)
+    // Only releases are wasm compiled starting with 0.3.6
```
axic, line 30: @cameel can you confirm these two comments? :)

apps/remix-ide/src/app/compiler/compiler-utils.js
```diff
@@ -25,8 +25,10 @@ export function canUseWorker (selectedVersion) {
   const version = semver.coerce(selectedVersion)
   const isNightly = selectedVersion.includes('nightly')
   return browserSupportWorker() && (
+    // All compiler versions (including nightlies) after 0.6.3 are wasm compiled
```
cameel, line 28:
```suggestion
    // All compiler versions (including nightlies) starting with 0.6.2 are wasm compiled
```

Not sure if all 0.6.2 nightlies are wasm. I'll have to check that.

apps/remix-ide/src/app/compiler/compiler-utils.js
cameel: We have wasm builds for all releases starting with 0.3.6, [except for 0.5.17](https://github.com/ethereum/solidity/issues/10329).

LianaHus: then we should include 0.5.17 too in the check.
Please check manually for 0.3.6 as I remember when I was adding this it has a problem of `Uncaught JavaScript exception: RangeError: Maximum call stack size exceeded`

cameel: Well, I did test both wasm and non-wasm 0.3.6 back when I was doing the comparison of generated bytecode (between binaries of a given version on all platforms) in the CI. That was using node.js. I did not get errors like this one so this must be a browser thing.

About the check: not sure if it's easy to do that here, but the simplest one would be to assume that whatever is in the `wasm/` dir needs the worker. We do plan to add a wasm build of 0.5.17 eventually so it's better if it's not excluded permanently.

apps/remix-ide/src/app/compiler/compiler-utils.js
LianaHus: those are also wasm Daniel said

cameel: ok. Just checked to confirm and I see that `soljson-v0.6.2+commit.bacdbe57.js` (i.e. the first nightly of the 0.6.2 series) seems to be the the first one compiled to wasm. `soljson-v0.6.1-nightly.2020.1.7+commit.8385256b.js` (last in 0.6.1) was asm.js.

apps/remix-ide/src/app/compiler/compiler-utils.js
LianaHus:
```
contract C {
    struct S {
        S[] arr;
    }

    S s;

    function test() {
        s.arr.length = 10;
    }
}
```

LianaHus: if you check with the code above, we still have Uncaught JavaScript exception: RangeError: Maximum call stack size exceeded. this was fixed when we moved to worker load of wasm builds. this one was an exception and was still throwing the exception so I left it out. I'm not sure why we have errors on loading now

LianaHus: but I guess this is not important as it is not working at all.

cameel: Ok this does fail but it seems to be a bug in the compiler or solc-js rather than it not being wasm. I just tried to compile that snippet locally using `solc-js` (`./solcjs test.sol --bin`) and it crashes this way on all versions between 0.3.6 and 0.4.8, both asm.js and wasm ones.

LianaHus:
checked in other PR. Force merging.

cameel:
Oops. Looks like this has been merged before @axic actually applied my suggestions to the code.
In https://github.com/ethereum/remix-project/commit/e4da3b0d94dc2b262fa5b64e576776e66f1285b3 I see that it still checks > 0.6.3 rather than >= 0.6.2 like it should.

LianaHus:
gt was changed to gte

LianaHus:
it should be >= 0.3.6

cameel:
The condition should be: `semver.gte(version, '0.6.2') || (semver.gte(version, '0.3.6') && !isNightly)`. I'll submit another PR for this in moment.

LianaHus:
i see thanks

PR #1129, opened 2021-04-20
Use binaries.soliditylang.org instead of solc-bin

axic:
While solc-bin.ethereum.org should continue working, binaries.soliditylang.org is the preferred URL.

It also seems that `ethereum.github.io` is still used, which should not be as it is not updated. It is present in some seemingly generated code:
```
./apps/remix-ide/src/assets/js/0.7.7/app.js:      var url = 'https://ethereum.github.io/solc-bin/bin/soljson-' + versionString + '.js';
./apps/remix-ide/src/assets/js/0.7.7/app.js:      baseurl: 'https://solc-bin.ethereum.org/bin'
```

I am not sure what is the source of these.

### DIFF
--- a/apps/remix-ide/manifest.json
+++ b/apps/remix-ide/manifest.json
@@ -27,5 +27,5 @@
     "<all_urls>"
   ],
 
-  "content_security_policy": "script-src 'self' https://solc-bin.ethereum.org/; object-src 'self'"
+  "content_security_policy": "script-src 'self' https://binaries.soliditylang.org/; object-src 'self'"
 }
--- a/apps/remix-ide/src/app/compiler/compiler-utils.js
+++ b/apps/remix-ide/src/app/compiler/compiler-utils.js
@@ -2,8 +2,8 @@ const semver = require('semver')
 const minixhr = require('minixhr')
 /* global Worker */
 
-export const baseURLBin = 'https://solc-bin.ethereum.org/bin'
-export const baseURLWasm = 'https://solc-bin.ethereum.org/wasm'
+export const baseURLBin = 'https://binaries.soliditylang.org/bin'
+export const baseURLWasm = 'https://binaries.soliditylang.org/wasm'
 
 export const pathToURL = {}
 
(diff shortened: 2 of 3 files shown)


cameel:
> It also seems that `ethereum.github.io` is still used, which should not be as it is not updated. It is present in some seemingly generated code:
> 
> ```
> ./apps/remix-ide/src/assets/js/0.7.7/app.js:      var url = 'https://ethereum.github.io/solc-bin/bin/soljson-' + versionString + '.js';
> ./apps/remix-ide/src/assets/js/0.7.7/app.js:      baseurl: 'https://solc-bin.ethereum.org/bin'
> ```
> 
> I am not sure what is the source of these.

This seems to be a bunch of precompiled JS from Remix 0.7.7 added to the repo only to support serving the Old UI when you add `&appVersion=0.7.7` in the URL: https://github.com/ethereum/remix-project/pull/9/files#diff-db51681a5e11cd54660fa9913ab45b857015f285ec5a26488718a23123b3b9b9

But the Solidity version selector in that old UI works fine so I don't know what it's used for.

LianaHus:
> > It also seems that `ethereum.github.io` is still used, which should not be as it is not updated. It is present in some seemingly generated code:
> > ```
> > ./apps/remix-ide/src/assets/js/0.7.7/app.js:      var url = 'https://ethereum.github.io/solc-bin/bin/soljson-' + versionString + '.js';
> > ./apps/remix-ide/src/assets/js/0.7.7/app.js:      baseurl: 'https://solc-bin.ethereum.org/bin'
> > ```
> > 
> > 
> > I am not sure what is the source of these.
> 
> This seems to be a bunch of precompiled JS from Remix 0.7.7 added to the repo only to support serving the old UI when you add `&appVersion=0.7.7` in the URL.
> 
> https://github.com/ethereum/remix-project/pull/9/files#diff-db51681a5e11cd54660fa9913ab45b857015f285ec5a26488718a23123b3b9b9
> 
> But the Solidity version selector in that old UI works fine so I don't know what it's used for.

this is the old version of Remix-IDE. We are not supporting it but decided not to remove it as it is used in some old video tutorials and users were asking to keep it.

Issue #950, opened 2021-03-10
File Explorer rerendering the entire tree.


yann300:
When opening a n-th level tree, we can easily see that the file explorer is rerendering the entire tree, although it would be better to just render what need to be rendered.

ioedeveloper:
It is more visible on localhost. This may be because of the time delay between websocket response.

PR #1140, opened 2021-04-25
Optimise File-Explorer

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/950


apps/remix-ide-e2e/src/tests/libraryDeployment.test.ts
```diff
@@ -100,14 +100,14 @@ function checkDeployShouldSucceed (browser: NightwatchBrowser, address: string,
...
       addressRef = address
     })
-    .waitForElementPresent('.instance:nth-of-type(3)')
-    .click('.instance:nth-of-type(3) > div > button')
+    .waitForElementPresent('.instance')
+    .click('.instance > div > button')
     .perform(() => {
       browser
-        .testConstantFunction(addressRef, 'get - call', null, '0:\nuint256: 45')
+        .testConstantFunction(addressRef, 'getInt - call', null, '0:\nuint256: 45')
```
yann300, line 110: calling  `get` should have worked. 
`getInt` is calling the lib, but here we are testing that the lib is correctly linked to the contract.
the contract is:
```
function get () public view returns (uint) {
          return lib.getInt();
      }
```

apps/remix-ide/src/app/files/fileProvider.js
```diff
@@ -63,11 +63,11 @@ class FileProvider {
     })
   }
 
-  exists (path, cb) {
+  async exists (path) {
```
yann300, line 66: why do you need to add `async` if the content of the function is not using `await` ?

apps/remix-ide/src/app/files/workspaceFileProvider.js
```diff
@@ -33,7 +33,7 @@ class WorkspaceFileProvider extends FileProvider {
     if (!this.workspace) this.createWorkspace()
     path = path.replace(/^\/|\/$/g, '') // remove first and last slash
     if (path.startsWith(this.workspacesPath + '/' + this.workspace)) return path
-    if (path.startsWith(this.workspace)) return this.workspacesPath + '/' + this.workspace
+    if (path.startsWith(this.workspace)) return path.replace(this.workspace, this.workspacesPath + '/' + this.workspace)
```
yann300, line 36: i don't understand this change, is this a fix for when a folder has the same name as the workspace?

apps/remix-ide/src/lib/helper.js
```diff
@@ -52,6 +50,27 @@ module.exports = {
   createNonClashingName (name, fileProvider, cb) {
     this.createNonClashingNameWithPrefix(name, fileProvider, '', cb)
   },
+  async checkNonClashingNameAsync (name, fileManager, prefix = '') {
```
yann300, line 53: should it be `createNonClashingNameAsync` ?

apps/remix-ide/src/app/files/fileProvider.js
ioedeveloper: Adding async returns a promise. I added it because fileManager expects a promise just like how remixdProvider returns a promise.

apps/remix-ide/src/app/files/workspaceFileProvider.js
ioedeveloper: This fixes an issue i had when querying provider.exists(). If i have a path like `default_workspace/contracts/ballot.sol`, before resolving if the path exists `removePrefix()` is called. Now with the previous check, since the path starts with `default_workspace`, `.workspaces/default_workspace` is returned and when this (`.workspaces/default_workspace`) is checked if it exists, it always returns **true.** My change fixes that.

apps/remix-ide/src/lib/helper.js
ioedeveloper: I can rename it to `createNonClashingNameAsync`. It is the same implementation but async without callbacks.

apps/remix-ide-e2e/src/tests/libraryDeployment.test.ts
ioedeveloper: @yann300 I will need your help to figure out why testing `get` from libs fail.

bunsenstraat:
Cannot create workspaces, it actually creates a workspace directory, empty, in the default workspace. Somehow it also creates workspaces, sometimes, but you only see the list of spaces created before you create the new one. However it doesn't happen always, but it happens. Tried in incognito, cleared the storage. It also throws a lot of errors. The selector is stuck at connecting to localhost and can't find .workspaces. Who said coding wasn't fun?

Screenshot 2021-05-06 at 13 08 30

Screenshot 2021-05-06 at 13 13 01

Screenshot 2021-05-06 at 13 14 43

Screenshot 2021-05-06 at 13 14 33

yann300 (approved):

Issue #938, opened 2021-03-03
Warn if not latest version (remixd)


yann300:

tizah:
I tried to run an old version of Remix 0.10.7-dev - (which was at this point bff81d457... on 3rd Dec 2020), but was not getting the warning message

PR #1161, opened 2021-05-03
Warn if current is not the last version

yann300:
fix https://github.com/ethereum/remix-project/issues/938


yann300:
same PR in remix-desktop: https://github.com/ethereum/remix-desktop/pull/48

libs/remixd/src/bin/remixd.ts
```diff
@@ -40,6 +54,8 @@ function startService<S extends 'git' | 'folder'> (service: S, callback: (ws: WS
     }).parse(process.argv)
   // eslint-disable-next-line
 
+  warnLatestVersion()
```
Aniket-Engg, line 57: `await` should be added here.

libs/remixd/src/bin/remixd.ts
```diff
@@ -8,6 +10,18 @@ import * as fs from 'fs-extra'
 import * as path from 'path'
 import * as program from 'commander'
 
+async function warnLatestVersion () {
+  const latest = await latestVersion('@remix-project/remixd')
+  const pjson = require('../package.json')
+  if (semver.eq(latest, pjson.version)) {
+    console.log('\x1b[32m%s\x1b[0m', `[INFO] you are using the latest version ${latest}`)
+  } else if (semver.gt(latest, pjson.version)) {
+    console.log('\x1b[33m%s\x1b[0m', `[WARN] latest update is ${latest}, you are using ${pjson.version}`)
```
Aniket-Engg, line 19: May be `latest update` can be replaced with `latest version of remixd`

LianaHus (changes requested):
what if they use the old package? not the one from remix-project. In that case they should uninstall it and install the new one not just update.
see https://remix-ide.readthedocs.io/en/latest/remixd.html?highlight=remixd#update-to-the-latest-remixd

yann300:
> what if they use the old package?


we can't show this info in oldest package. this update will apply as soon as it's released but not for previous version

Aniket-Engg (approved):

LianaHus (approved):

Issue #1008, opened 2021-03-23
Hardhat and Remix compilationResult


Aniket-Engg:
- [ ] If a contract is compiled using Remix, it can port compilationResult to hardhat
- [ ] If a contract is compiled using hardhat, it can port compilationResult to Remix

This may require plugin development.

Aniket-Engg:
A websocket plugin has been created to run the hardhat compilation from remix Solidity Compiler plugin.

PR #1177, opened 2021-05-11
compile with hardhat and remixd

Aniket-Engg:
This is related to https://github.com/ethereum/remix-project/issues/1008

**Implementation:**

- `Solidity Compiler` plugin will show an extra checkbox `Enable Hardhat Compilation` only when Remix is connected to a Hardhat project using remixd. A project is hardhat project if it has a file named `hardhat.config.js`
- If this checkbox is checked and compilation is run, along with compilation on Remix, it will run hardhat compilation for that project with compiler configuration set in `Solidity Compiler` plugin and will generate the artefacts if required.
- It creates a file in project root with remix compiler configuration named `remixCompiler.config.js` which can be deleted once user is not connecting hardhat project with Remix IDE
- Hardhat compilation logging will be shown in remixd process terminal
- Hardhat compilation can not be run in remixd readonly mode

### DIFF
--- a/apps/remix-ide/src/app/files/fileManager.js
+++ b/apps/remix-ide/src/app/files/fileManager.js
@@ -57,12 +57,6 @@ class FileManager extends Plugin {
     this.mode = mode
   }
 
-  async compileWithHardhat (fileContent) {
-    const configFilePath = 'remixCompiler.config.js'
-    this.setFileContent(configFilePath, fileContent)
-    return await this.appManager.call('hardhat', 'compile', configFilePath)
-  }
-
   limitPluginScope (path) {
     return path.replace(/^\/browser\//, '').replace(/^browser\//, '') // forbids plugin to access the root filesystem
   }
--- a/apps/remix-ide/src/app/files/fileManager.js
+++ b/apps/remix-ide/src/app/files/fileManager.js
@@ -57,6 +57,11 @@ class FileManager extends Plugin {
     this.mode = mode
   }
 
+  async compileWithHardhat (cmd) {
+    console.log('Inside compileWithHardhat, calling hardhat compile using appManager')
+    return await this.appManager.call('hardhat', 'compile', cmd)
+  }
+
   limitPluginScope (path) {
     return path.replace(/^\/browser\//, '').replace(/^browser\//, '') // forbids plugin to access the root filesystem
   }
--- a/apps/remix-ide/src/app/files/fileManager.js
+++ b/apps/remix-ide/src/app/files/fileManager.js
@@ -57,9 +57,10 @@ class FileManager extends Plugin {
     this.mode = mode
   }
 
-  async compileWithHardhat (cmd) {
-    console.log('Inside compileWithHardhat, calling hardhat compile using appManager')
-    return await this.appManager.call('hardhat', 'compile', cmd)
+  async compileWithHardhat (fileContent) {
+    const configFilePath = 'remixCompiler.config.js'
+    this.setFileContent(configFilePath, fileContent)
+    return await this.appManager.call('hardhat', 'compile', configFilePath)
   }
 
   limitPluginScope (path) {
--- a/apps/remix-ide/src/app/files/remixd-handle.js
+++ b/apps/remix-ide/src/app/files/remixd-handle.js
@@ -22,7 +22,7 @@ const profile = {
   name: 'remixd',
   displayName: 'RemixD',
   url: 'ws://127.0.0.1:65520',
-  methods: ['folderIsReadOnly', 'resolveDirectory', 'get', 'exists', 'isFile', 'set', 'rename', 'remove', 'isDirectory', 'list', 'createDir'],
+  methods: ['folderIsReadOnly', 'hardhatCompile', 'resolveDirectory', 'get', 'exists', 'isFile', 'set', 'rename', 'remove', 'isDirectory', 'list', 'createDir'],
   events: [],
   description: 'Using Remixd daemon, allow to access file system',
   kind: 'other',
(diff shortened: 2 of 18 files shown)


apps/remix-ide/src/app/tabs/compileTab/compileTab.js
```diff
@@ -80,6 +80,22 @@ class CompileTab {
...
+          solidity: '${currentVersion.substring(0, currentVersion.indexOf('+commit'))}',
+          settings: {
+            optimizer: {
+              enabled: ${optimize},
+              runs: ${runs}
+            }
+          }
+        }
+        `
+        this.fileManager.compileWithHardhat(fileContent).then(console.log)
```
yann300, line 97: could you directly do the plugin call from the solidity compiler?

Aniket-Engg: ok

apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -57,6 +57,12 @@ class FileManager extends Plugin {
     this.mode = mode
   }
 
+  async compileWithHardhat (fileContent) {
+    const configFilePath = 'remixCompiler.config.js'
```
yann300, line 61: maybe this should be `'remix-compiler.config.js'`

apps/remix-ide/src/app/tabs/compile-tab.js
```diff
@@ -90,6 +90,13 @@ class CompileTab extends ViewPlugin {
    */
 
   listenToEvents () {
+    this.on('filePanel', 'setWorkspace', (workspace) => {
+      this.compileTabLogic.isHardhatProject().then((result) => {
+        if (result) this.compilerContainer.hardhatCompilation.style.display = 'flex'
```
yann300, line 95: instead of hiding, maybe this should be disabled.

apps/remix-ide/src/app/files/remixd-handle.js
```diff
@@ -81,7 +81,7 @@ export class RemixdHandle extends WebsocketPlugin {
           }
         }, 3000)
         this.locahostProvider.init(() => {})
-        // this.call('manager', 'activatePlugin', 'git')
+        this.call('manager', 'activatePlugin', 'hardhat')
```
yann300, line 84: there's also other place where you sould `'deactivatePlugin`

apps/remix-ide/src/app/tabs/compileTab/compileTab.js
yann300: that'd be still better to not have the hardhat plugin call in the filemanager. 
> have to make this method async, even for workspaces apart from local

`compileWithHardhat` is also `async` .. I don't see where is the issue.
And you don't need to `await` the hardhat compilation, just to manage the result in a promise resolve/reject

Aniket-Engg: this is done

apps/remix-ide/src/app/files/fileManager.js
Aniket-Engg: Done

apps/remix-ide/src/app/tabs/compile-tab.js
Aniket-Engg: There may be more checkboxes in future and only disabling them can take a lot of place. So as discussed, I will leave it like that

apps/remix-ide/src/app/files/remixd-handle.js
Aniket-Engg: done

apps/remix-ide/src/app/tabs/compileTab/compileTab.js
```diff
@@ -78,8 +78,30 @@ class CompileTab {
...
+            optimizer: {
+              enabled: ${optimize},
+              runs: ${runs}
+            }
+          }
+        }
+        `
+        const configFilePath = 'remix-compiler.config.js'
+        this.fileManager.setFileContent(configFilePath, fileContent)
+        this.fileManager.appManager.call('hardhat', 'compile', configFilePath)
```
yann300, line 103: you could just do `fileManager.call('hardhat', 'compile', configFilePath)` or better to turn this component as a plugin.

Aniket-Engg: updated

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -251,6 +255,7 @@ export const Workspace = (props: WorkspaceProps) => {
       // If 'connect to localhost' is clicked from home tab, mode is not 'localhost'
       if (props.fileManager.mode === 'localhost') {
         await setWorkspace(NO_WORKSPACE)
+        console.log('remixdExplorer.hide')
```
yann300, line 258: probably need to be removed

Aniket-Engg: ah, definitely

yann300 (approved):

Aniket-Engg:
Testing steps:

1. Create a hardhat project
2. Share it as a folder while running remixd
3. Open remix IDE on browser and 'Connect to localhost'
4. Go to 'Solidity Compiler' plugin, there will be one extra checkbox with label 'Enable Hardhat Compilation'
5. Check the box for 'Enable Hardhat Compilation' and compile a file
6. Compilation will be done on remix browser as well as on hardhat project with the configuration from `Solidity Compiler` plugin
7. Artifacts should be generated in hardhat project root
8. Result of compilation will be shown in remixd terminal console and Remix IDE terminal on browser

PR #1179, opened 2021-05-12
decode calldata during debugging

yann300:
fix https://github.com/ethereum/remix-project/issues/481


Aniket-Engg (approved):
change set looks fine

Issue #761, opened 2021-01-20
RegisterMenuItem / CustomAction


yann300:
Following https://github.com/ethereum/remix-project/pull/752.

We need to continue to address stuff like:
 - [x] name of the plugin (fileExplorers with or without "s", or maybe this can be named "filePanel")
 - [x] making the api generic enough so we can apply it to the runtab, compiletab, other plugins..
 - [ ] would it be conceivable to preload custom action (before plugin is activated)

Please update the list with anything relevant

bunsenstraat:
https://github.com/ethereum/remix-project/pull/1182
https://github.com/ethereum/remix-plugin/pull/368

PR #1182, opened 2021-05-14
added customactions

bunsenstraat:
example in client:

id should be unique, so best to use this.name which is the plugin name
```
this.methods = ["customAction"];

const cmd: customAction = {
      id: this.name,
      name: "customAction",
      type: ["file","folder"],
      extension: [],
      path: [],
      pattern: []
}

this.call("filePanel","registerContextMenuItem",cmd)
 
async customAction(o:customAction){
    console.log("custom action called", o)
}

```
Closes https://github.com/ethereum/remix-project/issues/761


libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
```diff
@@ -163,10 +166,7 @@ export const FileExplorer = (props: FileExplorerProps) => {
   useEffect(() => {
     if (contextMenuItems) {
       setState(prevState => {
-        // filter duplicate items
-        const items = contextMenuItems.filter(({ name }) => prevState.actions.findIndex(action => action.name === name) === -1)
-
-        return { ...prevState, actions: [...prevState.actions, ...items] }
+        return { ...prevState, actions: [...initialActions, ...contextMenuItems] }
```
ioedeveloper, line 169: This might not work because in the copy & paste PR (https://github.com/ethereum/remix-project/pull/1209/files#diff-12461633b1b7400c1f8d4b9e2480ab50e052db7f4c71591121d2b47f8077c677R244, https://github.com/ethereum/remix-project/pull/1209/files#diff-12461633b1b7400c1f8d4b9e2480ab50e052db7f4c71591121d2b47f8077c677R253), context menu items can be added and removed during runtime for file-explorer. It will be better to have `{ ...prevState, actions: [...prevState.actions, ...contextMenuItems] }`. We still need to check for duplicates because the check for duplicate in file-panel is not sufficient. File-panel does not have access to `initialActions` and cannot prevent adding menus already present in `initailActions`.

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -102,11 +104,20 @@ module.exports = class Filepanel extends ViewPlugin {
...
+    if (this.registeredMenuItems.filter((o) => {
+      return o.id === item.id & o.name === item.name
+    }).length) throw new Error(`Action ${item.name} already exists on ${item.id}`)
     this.registeredMenuItems = [...this.registeredMenuItems, item]
     this.renderComponent()
   }
 
+  removePluginActions (plugin) {
+    this.registeredMenuItems = this.registeredMenuItems.filter((item) => {
+      return item.id !== plugin.name || item.sticky === true
```
ioedeveloper, line 116: What do you mean by `item.sticky`?

package.json
```diff
@@ -161,7 +161,6 @@
     "jszip": "^3.6.0",
     "lodash": "^4.17.21",
     "latest-version": "^5.1.0",
-    "lodash": "^4.17.21",
```
ioedeveloper, line 164: Any reason why lodash is removed?

bunsenstraat: I didn't do that, don't know why this is.

apps/remix-ide/src/app/panels/file-panel.js
bunsenstraat: It remains in the context menu, so that when clicked the plugin will be activated and the method called. this allows us to for example include methods at app startup or any other phase. For example for dgit that would be useful. when the plugin is deactivated it doesn't dissappear.

libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
bunsenstraat: Ok sure, but there needs to be a way for items to be removed when contextMenuItems prop changes, in this case shrinks when a plugin is deactivated. however way that is done is fine. but likes this all the menuitems will remain after deactivation.

ioedeveloper: Yeah, you're right. Currently, the `useEffect` for `contextMenuItems` makes use of `addMenuItems` https://github.com/ethereum/remix-project/pull/1209/files#diff-12461633b1b7400c1f8d4b9e2480ab50e052db7f4c71591121d2b47f8077c677R187. We can do the same for when items are removed since the file-panel manages the removal of menu items. We would just need to call `removeMenuItems` with an array of the ids.

apps/remix-ide/src/app/panels/file-panel.js
ioedeveloper: Ah! Alright.

ioedeveloper (approved):

andro54:
It,s good

libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
bunsenstraat: @ioedeveloper this is still an unresolved issue. So what do we do? I don't want to mess in your react code but I can make a suggestion if needed.

ioedeveloper (approved):

PR #1184, opened 2021-05-14
Multi select & multi context menu actions & deleting multi

bunsenstraat:
Action type can now have 'multi', 'file', 'folder' as a type

Multi means it supports multiple selected items. The context item will then have to meet all the conditions of all the items selected. If one condition fails, for example an extension doesn't match on any of the selected items, the action cannot be performed. 

The multi key is Shift maybe that can be a setting? or we have a setting somewhere? 

```
{
      id: 'delete',
      name: 'Delete',
      type: ['file', 'folder', 'multi'],
      path: [],
      extension: [],
      pattern: []
    }
```


libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
```diff
@@ -280,29 +304,32 @@ export const FileExplorer = (props: FileExplorerProps) => {
     }
   }
 
-  const deletePath = async (path: string) => {
+  const deletePath = async (path: string | string[]) => {
```
ioedeveloper, line 307: I think it will be better to create a new function that accepts `string[]` and uses the implementation of `deletePath`.

libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
```diff
@@ -88,6 +85,7 @@ export const FileExplorer = (props: FileExplorerProps) => {
       hide: true,
       title: '',
       message: '',
+      children: <></>,
```
ioedeveloper, line 88: I'm not sure we need children for what you are trying to achieve.

bunsenstraat (commented):
@ioedeveloper If we don't do this  https://github.com/ethereum/remix-project/pull/1184/commits/943890a2489a1c66586406af695b7fb84091d5c0 it will only take into account the type property, not the selected files/folders and you'd not be able to create an action that applies to files only, because when selecting folder and some files, right click on the files it would see type == file and allow the action to be visible while you'd also have a folder selected

ioedeveloper (approved):

Issue #850, opened 2021-02-08
Rt click to publish a folder or file to a GIST


ryestew:
It is currently hard to publish a few files to a GIST because you have to erase everything else to do that.  If you could publish with a right click it would help a lot.

PR #1185, opened 2021-05-14
Publish files and folder to gist

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/850


bunsenstraat:
If I  type in a file, and then not trigger the save action, for example by switching files or whatever but directly click on publish to gist it will publish the old content. So maybe save first then gist?

yann300:
It seems that the filename should be sanitized when importing a gist:

 - publish a file
 - in gist, the title looks like `yann300/3_Ballot.sol`
 - I re-import it from the home page, in the tab, the file name appears as `gist-<id>/<filename>`
 - when I push the changes to gist, the title is now `yann300/gist-0c5b0ed330a050e7a3f958cd83e04ba9...3_Ballot.sol`

yann300 (approved):

ioedeveloper:
> this will only replace the first occurrence of `...` .
> try `test.replace(/\.\.\./g, '/')`

Yes, thanks.

yann300:
I just tried https://github.com/ethereum/remix-project/pull/1185/commits/ec52d92d7b4aa3845ce7260d0feb0f0829318bfa this works well

PR #1190, opened 2021-05-17
fix connect to localhost from home tab

### LINKED ISSUE #1183: Issues in switching to and from localhost workspace
- [ ] If current workspace is `xyz` and I click on `connect to localhost` then on `cancel` in modal, workspace is automatically switched to `default_workspace`
- [ ] If current workspace is `xyz` and I click on `connect to localhost`, click on `connect` in modal but remixd is not running, workspace is automatically switched to `default_workspace`
- [ ] If current workspace is `localhost` and I switch to workspace `xyz`, workspace is automatically switched to `default_workspace`

I analysed 3rd point locally and I see that even after automatically switching to `default_workspace`, file manager mode is `localhost`

Aniket-Engg:
fixes #1183

### DIFF
--- a/libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
+++ b/libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
@@ -278,6 +278,10 @@ export const Workspace = (props: WorkspaceProps) => {
         setState(prevState => {
           return { ...prevState, hideRemixdExplorer: true, loadingLocalhost: false }
         })
+      } else {
+        setState(prevState => {
+          return { ...prevState, loadingLocalhost: false }
+        })
       }
     },
     show: () => {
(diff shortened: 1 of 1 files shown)


ioedeveloper (approved):

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -91,6 +91,7 @@ export const Workspace = (props: WorkspaceProps) => {
 
   const localhostDisconnect = () => {
     if (state.currentWorkspace === LOCALHOST) setWorkspace(props.workspaces.length > 0 ? props.workspaces[0] : NO_WORKSPACE)
+    else setWorkspace(state.currentWorkspace) // Useful to switch to last selcted workspace when remixd is disconnected
```
yann300, line 94: is this actually making a difference? I think this could lead to some unwanted behavior...

yann300 (approved):

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
Aniket-Engg: This is  a way-around for now. This will be improved in a different refactor PR. An issue will be opened to track this.

Aniket-Engg: Here is the issue: https://github.com/ethereum/remix-project/issues/1197

PR #1194, opened 2021-05-17
Catch reserved keywords and display modal

ioedeveloper:


libs/remix-ui/file-explorer/src/lib/file-explorer.tsx
```diff
@@ -650,15 +650,31 @@ export const FileExplorer = (props: FileExplorerProps) => {
         }, null)
       } else {
         if (state.focusEdit.isNew) {
-          state.focusEdit.type === 'file' ? createNewFile(joinPath(parentFolder, content)) : createNewFolder(joinPath(parentFolder, content))
-          removeInputField(parentFolder)(dispatch)
+          if ((content === props.name) || (content.indexOf('gist-') === 0)) {
```
yann300, line 653: as it is here it will be a bit cumbersome to add new reserved keyword... might have been clearer to keep the reserved keyword in a list as a property in file-explorer, or to add a `hasReservedKeyword()`.

yann300 (approved):

PR #1202, opened 2021-05-18
Make compiler logic a plugin

yann300:
this PR:

 - fix a bug which occur when compiling with "ctrl-s"
 - make compileTabLogic a plugin so it can directly call hardhat plugin

### DIFF
--- a/apps/remix-ide/src/remixAppManager.js
+++ b/apps/remix-ide/src/remixAppManager.js
@@ -14,7 +14,7 @@ const requiredModules = [ // services + layout views + system views
 const dependentModules = ['git', 'hardhat'] // module which shouldn't be manually activated (e.g git is activated by remixd)
 
 export function isNative (name) {
-  const nativePlugins = ['vyper', 'workshops', 'debugger', 'remixd', 'menuicons']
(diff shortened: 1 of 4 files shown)


apps/remix-ide/src/remixAppManager.js
```diff
@@ -14,7 +14,7 @@ const requiredModules = [ // services + layout views + system views
 const dependentModules = ['git', 'hardhat'] // module which shouldn't be manually activated (e.g git is activated by remixd)
 
 export function isNative (name) {
-  const nativePlugins = ['vyper', 'workshops', 'debugger', 'remixd', 'menuicons']
+  const nativePlugins = ['vyper', 'workshops', 'debugger', 'remixd', 'menuicons', 'solidity']
```
Aniket-Engg, line 17: not sure if `solidity-logic` should be added instead of `solidity`

yann300: this `isNative` is used in the `canActivate` function, it allows solidity to activate solidity-logic

Aniket-Engg: ok

Aniket-Engg (approved):

PR #1204, opened 2021-05-19
show hh compilation result in terminal

### LINKED ISSUE #1201: Show hardhat compilation result in remix terminal
In continuation with https://github.com/ethereum/remix-project/pull/1177

Aniket-Engg:
fixes #1201 

- remix-plugin deps updated to latest to use terminal log API
- Hardhat compilation result shown in remix terminal
- remixd logging improved
- listener name is dynamically shown on remixd console according to port
- Run hardhat web-socket only if a hardhat project is connected. Existence of Hardhat config file is checked to label it as a hardhat project


apps/remix-ide/src/app/tabs/compileTab/compileTab.js
```diff
@@ -109,7 +109,11 @@ class CompileTab extends Plugin {
         `
         const configFilePath = 'remix-compiler.config.js'
         this.fileManager.setFileContent(configFilePath, fileContent)
-        this.call('hardhat', 'compile', configFilePath)
+        this.call('hardhat', 'compile', configFilePath).then((result) => {
```
yann300, line 112: I thought we could so something like `this.call('terminal', 'log', result)` instead of using event, but that works too, will add @bunsenstraat for review.

Aniket-Engg: Terminal plugin is not exposing any methods: https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/panels/terminal.js#L29

yann300: we have a pr from an external dev which expose terminal API, that maybe time for the terminal to have a minimal API.
let see what @bunsenstraat think

bunsenstraat: https://github.com/ethereum/remix-project/pull/1210

this.call('terminal', 'log', result) will now work, it will also deal with \n and so on

bunsenstraat (commented):
this.call('terminal', 'log', result) will work now. No need for events.

yann300 (approved):

bunsenstraat (approved):

PR #1206, opened 2021-05-19
Add hardhat as an explicit option in udapp

Aniket-Engg:


apps/remix-ide/src/app/tabs/runTab/settings.js
```diff
@@ -244,16 +244,29 @@ class SettingsUI {
...
   setExecutionContext (context) {
-    this.blockchain.changeExecutionContext(context, () => {
-      modalDialogCustom.prompt('External node request', this.web3ProviderDialogBody(), 'http://127.0.0.1:8545', (target) => {
-        this.blockchain.setProviderFromEndpoint(target, context, (alertMsg) => {
-          if (alertMsg) addTooltip(alertMsg)
-          this.setFinalContext()
-        })
+    if (context === 'Hardhat Provider') {
+      this.blockchain.changeExecutionContext(context, () => {
+        modalDialogCustom.prompt('Hardhat node request', this.hardhatProviderDialogBody(), 'http://127.0.0.1:8545', (target) => {
```
yann300, line 249: wasn't it possible to put this in https://github.com/ethereum/remix-project/pull/1206/files#diff-aedcee9b7a11a19d2ec1b9a1831bd5d4985392e3701b74003c0171f1b2e74c2bR21 ?
I mean your way work but it'd be better to have hardhat stuff in a single.
Also having declared a whole plugin but still need to do that is a bit overkill.

apps/remix-ide/src/app/tabs/hardhat-provider.js
```diff
@@ -58,7 +55,8 @@ const sendAsyncInternal = (provider, data, resolve, reject) => {
       resolve(message)
     })
   } else {
-    resolve({ jsonrpc: '2.0', result: [], id: data.id })
+    const result = data.method === 'net_listening' ? 'canceled' : []
```
yann300, line 58: actually maybe a better way would have been to `reject` if the provider is not set..

apps/remix-ide/src/blockchain/execution-context.js
```diff
@@ -189,14 +188,16 @@ export class ExecutionContext {
...
-      if (!err && isConnected) {
+      if (!err && isConnected === true) {
         this.executionContext = context
         this._updateBlockGasLimit()
         this.event.trigger('contextChanged', [context])
         this.event.trigger('web3EndpointChanged')
         cb()
+      } else if (isConnected === 'canceled') {
+        web3.setProvider(oldProvider)
+        cb()
```
yann300, line 200: yes, indeed that's a clever solution.
Perhaps `cb()` should also return an `error` (same as in the `else` below)

yann300 (approved):

apps/remix-ide/src/app/tabs/hardhat-provider.js
Aniket-Engg: `reject` will also show a predefined error on toaster

apps/remix-ide/src/blockchain/execution-context.js
Aniket-Engg: Generally, no message is shown on clicking on `cancel` in a modal

apps/remix-ide/src/app/tabs/runTab/settings.js
Aniket-Engg: done

yann300 (approved):

PR #1214, opened 2021-05-21
Fix For Float Number In Value Field

### LINKED ISSUE #1100: can't input float number in value field
hi, I can't input 0.1 eth in the `value` field, please check. The point is just can't type into the field. I type `0.1` get `01`

image

@yann300

Patil2099:
Fixes #1100


bunsenstraat:
You're editing js files. You should edit the TS components.

PR #1248, opened 2021-06-01
remove double getworkspace call

bunsenstraat:
A new call to getworkspaces was added via https://github.com/ethereum/remix-project/pull/1140 to create workspaces and get them 
https://github.com/ethereum/remix-project/blob/f974fcf823a6852c232ddc1e3357c722bb4d246e/apps/remix-ide/src/app/panels/file-panel.js#L174 
So we don't need the first one. 

Also a fix was need as a test failed that preloads a file when there is no workspace.
The file creation has a delay which caused the filemanager not to find the file. Writing directly to filesystem solves this particular problem.

### DIFF
--- a/apps/remix-ide/src/app/panels/file-panel.js
+++ b/apps/remix-ide/src/app/panels/file-panel.js
@@ -14,6 +14,7 @@ const examples = require('../editor/examples')
 const GistHandler = require('../../lib/gist-handler')
 const QueryParams = require('../../lib/query-params')
 const modalDialogCustom = require('../ui/modal-dialog-custom')
+const path = require('path')
 /*
   Overview of APIs:
    * fileManager: @args fileProviders (browser, shared-folder, swarm, github, etc ...) & config & editor
@@ -67,7 +68,7 @@ module.exports = class Filepanel extends ViewPlugin {
   }
 
   render () {
-    this.initWorkspace().then(() => this.getWorkspaces()).catch(console.error)
+    this.initWorkspace().then(async () => await this.getWorkspaces()).catch((error) => console.log(error))
     return this.el
   }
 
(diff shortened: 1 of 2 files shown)


apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -67,7 +68,7 @@ module.exports = class Filepanel extends ViewPlugin {
   }
 
   render () {
-    this.initWorkspace().then(() => this.getWorkspaces()).catch(console.error)
+    this.initWorkspace().then(async () => await this.getWorkspaces())
```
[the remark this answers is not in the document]
yann300: I don't think this changes anything. cause the `render` will still not wait. or am i missing something?

apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -165,13 +166,17 @@ module.exports = class Filepanel extends ViewPlugin {
...
     // insert example contracts if there are no files to show
-    this._deps.fileProviders.browser.resolveDirectory('/', async (error, filesList) => {
-      if (error) console.error(error)
-      if (Object.keys(filesList).length === 0) {
-        await this.createWorkspace('default_workspace')
-      }
-      this.getWorkspaces()
+    return new Promise((resolve, reject) => {
+      this._deps.fileProviders.browser.resolveDirectory('/', async (error, filesList) => {
+        if (error) console.error(error)
```
yann300, line 174: you need to `reject` here

yann300: and `return` too

bunsenstraat: > you need to `reject` here



> and `return` too

it's already returning a new promise

yann300: I mean instead of doing `if (error) console.error(error)`
maybe this should do if (error) return reject(error)`

PR #1268, opened 2021-06-08
refactor setworkspace event

bunsenstraat:
problem to be solved:
- setworkspace should only be fired when you're sure the new fileprovider is ready to accept calls
- plugins that react to this event then call the correct fileprovider
- remixd knows when the provider is ready, only then fire the event
- now the events will be fired correctly also when going back to default_workspace


yann300 (approved):

Aniket-Engg (approved):
`workspace already exists` error not appearing anymore

ioedeveloper (approved):

PR #1286, opened 2021-06-15
fix for injected web3 call

### LINKED ISSUE #1282: [Bug] Contract call failing in Remix UI (failing to read on-chain data)
I was trying to call my contract that's deployed to Kovan, but I kept getting this Metamask error in my browser console:
``Invalid params: unknown field \`timestamp\`, expected one of \`type\`, \`from\`, \`to\`, \`gasPrice\`, \`gas\`, \`value\`, \`data\`, \`nonce\`, \`accessList\`.``
I tried with my Kovan Infura provider instead on Metamask, still got the same error:
``call to PriceConsumerV3.getLatestPrice errored: Error: Returned error: Invalid params: unknown field `timestamp`, expected one of `type`, `from`, `to`, `gasPrice`, `gas`, `value`, `data`, `nonce`, `accessList`. 
``
I double-checked that the network was switched to Kovan for both times.

The contract call works just fine on Etherscan's contract UI though: https://kovan.etherscan.io/address/0x31C09Eaf8bA8D8FC9Ab044896F870cD737617FF6#readContract

Contract code is available on that same Etherscan link above.

Aniket-Engg:
fixes #1282

### DIFF
--- a/libs/remix-lib/src/execution/txRunnerWeb3.ts
+++ b/libs/remix-lib/src/execution/txRunnerWeb3.ts
@@ -72,13 +72,19 @@ export class TxRunnerWeb3 {
     const tx = { from: from, to: to, data: data, value: value }
 
     if (useCall) {
-      const tag = Date.now() // for e2e reference
       tx['gas'] = gasLimit
-      // tx['timestamp'] = timestamp
-      return this.getWeb3().eth.call(tx, function (error, result: any) {
-        if (error) return callback(error)
-        callback(null, {
-          result: result
+      tx['timestamp'] = timestamp
+      return this._api.detectNetwork((err, network) => {
+        if (err) {
+          console.log(err)
+          return
+        } else if (network && network.name === 'Kovan') delete tx['timestamp']
+
+        return this.getWeb3().eth.call(tx, function (error, result: any) {
+          if (error) return callback(error)
+          callback(null, {
+            result: result
+          })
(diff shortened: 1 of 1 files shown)


yann300 (approved):

libs/remix-lib/src/execution/txRunnerWeb3.ts
```diff
@@ -72,13 +72,19 @@ export class TxRunnerWeb3 {
...
       tx['timestamp'] = timestamp
-      return this.getWeb3().eth.call(tx, function (error, result: any) {
-        if (error) return callback(error)
-        callback(null, {
-          result: result
+      return this._api.detectNetwork((err, network) => {
+        if (err) {
+          console.log(err)
+          return
+        } else if (network && network.name === 'Kovan') delete tx['timestamp']
```
yann300, line 81: could you add a comment for explaining why we are removing the timestamp only for kovan?

Aniket-Engg: sure

yann300 (approved):

libs/remix-lib/src/execution/txRunnerWeb3.ts
Bingyy: Do we need to upgrade metamask to use this updated code?

PR #1306, opened 2021-06-22
fix hardhat modal reappearing

### LINKED ISSUE #1300: hardhat provider modal won't stop reappearing
When I stopped the hardhat node and then tried to turn off the Remix-hardhat provider, I cant get the modal to stop reappearing.  I hit cancel and it comes right back.
image

Aniket-Engg:
fixes #1300


apps/remix-ide/src/app/tabs/hardhat-provider.js
```diff
@@ -32,19 +32,30 @@ export default class HardhatProvider extends Plugin {
...
           this.sendAsyncInternal(data, resolve, reject)
         }, () => {
-          this.sendAsyncInternal(data, resolve, reject)
+          // If 'cancel' is clicked while trying to connect, handle it in custom manner
+          if (data.method === 'net_listening') resolve({ jsonrpc: '2.0', result: 'canceled', id: data.id })
+          else {
+            // When node is abruptly stopped, modal will appear
+            // On which clicking on 'Cancel' will set the Envrionment to VM
+            this.blockchain.changeExecutionContext('vm')
+            this.provider = this.blockchain.getCurrentProvider()
```
yann300, line 56: so `this.provider` will contain the VM provider in that case?

yann300: if there's no connection, i'd just reset `this.provider` in that case.

Aniket-Engg: That is the issue, it opens modal again if `(!this.provider || data.method === 'net_listening')` returns true. with VM provider, it will go to `else` and in `sendAsyncInternal ` , it will throw error on failing `(this.blockchain.getProvider() !== 'Hardhat Provider' && data.method !== 'net_listening')` condition.

yann300: ok maybe set it to a fake provider (object which has an empty sendAsync method), because setting it to the current provider is note really accurate.

Aniket-Engg: We set it to VM when user clicks on `cancel`

ryestew (commented):
I am running the PR locally and have started the hardhat node and have the hardhat provider plugin loaded - but I cannot get access to the accounts - so I have the following warning:

image

ryestew:
On the plus side the intended fix seems to be working - when I stop the hardhat node - I get the modal warning and then click cancel - and it does not reappear.

ryestew:
A side note - that is not part of this PR and I don't think it is really a problem - but maybe it is an unintended consequence- if I have the hardhat node running but the hardhat provider is NOT activated and I click on web3 provider, I can connect to the hardhat node.

Aniket-Engg:
> I am running the PR locally and have started the hardhat node and have the hardhat provider plugin loaded - but I cannot get access to the accounts - so I have the following warning:
> 
> image

ok, that is strange. I have never got this warning

Aniket-Engg:
> A side note - that is not part of this PR and I don't think it is really a problem - but maybe it is an unintended consequence- if I have the hardhat node running but the hardhat provider is NOT activated and I click on web3 provider, I can connect to the hardhat node.

Yes, that is supposed to be working.

yann300 (approved):

ryestew (approved):

Issue #1155, opened 2021-04-30
Debugger: Enhanced code display


chriseth:
The code display in the debugger is small enough for most uses, but sometimes it can be useful to know what is further up or further down in the code, so it would be nice if the display could be scrollable or there would be the possibility to enlarge the code view.

![image](https://user-images.githubusercontent.com/9073706/116675040-aea9e580-a9a5-11eb-9e75-512d32523313.png)

Another feature for the code view could be to display if a conditional jump will be taken or not, so that we already know in advance that we will jump in to steps. Maybe opcodes that will be executed soon could be highlighted, or there could be a line visualizing the control flow:

![cf](https://user-images.githubusercontent.com/9073706/116676107-e1081280-a9a6-11eb-8046-be3b239fc0bd.png)

chriseth:
(updated the second image)

chriseth:
Similarly, a return, revert or stop statement could end the line in a circle or something like that. Of course, this line should only be drawn for steps that are close to the current step. It could fade out if it is more than e.g. 20 steps away.

chriseth:
Illustration about how to visualize a STOP / RETURN / REVERT:
![image](https://user-images.githubusercontent.com/9073706/123070548-26c7e100-d414-11eb-8e95-e0959eabc3e8.png)

slightly different illustrations could be used for STOP/RETURN and REVERT and just going out of gas.

chriseth:
Wohoo! This is great!
![image](https://user-images.githubusercontent.com/9073706/125813723-53bb62c1-9fb8-41d1-9cde-95036be064b1.png)
Thanks a lot!

Issue #578, opened 2020-11-15
Debugger Config UI design needs a slight makeover


ryestew:
The top and bottom margin of the Debugger config stuff needs an update.
The new Debugger Config currently looks like this:
<img width="421" alt="image" src="https://user-images.githubusercontent.com/6050001/99161420-24ecb100-26c0-11eb-9fe3-ff83e7010626.png">

It used to look like this:
![image](https://user-images.githubusercontent.com/6050001/99161274-af341580-26be-11eb-8e15-5dc5369728f0.png)

So now the hash of the txn is not the at the top of the screen - and the first thing people see is the Debugger Configuration stuff.  I'm not sure this should be at the top.  But let's leave it there for now.  But It seems not important enough to take up so much vertical height ( 2 lines for a single checkbox).

Here's a version with the spacing improved:
![image](https://user-images.githubusercontent.com/6050001/99161322-29649a00-26bf-11eb-9fd1-086bd934559d.png)

Here's a version with the config checkbox in a collapsed card:
![image](https://user-images.githubusercontent.com/6050001/99161340-6597fa80-26bf-11eb-91dd-360d43bd8ae2.png)

This puts everything on a single line and makes it fit in with the rest of the page.  Although maybe we should consider putting it lower down on on the module.

We need to update docs as well
(Liana)

LianaHus:
to make things easier/clearer we can remove the label 'Debugger Configuration" and leave one line with checkbox only.
if later we add more configuration that can be restored(expandable for example as you're suggesting)
i don't see any other place to put this. somewhere at the end of the plugin will not work

PR #1315, opened 2021-06-23
Improve opcode view when debugging

yann300:
fix https://github.com/ethereum/remix-project/issues/1155
fix https://github.com/ethereum/remix-project/issues/578

### DIFF
--- a/libs/remix-debug/src/code/codeManager.ts
+++ b/libs/remix-debug/src/code/codeManager.ts
@@ -147,13 +147,15 @@ export class CodeManager {
 
   private retrieveIndexAndTrigger (codeMananger, address, step, code) {
     let result
+    let next
     try {
       result = codeMananger.getInstructionIndex(address, step)
+      next = codeMananger.getInstructionIndex(address, step + 1)
     } catch (error) {
       return console.log(error)
     }
     try {
-      codeMananger.event.trigger('changed', [code, address, result])
+      codeMananger.event.trigger('changed', [code, address, result, next])
     } catch (e) {
       console.log('dispatching event failed', e)
     }
--- a/libs/remix-debug/src/code/codeManager.ts
+++ b/libs/remix-debug/src/code/codeManager.ts
@@ -159,7 +159,7 @@ export class CodeManager {
       if (values) {
         for (const value of values) {
           if (value.address === address) {
-            returnInstructionIndexes.push({ instructionIndex: this.getInstructionIndex(address, value.index), address })        
+            returnInstructionIndexes.push({ instructionIndex: this.getInstructionIndex(address, value.index), address })
           }
         }
       }
@@ -168,7 +168,7 @@ export class CodeManager {
       if (values) {
         for (const value of values) {
           if (value.address === address) {
-            outOfGasInstructionIndexes.push({ instructionIndex: this.getInstructionIndex(address, value.index), address })        
+            outOfGasInstructionIndexes.push({ instructionIndex: this.getInstructionIndex(address, value.index), address })
           }
         }
       }
--- a/libs/remix-debug/src/code/codeManager.ts
+++ b/libs/remix-debug/src/code/codeManager.ts
@@ -148,8 +148,8 @@ export class CodeManager {
   private async retrieveIndexAndTrigger (codeMananger, address, step, code) {
     let result
     let next
-    let returnInstructionIndexes = []
-    let outOfGasInstructionIndexes = []
+    const returnInstructionIndexes = []
+    const outOfGasInstructionIndexes = []
 
     try {
       result = codeMananger.getInstructionIndex(address, step)
@@ -170,10 +170,9 @@ export class CodeManager {
           outOfGasInstructionIndexes.push({ instructionIndex: this.getInstructionIndex(address, value.index), address })
         }
       }
-
-      } catch (error) {
-        return console.log(error)
-      }
+    } catch (error) {
+      return console.log(error)
+    }
     try {
       codeMananger.event.trigger('changed', [code, address, result, next, returnInstructionIndexes, outOfGasInstructionIndexes])
     } catch (e) {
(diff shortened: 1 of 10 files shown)


libs/remix-debug/src/code/codeManager.ts
```diff
@@ -145,15 +145,36 @@ export class CodeManager {
...
-  private retrieveIndexAndTrigger (codeMananger, address, step, code) {
+  private async retrieveIndexAndTrigger (codeMananger, address, step, code) {
     let result
+    let next
+    const returnInstructionIndexes = []
+    const outOfGasInstructionIndexes = []
+
     try {
       result = codeMananger.getInstructionIndex(address, step)
+      next = codeMananger.getInstructionIndex(address, step + 1)
```
ioedeveloper, line 156: I think line 156 to 164 can come in a single loop.

libs/remix-debug/src/code/codeManager.ts
```diff
@@ -145,15 +145,36 @@ export class CodeManager {
...
+
+      let values = this.traceManager.getAllStopIndexes()
+      values = values.filter((value) => value.address === address)
+      if (values) {
+        for (const value of values) {
+          returnInstructionIndexes.push({ instructionIndex: this.getInstructionIndex(address, value.index), address })
+        }
+      }
+
+      values = this.traceManager.getAllOutofGasIndexes()
```
ioedeveloper, line 166: Same here too.

libs/remix-ui/debugger-ui/src/lib/vm-debugger/assembly-items.tsx
```diff
@@ -4,24 +4,32 @@ import './styles/assembly-items.css'
...
-    registerEvent && registerEvent('codeManagerChanged', (code, address, index) => {
-      dispatch({ type: 'FETCH_OPCODES_SUCCESS', payload: { code, address, index } })
+    registerEvent && registerEvent('codeManagerChanged', (code, address, index, nextIndex, returnInstructionIndexes, outOfGasInstructionIndexes) => {
+      dispatch({ type: 'FETCH_OPCODES_SUCCESS', payload: { code, address, index, nextIndex, returnInstructionIndexes, outOfGasInstructionIndexes } })
     })
   }, [])
 
   useEffect(() => {
-    if (selectedItem !== assemblyItems.index) {
+    console.log('useEffect', assemblyItems.index)
```
ioedeveloper, line 22: You forgot to take this console.log

libs/remix-ui/debugger-ui/src/lib/vm-debugger/assembly-items.tsx
```diff
@@ -30,15 +38,108 @@ export const AssemblyItems = ({ registerEvent }) => {
       if (currentItem.firstChild) {
         currentItem.firstChild.removeAttribute('style')
       }
-      const codeView = asmItemsRef.current
+    }
+
+    currentItem = refs.current[nextSelectedItem] ? refs.current[nextSelectedItem] : null
+
+    if (currentItem) {
```
ioedeveloper, line 45: This can refactored to be less repeatitive.
```
    if (currentItem) {
      currentItem.removeAttribute('selected')
      currentItem.removeAttribute('style')
      if (currentItem.firstChild) {
        currentItem.firstChild.removeAttribute('style')
      }
    }
```

libs/remix-debug/src/code/codeManager.ts
yann300: done

ryestew (commented):
I wanted to go through Christians points to go over your implementation:

Here's what I have in the opcodes when there is a STOP in the opcodes panel:
image

This is to address Christians points:
## Point 1.
**The code display in the debugger is small enough for most uses, but sometimes it can be useful to know what is further up or further down in the code**

So you have the next opcode with a dotted border.  (the top of this box is not so happy to be touching the solid border above - but this is a minor point) Is the next opcode the important one?  The next one - in my opinion is easy to spot - but one 4 or 5 down could be tricky...I think.  Does something like this address his comment better?

image


## Point 2. 
**A return, revert or stop statement could end the line in a circle or something like that**
llustration about how to visualize a STOP / RETURN / REVERT:
image

slightly different illustrations could be used for STOP/RETURN and REVERT and just going out of gas.

I think your dotted red box is a good answer - at least for STOP.  
Is Christian looking for a clue about how far back REVERT will jump back to?

For issue: https://github.com/ethereum/remix-project/issues/578 - I think your solution is fine.

yann300:
> The next one - in my opinion is easy to spot - but one 4 or 5 down could be tricky...I think.

the next one is easy to spot when it's not a Jump. In case of a Jump, the JumpDest can be anywhere.

>  Does something like this address his comment better?

I think that in that case he just meant that the opcode view should be scrollable (in the PR more opcodes are loaded, but not all of it), that would come in the a next PR (not this release)

yann300:
> slightly different illustrations could be used for STOP/RETURN and REVERT and just going out of gas.

There's already a different illustration for outofgas and the rest.

yann300:
> Is Christian looking for a clue about how far back REVERT will jump back to?

moving one step forward (after the revert) would anyway go to the place where the execution is reverted back to. so I think that is ok..

yann300:
> On a side note - I adore dotted borders - but it is so hard to find the right place to use them. I think the red dots are a nice usage.

Actually I just tried to replace dot with background color, but I couldn't find the right way to do

ryestew (approved):

chriseth:
>  The next one - in my opinion is easy to spot - but one 4 or 5 down could be tricky...I think.
>> the next one is easy to spot when it's not a Jump. In case of a Jump, the JumpDest can be anywhere.

While this is true, it is especially useful to know whether or not I can still press "forward" five times without a jump happening. So the current implementation only tells me whether or not a jump will be taken, but only when I'm already at the jump itself. It would be really cool to have the ability to look forward in time 4 or five steps.

Maybe we don't even need borders at all. What about drawing opcodes that will be executed during the next 5-6 steps slightly brighter (because they are "active") while drawing opcodes that will not be executed slightly darker? This would also make it visible whether or not a jump is taken and whether an opcode is a stop opcode or not.

PR #1324, opened 2021-06-24
move compiler to core plugins

bunsenstraat:
Plugins that are not views and have independent use from the IDE can be moved out of IDE and published as separate npm packages. To do this I had to
- remove dependency on globalregistry, replace it by plugin calls
- remove constructor args as much as possible, and replace them by plugin calls
- create nx, workspace entries
- I also changed settings to return any key/value storage through the API, this.call('settings','get',key)
- certain libs had to be moved to remix-solidity

In this use case I moved files related to compiler, to use them in VSCODE
- CompilerImports
- CompilerArtefacts
- FetchAndCompile
- CompilerMetadata
- OffsetToLineColumnConverter
Also these helpers/libs have been moved:
- CompilerAbstract
- compiler helpers
- compiler utils
There are also small updates to remix-lib & url-resolver


apps/remix-ide/src/app.js
```diff
@@ -16,7 +16,12 @@ import { HiddenPanel } from './app/components/hidden-panel'
 import { VerticalIcons } from './app/components/vertical-icons'
 import { LandingPage } from './app/ui/landing-page/landing-page'
 import { MainPanel } from './app/components/main-panel'
-import FetchAndCompile from './app/compiler/compiler-sourceVerifier-fetchAndCompile'
+
+import { CompilerArtefacts } from '@remix-core-plugin/compiler-artefacts'
```
yann300, line 20: Perhaps this need to be prefixed with `@remix-project/` instead of `@remix-core-plugin`.
`remix-project` is the name of the org in npm.
If this is just an alias then I guess this is ok
@ioedeveloper @Aniket-Engg  any thoughts?

yann300:
could you add a more descriptive message about what you want to achieve with this PR.
I think other reviewers might miss some context.

apps/remix-ide/src/app.js
bunsenstraat: Sure I did it this way because you suggested it but maybe you meant something else.

yann300: sure, I mean that `@remix-core-plugin` don't rely to any npm organization we own, so I don't see how you can publish that.
I was thinking that this could be something like:
`import { CompilerArtefacts } from '@remix-project/core-plugin'`

bunsenstraat: Ok, so then we bundle all the plugins in core-plugin package?

bunsenstraat:
> could you add a more descriptive message about what you want to achieve with this PR.
> I think other reviewers might miss some context.

@yann300 @LianaHus I added a description to explain

nx.json
```diff
@@ -101,6 +101,22 @@
     },
     "remix-ui-checkbox": {
       "tags": []
+    },
+    "compiler-artefacts": {
```
ioedeveloper, line 105: Add library names to lint command in package.json https://github.com/ethereum/remix-project/blob/master/package.json#L44

ioedeveloper (changes requested):

apps/remix-ide/src/app/tabs/compile-tab.js
```diff
@@ -2,7 +2,7 @@
 import { ViewPlugin } from '@remixproject/engine-web'
 import * as packageJson from '../../../../../package.json'
 import publishToStorage from '../../publishToStorage'
-import { compile } from '../compiler/compiler-helpers'
```
ioedeveloper, line 5: Will your changes to import path and plugin calls affect my react solidity compiler move PR?

bunsenstraat: I think so yes. I'm going to move everything now to @remix-project/core-plugin as suggested by @yann300

nx.json
bunsenstraat: ok

apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -22,7 +22,7 @@ const profile = {
   icon: 'assets/img/fileManager.webp',
   permission: true,
   version: packageJson.version,
-  methods: ['file', 'exists', 'open', 'writeFile', 'readFile', 'copyFile', 'copyDir', 'rename', 'mkdir', 'readdir', 'remove', 'getCurrentFile', 'getFile', 'getFolder', 'setFile', 'switchFile', 'refresh'],
+  methods: ['file', 'exists', 'open', 'writeFile', 'readFile', 'copyFile', 'copyDir', 'rename', 'mkdir', 'readdir', 'remove', 'getCurrentFile', 'getFile', 'getFolder', 'setFile', 'switchFile', 'refresh', 'getProviderOf', 'getProviderByName'],
```
yann300, line 25: `getProviderOf` and `getProviderByName` returns a complex object and won't pass through the iframe. but I guess it's fine  if we only plan to use it for internal usage.

bunsenstraat: that is the case yes. There was no other way...

bunsenstraat: I could also expose all the provider methods of course...

yann300 (approved):

PR #1334, opened 2021-06-28
Debugging against various fork

yann300:


apps/remix-ide/src/blockchain/execution-context.js
```diff
@@ -146,7 +143,7 @@ export class ExecutionContext {
         this.askPermission()
         this.executionContext = context
         web3.setProvider(injectedProvider)
-        this._updateBlockGasLimit()
+        this._updateChainContext()
```
bunsenstraat, line 146: No need to wait for this? because it affects currentfork?

ioedeveloper (approved):

apps/remix-ide/src/blockchain/execution-context.js
yann300: good point, will check that asap

yann300: That works ok like this, but it's better to `await`. please see the pr https://github.com/ethereum/remix-project/pull/1340

Issue #587, opened 2020-11-17
Make the solidity compiler a React component


yann300:

yann300:
- [ ] ModaDialog
 - [ ] Toaster

PR #1339, opened 2021-06-29
React compiler

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/587


libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
```diff
@@ -9,7 +9,7 @@ import { Renderer } from '@remix-ui/renderer'
 import './css/style.css'
 
 export const SolidityCompiler = (props: SolidityCompilerProps) => {
-  const { editor, config, queryParams, plugin, compileTabLogic, compiledFileName, fileProvider, fileManager, contractsDetails, setHardHatCompilation, contractMap, compileErrors } = props
+  const { editor, config, queryParams, plugin, compileTabLogic, compiledFileName, fileProvider, fileManager, contractsDetails, setHardHatCompilation, contractMap, compileErrors, isHardHatProject } = props
```
bunsenstraat, line 12: Can't we do it so that all these can be called as plugins? instead of having them in the constructors or props.

yann300:
got this error:
```
Uncaught TypeError: Cannot read property 'resolveAndSave' of undefined
    at Compiler.handleImportCall (main.js:25735)
    at Compiler.gatherImports (main.js:17040)
    at Compiler.internalCompile (main.js:16675)
    at Compiler.onCompilationFinished (main.js:16792)
    at Worker.<anonymous> (main.js:16962)
```
after having compiled a contract which contain an non existing import:
```
// SPDX-License-Identifier: GPL-3.0

import "notexisting.sol";

contract Storage {
}

```

LianaHus:
You have some commented out code there. Should those be removed?

ioedeveloper:
> You have some commented out code there. Should those be removed?

I can take them out. They're no longer needed.

LianaHus (approved):

PR #1356, opened 2021-07-05
Add solidity compile custom action

yann300:
It doesn't necessarily need to be merged now. I've put some comment regarding the QA document regarding that.


bunsenstraat (approved):
If you use 1 type to filter it works, so it's fine

yann300:
> If you use 1 type to filter it works, so it's fine

you mean that the action doesn't appear if it's not a sol file? feel free to push changes in this PR anyway

bunsenstraat:
Yes only one filter counts. Not two.

PR #1363, opened 2021-07-07
Fix react warning in settings tab

ioedeveloper:


libs/remix-ui/settings/src/lib/remix-ui-settings.tsx
```diff
@@ -71,39 +71,47 @@ export const RemixUiSettings = (props: RemixUiSettingsProps) => {
...
-          </label>
+  const generalConfig = () => {
+    const isMetadataChecked = (props.config.get('settings/generate-contract-metadata') === null) || (props.config.get('settings/generate-contract-metadata') === undefined) ? false : props.config.get('settings/generate-contract-metadata')
+    const isEthereumVMChecked = (props.config.get('settings/always-use-vm') === null) || (props.config.get('settings/always-use-vm') === undefined) ? false : props.config.get('settings/always-use-vm')
+    const isEditorWrapChecked = (props.config.get('settings/text-wrap') === null) || (props.config.get('settings/text-wrap') === undefined) ? false : props.config.get('settings/text-wrap')
+    const isPersonalChecked = (props.config.get('settings/personal-mode') === null) || (props.config.get('settings/personal-mode') === undefined) ? false : props.config.get('settings/personal-mode')
+    const isMatomoChecked = (props.config.get('settings/matomo-analytics') === null) || (props.config.get('settings/matomo-analytics') === undefined) ? false : props.config.get('settings/matomo-analytics')
```
bunsenstraat, line 79: any reason why not like this
```
    const isMetadataChecked = props.config.get('settings/generate-contract-metadata') || false
    const isEthereumVMChecked = props.config.get('settings/always-use-vm') || false
    const isEditorWrapChecked = props.config.get('settings/text-wrap') || false
    const isPersonalChecked = props.config.get('settings/personal-mode') || false
    const isMatomoChecked = props.config.get('settings/matomo-analytics') || false
```

ioedeveloper: Yeah, good question!

bunsenstraat (approved):

PR #1365, opened 2021-07-07
Do not try removing if it doesn't exist

bunsenstraat:


ioedeveloper (approved):

Aniket-Engg (commented):
This still shows the error in console while trying locally

bunsenstraat:
The solution isn't working for tab-proxy. It only happens when you have files opened that you are deleting. Because it tries to re-open the files. This solution doesn't work. But it doesn't block the UI.

PR #1371, opened 2021-07-12
Fix value formating check

yann300:
Ganache CLI returns a sligthly different format for the transaction trace (some values aren't prefixed with 0x).


bunsenstraat (approved):

yann300:
for reproducing it:
 - debug `Ballot / delegate` with ganache-cli

Issue #1338, opened 2021-06-29
URL parameter for importing code from Github


dmitrydao:
## Overview
This is a feature request. 
Add extra parameter to the URL when importing code from Github.

## Issue

Remix allows importing Solidity code from Github but it doesn't add an extra parameter to the URL after importing. This makes it impossible to create a link that people can use to open a code sample in Remix.

Our team currently uses gists for that but it doesn't scale well when you have a large team. What happens is that multiple people own different gists and it makes it hard to have a centralized way to update different code snippets. Also, sometimes you need to update the code fast but the owner of a particular gist might be unavailable, on vacation, or even no longer working in the company. We need to fix that.

## Solution

Implement the same logic for importing Github code as gists by adding an additional parameter, such as`&github=https://github.com/documentation_repo/example.sol`. 

The end result should look the following way: `https://remix.ethereum.org/#optimize=false&runs=200&evmVersion=null&version=soljson-v0.8.4+commit.c7e474f2.js&github=https://github.com/documentation_repo/example.sol`

## Use cases
The best use case for this is the developer portal. Our team wants to update code samples once and make developers open them in Remix and load that code in our dev portal. In this case, we have to maintain only one version of the code which will be in the documentation repository.

yann300:
ok we'll have a look to it.
in the meantime, what do you think of https://remix-ide.readthedocs.io/en/latest/locations.html?highlight=code%20sample#load-an-encoded-base64-string-into-a-sol-file
is that answering your need?

dmitrydao:
Appreciate you pick it up. I think your solution will require encoding all of the code samples, which is not exactly what we're looking for and is not a scalable way for the long term. 

It will actually require more work from us, like encoding each time after the code samples have been changed and replacing that value with a new one after each update. 

But thanks for the suggestion anyway. I do hope you can figure out the way I outlined in the original issue.

PR #1376, opened 2021-07-13
Handle url param

yann300:
fix https://github.com/ethereum/remix-project/issues/1338

### DIFF
--- a/apps/remix-ide-e2e/src/tests/url.spec.ts
+++ b/apps/remix-ide-e2e/src/tests/url.spec.ts
@@ -17,7 +17,7 @@ module.exports = {
     return sources
   },
 
-  'Should load the code from URL params': function (browser: NightwatchBrowser) {
+  'Should load the code from URL params (code param)': function (browser: NightwatchBrowser) {
     browser
       .pause(5000)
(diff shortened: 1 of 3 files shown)


Aniket-Engg (commented):
Can you please add e2e tests for this?

Also, with this implementation, if `code` and `url` both are in params,  I am not sure if file with `code` param will be created.

yann300:
>  I am not sure if file with code param will be created.

this should be created but filemanager won't switch to it.

Aniket-Engg (commented):
This would be great if you can add an e2e with `code` and `url` both fields at once in URL

yann300:
in order to do manual testing:
- load http://127.0.0.1:8080/#optimize=true&runs=300&evmVersion=istanbul&version=soljson-v0.7.4+commit.3f05b770.js&url=https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/contracts/app/solidity/mode.sol
- make sure `mode.sol` is loaded and switched to.

Aniket-Engg (approved):
This looks fine now.

Just to add, no matter in which sequence `code` and `url` are assigned in URL, when both are in URL, editor will always open the file from `url` param

PR #1382, opened 2021-07-14
Highlight the last 5 items instead of only one

yann300:
fix https://github.com/ethereum/remix-project/issues/1155
This PR is addressing https://github.com/ethereum/remix-project/pull/1315#issuecomment-872184565


yann300:
for manual testing:

 - deploy a contract
 - start debugging
 - expected: the current opcode should be highlighted
 - expected: the next 5 opcodes should be set with a different color
 - move forward in the execution
 - expected: when the execution is about to reach a JUMP, it should be obvious (seeing the 5 next opcodes) that a JUMP is about to be taken.

Aniket-Engg (approved):
Looks fine to me

PR #1387, opened 2021-07-15
Fix Loading with code or url param

yann300:
In that situation, the settings tab activates and get deactivated.
The issue apparently due to an issue with the ordering of system plugin activation.


bunsenstraat:
test with http://127.0.0.1:8080/#optimize=true&runs=300&evmVersion=istanbul&version=soljson-v0.7.4+commit.3f05b770.js&url=https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/contracts/app/solidity/mode.sol

bunsenstraat (approved):

PR #1401, opened 2021-07-26
fix static analysis event listen issue

### LINKED ISSUE #1400: Static Analysis Plugin Not working After Deactivation and Re-activation
#### Steps To Reproduce
- compile file 
- analysis will be shown
- switch workspace
- decativate and activate again
- then compile again
**compilation event doesn’t seem to be received by the plugin**

Aniket-Engg:
fixes #1400

### DIFF
--- a/libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
+++ b/libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
@@ -65,7 +65,7 @@ export const RemixUiStaticAnalyser = (props: RemixUiStaticAnalyserProps) => {
 
   useEffect(() => {
     compilation(props.analysisModule, dispatch)
-  }, [])
+  }, [props])
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
```diff
@@ -65,7 +65,7 @@ export const RemixUiStaticAnalyser = (props: RemixUiStaticAnalyserProps) => {
 
   useEffect(() => {
     compilation(props.analysisModule, dispatch)
-  }, [])
+  }, [props])
```
ioedeveloper, line 68: I think it is better to listen on props.analysisModule instead.

Aniket-Engg: I tried, this `useEffect` was not triggered with that too

ioedeveloper (approved):

PR #1428, opened 2021-07-30
info icon added for External integration tools usage

### LINKED ISSUE #1418: Warn users about the fact that they need to install hardhat and slither
Activating slither and hardhat is pretty straight forward,
but it requires the user to install locally both app, which is not clear presented in the current interface.

Aniket-Engg:
Fix #1418

### DIFF
--- a/libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
+++ b/libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useReducer } from 'react'
+import React, { useEffect, useState, useReducer, CSSProperties } from 'react'
 import Button from './Button/StaticAnalyserButton' // eslint-disable-line
 import remixLib from '@remix-project/remix-lib'
 import _ from 'lodash'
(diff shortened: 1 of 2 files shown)


libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
```diff
@@ -401,16 +401,16 @@ export const RemixUiStaticAnalyser = (props: RemixUiStaticAnalyserProps) => {
...
           <RemixUiCheckbox
             id="enableSlither"
             inputType="checkbox"
             onClick={handleSlitherEnabled}
             checked={slitherEnabled}
             label="Enable Slither Analysis"
             onChange={() => {}}
-            visibility = {showSlither}
           />
+          <a href="https://remix-ide.readthedocs.io/en/latest/slither.html#enable-slither-analysis" target="_blank"><i className="ml-2 fas fa-info" title="Know how to use Slither Analysis"></i></a>
```
LianaHus, line 413: small notes from me:
ml-3 looks better to me.
I would not put two tags at the same line

LianaHus (approved):

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
Aniket-Engg: ok

Issue #1433, opened 2021-08-03
Hometab import from GITHUB & from HTTPS don't work


ryestew:
Take a file like the examples in the modal or like this file - for github import:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/finance/PaymentSplitter.sol

or like the example in https and it doesn't work

ioedeveloper:
@ajsarun Your issue raised in the gitter channel is being tracked here. https://github.com/ethereum/remix-project/issues/1433

Aniket-Engg:
A similar one: https://files.gitter.im/578fa63cc2f0db084a23e12f/z9sd/image.png

PR #1437, opened 2021-08-03
Fixed Github import from homepage

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/1433

### DIFF
--- a/libs/remix-core-plugin/src/lib/compiler-content-imports.ts
+++ b/libs/remix-core-plugin/src/lib/compiler-content-imports.ts
@@ -22,7 +22,12 @@ export class CompilerImports extends Plugin {
 
   async setToken () {
     const protocol = typeof window !== 'undefined' && window.location.protocol
-    const token = await this.call('settings', 'get', 'settings/gist-access-token')
+    let token
+    try {
+      token = await this.call('settings', 'get', 'settings/gist-access-token')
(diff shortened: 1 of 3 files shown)


LianaHus (approved):

Aniket-Engg (approved):

Aniket-Engg (commented):
Loaded contract can not be compiled. It seems it doesn't fetch all the imported files to the file explorer

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -22,7 +22,12 @@ export class CompilerImports extends Plugin {
 
   async setToken () {
     const protocol = typeof window !== 'undefined' && window.location.protocol
-    const token = await this.call('settings', 'get', 'settings/gist-access-token')
+    let token
+    try {
+      token = await this.call('settings', 'get', 'settings/gist-access-token')
```
yann300, line 27: if this fails, you might not want to do `this.urlResolver.setGistToken(token, protocol)`

yann300 (approved):

yann300:
That fix can go in for now.
There's another related issue: if one import a github files which contain relative import, those won't be resolved.
Already discussed with @LianaHus about a possible fix for that.

Issue #1455, opened 2021-08-10
Debugger panel entirely disappears during debugging session.


grapevinegizmos:
Trying to debug a contract (which has some obvious bugts), the debugger pane entirely disappears.

No controls, just a blank panel.  

A picture attached, plus the code I was tyring to debug.
<img width="2254" alt="Screen Shot 2021-08-09 at 10 08 07 PM" src="https://user-images.githubusercontent.com/8684037/128811327-ff005667-81d5-4b3f-bb55-f5ddedc857f9.png">
`pragma solidity 0.8.4;
 

```
contract Bindings {
    string constant StartMark = "${";
    string constant EndMark = "}";
    address immutable owner;

    enum BindingType {
        Constraint,
        Entitlement,
        Undertaking,
        Representation,
        Template
    }
```

    struct Binding {
        string id;
        string name;
        uint256 version;
        string bindingText;
        string[] templateNames;
        BindingType bindingType;
        string author;
        bool enabled;
        uint256 timestamp;
    }

    //Maps id to a binding struct.  The id is generated  from the hash of the text.
    mapping(string => Binding) bindings;
    //mapping by name maps names to ids
    mapping(string => string[]) bindingsByName;

    constructor() {
        owner = msg.sender;

    }

    function createBinding(
        string calldata name,
        string calldata bindingText,
     //   BindingType bindingType,
        string calldata author
    //    string[] calldata templateNames
    ) public {

        //compute the hash of the text as the id.
        bytes32  id = keccak256(abi.encodePacked(bindingText));
        string memory sid = string(abi.encodePacked(id));

        //find the number of existing binding with the same name.
        uint count = bindingsByName[name].length;
        //declar the for the bindings by name mapping (nameIds)
        string[] memory nameIds = bindingsByName[name];
        if (count == 0) {
            nameIds[0] = sid;
            bindingsByName[name] = nameIds;
        } else {
            bindingsByName[name].push(sid);
        }
        Binding memory binding;
        
        binding.author = author;
        binding.bindingText = bindingText;
        binding.bindingType = BindingType.Entitlement;
        binding.id = string(abi.encodePacked(id));
        binding.name = name;
        binding.templateNames[0]="Video";
        binding.templateNames[1]="Duration";
        binding.timestamp = block.timestamp;
        binding.version = nameIds.length;
        bindings[binding.id] = binding;
    }

    function getText(string calldata name) public view returns (string memory) {
        string[] memory bindingsForName = bindingsByName[name];
[...]

PR #1457, opened 2021-08-10
Fix debugger

yann300:
fix https://github.com/ethereum/remix-project/issues/1455


Aniket-Engg (approved):
Change set looks fine. Please check if some unit tests can be added in support.

PR #1459, opened 2021-08-10
Integrate hardhat console log

Aniket-Engg:


yann300:
- I think the log needs to be displayed before the transaction result in the terminal.
 - when logging `console.log(num, "okok");` the log is `3okok`. There should be some spacing.
 - we maybe don't need to display `console.log:`

Aniket-Engg:
> * I think the log needs to be displayed before the transaction result in the terminal.
> * when logging `console.log(num, "okok");` the log is `3okok`. There should be some spacing.
> * we maybe don't need to display `console.log:`

I tried to follow the pattern from Hardhat itself which is as:
```
Contract deployment: <UnrecognizedContract>
  Contract address:    0x5fbdb2315678afecb367f032d93f642f64180aa3
  Transaction:         0xf7b8ea6e00c3ddecaa65d364a78f5edddb543d2e65e65840bca3c39bb8643975
  From:                0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
  Value:               0 ETH
  Gas used:            499747 of 499747
  Block #1:            0x429662673bdd8281f818d1f8a7c266faa916003a1de95e8dc7aa4adf30ac1e15

  console.log:
    Deploying a Greeter with greeting: cycusas
    Deploying a Greeter with greeting:4342424 cycusas
```

Aniket-Engg:
> when logging console.log(num, "okok"); the log is 3okok. There should be some spacing.

I think that is normal logging. We should not customize it.

libs/remix-lib/src/web3Provider/web3VmProvider.ts
```diff
@@ -206,6 +210,29 @@ export class Web3VmProvider {
...
+      let payload = memory.substring(payloadStart * 2, memory.length)
+      const fnselectorStr = payload.substring(0, 8)
+      const fnselectorStrInHex = '0x' + fnselectorStr
+      const fnselector = parseInt(fnselectorStrInHex)
+      const fnArgs = ConsoleLogs[fnselector]
+      const iface = new ethers.utils.Interface([`function log${fnArgs} view`])
+      const functionDesc = iface.getFunction(`log${fnArgs}`)
+      const sigHash = iface.getSighash(`log${fnArgs}`)
+      if (fnArgs.includes('uint') && sigHash !== fnselectorStrInHex) {
+        payload = payload.replace(fnselectorStr, sigHash)
```
yann300, line 227: you don't need to prepend `0x` in that case?

Aniket-Engg: No, sigHash returned from `iface.getSighash(`log${fnArgs}`)` already contains it

yann300:
> > when logging console.log(num, "okok"); the log is 3okok. There should be some spacing.
> 
> I think that is normal logging. We should not customize it.

When running `console.log('test', 'test')` in the browser console, `test test` is displayed, not `testtest`.
does hardhat have a space? if not let's keep the same.

yann300:
> > * I think the log needs to be displayed before the transaction result in the terminal.
> > * when logging `console.log(num, "okok");` the log is `3okok`. There should be some spacing.
> > * we maybe don't need to display `console.log:`
> 
> I tried to follow the pattern from Hardhat itself which is as:
> 
> ```
> Contract deployment: <UnrecognizedContract>
>   Contract address:    0x5fbdb2315678afecb367f032d93f642f64180aa3
>   Transaction:         0xf7b8ea6e00c3ddecaa65d364a78f5edddb543d2e65e65840bca3c39bb8643975
>   From:                0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
>   Value:               0 ETH
>   Gas used:            499747 of 499747
>   Block #1:            0x429662673bdd8281f818d1f8a7c266faa916003a1de95e8dc7aa4adf30ac1e15
> 
>   console.log:
>     Deploying a Greeter with greeting: cycusas
>     Deploying a Greeter with greeting:4342424 cycusas
> ```

ok let's keep the same then

Aniket-Engg:
> > > when logging console.log(num, "okok"); the log is 3okok. There should be some spacing.
> > 
> > 
> > I think that is normal logging. We should not customize it.
> 
> When running `console.log('test', 'test')` in the browser console, `test test` is displayed, not `testtest`.
> does hardhat have a space? if not let's keep the same.

I checked. Hardhat has a space in between. Will do it.

yann300 (approved):

Issue #331, opened 2020-04-14
Bypass the remix debug nodes


yann300:
There should be an option to force the debug rpc calls to go through the current node instead of redirecting the call to our debug node (mainet net is currently down).
At least perhaps we can still use the current node if we detect that the node used for debugging is not available.

PR #1460, opened 2021-08-11
Force using local node

yann300:
fix https://github.com/ethereum/remix-project/issues/331

### DIFF
--- a/libs/remix-debug/src/debugger/debugger.ts
+++ b/libs/remix-debug/src/debugger/debugger.ts
@@ -99,38 +99,28 @@ export class Debugger {
     this.debugger.web3 = web3
   }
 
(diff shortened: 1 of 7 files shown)


libs/remix-debug/src/debugger/debugger.ts
```diff
@@ -100,9 +100,9 @@ export class Debugger {
   }
 
   debug (blockNumber, txNumber, tx, loadingCb): Promise<void> {
-    const web3 = this.debugger.web3
-
     return new Promise((resolve, reject) => {
```
yann300, line 103: make it `async`

yann300:
maybe we should check how this behave with ganache and a forked mainet

Aniket-Engg (approved):

Issue #1456, opened 2021-08-10
running hardhat on remix desktop


yann300:
Fix running hardhat with remix desktop : https://github.com/ethereum/remix-desktop/issues/67

PR #1465, opened 2021-08-12
Fix hardhat & slither using remix desktop

yann300:
fix https://github.com/ethereum/remix-project/issues/1456
fix https://github.com/ethereum/remix-desktop/issues/67

### DIFF
--- a/libs/remixd/src/services/slitherClient.ts
+++ b/libs/remixd/src/services/slitherClient.ts
@@ -149,13 +149,9 @@ export class SlitherClient extends PluginClient {
         if (existsSync(outputFileAbsPath)) {
           let report = readFileSync(outputFileAbsPath, 'utf8')
           report = JSON.parse(report)
-          try {
-            unlink(outputFileAbsPath, (err) => {
-              console.log(err)
-            })
-          } catch (e) {
-            console.log(e)
-          }
+          unlink(outputFileAbsPath, (err) => {
+            if (err) console.log(err)
+          })
           if (report['success']) {
             response['status'] = true
             if (!report['results'] || !report['results'].detectors || !report['results'].detectors.length) {
--- a/libs/remixd/src/services/slitherClient.ts
+++ b/libs/remixd/src/services/slitherClient.ts
@@ -125,16 +125,16 @@ export class SlitherClient extends PluginClient {
       const evmOption: string = evmVersion ? ` --evm-version ${evmVersion}` : ''
       let solcArgs = ''
       if (optimizeOption) {
-          solcArgs += ' ' + optimizeOption
+        solcArgs += ' ' + optimizeOption
       }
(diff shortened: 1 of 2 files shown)


libs/remixd/src/services/slitherClient.ts
```diff
@@ -114,21 +114,30 @@ export class SlitherClient extends PluginClient {
...
+      }
+      if (solcArgs) {
+        solcArgs = `--solc-args "${solcArgs.trimStart()}"`
+      }
       const solcRemaps = remaps ? `--solc-remaps "${remaps}"` : ''
 
       const outputFile: string = 'remix-slitherReport_' + Math.floor(Date.now() / 1000) + '.json'
       const cmd: string = `slither ${filePath} ${solcArgs} ${solcRemaps} --json ${outputFile}`
       console.log('\x1b[32m%s\x1b[0m', '[Slither Analysis]: Running Slither...')
+      console.log(cmd)
```
Aniket-Engg, line 140: This can be removed

libs/remixd/src/services/slitherClient.ts
```diff
@@ -140,6 +149,13 @@ export class SlitherClient extends PluginClient {
         if (existsSync(outputFileAbsPath)) {
           let report = readFileSync(outputFileAbsPath, 'utf8')
           report = JSON.parse(report)
+          try {
+            unlink(outputFileAbsPath, (err) => {
```
Aniket-Engg, line 153: We can also use `unlinkSync`

yann300: we don't really need to wait for the result.

Aniket-Engg: ok

libs/remixd/src/services/slitherClient.ts
```diff
@@ -140,6 +148,13 @@ export class SlitherClient extends PluginClient {
...
```
Aniket-Engg, line 152: not sure if we use try-catch with callback

Aniket-Engg (approved):

bunsenstraat (approved):

libs/remixd/src/services/slitherClient.ts
```diff
@@ -140,6 +148,9 @@ export class SlitherClient extends PluginClient {
         if (existsSync(outputFileAbsPath)) {
           let report = readFileSync(outputFileAbsPath, 'utf8')
           report = JSON.parse(report)
+          unlink(outputFileAbsPath, (err) => {
+            console.log(err)
```
bunsenstraat, line 152: if (err) console.log(err)
otherwise it always prints null everytime

bunsenstraat (changes requested):

bunsenstraat (approved):

PR #1475, opened 2021-08-13
updated tx type

### LINKED ISSUE #1470: Deploying on mainnet with MetaMask errors
Deploying on testnets seem to use legacy transactions, but the mainnet seems updated to 1559 dynamic txs. Unfortunately, it seems the call does not adhere to the specs:

- It sends a `type` parameter along with the tx, which I think is a noop. Only the return from nodes contain the type (unless maybe MetaMask needs it?)
- The type of the `type` sent is a decimal (`2` in this specific case), but MetaMask expects all fields to be strings (hex): https://metamask.github.io/metamask-extension/controllers_transactions_tx-state-manager.js.html#line207

Aniket-Engg:
fixes #1470


LianaHus (approved):

yann300:
added in https://github.com/ethereum/remix-project/pull/1476

Issue #673, opened 2020-12-10
remix-tests/Solidity unit testing plugin Improvements


Aniket-Engg:
- [ ] Custom compiler context for remix-tests CLI (#540 )
- [ ] Dynamic tests generation based on contract (Research needed)
- [ ] Support for logging in test file
- [ ] Enable debugging for unit testing transactions

PR #1496, opened 2021-08-18
hhlogs in remix-tests

Aniket-Engg:
Related to #673


libs/remix-tests/src/testRunner.ts
```diff
@@ -284,59 +284,63 @@ export function runTest (testName: string, testObject: any, contractDetails: Com
...
-            for (const event of events) {
-              const eIndex = assertionEventHashes.indexOf(event.raw.topics[0]) // event name topic will always be at index 0
-              if (eIndex >= 0) {
-                const testEvent = web3.eth.abi.decodeParameters(assertionEvents[eIndex].params, event.raw.data)
-                if (!testEvent[0]) {
-                  const assertMethod = testEvent[2]
-                  if (assertMethod === 'ok') { // for 'Assert.ok' method
-                    testEvent[3] = 'false'
-                    testEvent[4] = 'true'
+          web3.eth.getHHLogsForTx(receipt.transactionHash).then((hhLogs) => {
```
yann300, line 287: could you put all this in a separate file? also maybe using `await` instead of promise.

apps/remix-ide/src/app/tabs/test-tab.js
```diff
@@ -180,6 +180,7 @@ module.exports = class TestTab extends ViewPlugin {
   }
 
   testCallback (result, runningTests) {
+    console.log('result in testCallback', result)
```
yann300, line 183: should be remove

libs/remix-tests/src/runTestSources.ts
```diff
@@ -5,7 +5,7 @@ import { deployAll } from './deployer'
 import { runTest } from './testRunner'
 
 import Web3 from 'web3'
-import { Provider } from '@remix-project/remix-simulator'
```
yann300, line 8: these changes concern `runTestSources`, should we do the same for `runTestFiles`?

apps/remix-ide/src/app/tabs/test-tab.js
Aniket-Engg: This is there to discuss with Liana

libs/remix-tests/src/runTestSources.ts
Aniket-Engg: I am not sure if external import works fine with the `runTestFiles`

yann300: which resolver is it used in that case?

libs/remix-tests/tests/testRunner.spec.ts
```diff
@@ -72,13 +77,15 @@ async function compileAndDeploy(filename: string, callback: Function) {
       }
     }
   ], function (_err: Error | null | undefined, contracts: any): void {
-    callback(null, compilationData, contracts, sourceASTs, accounts)
+    console.log('Outside compileAndDeploy')
```
LianaHus, line 80: forgot to remove

LianaHus: also above

apps/remix-ide/src/app/tabs/test-tab.js
Aniket-Engg: Removed now

libs/remix-tests/src/runTestSources.ts
Aniket-Engg: Probably none

libs/remix-tests/tests/testRunner.spec.ts
Aniket-Engg: removed

yann300 (approved):

PR #1502, opened 2021-08-19
hardhat logs formatting improved

Aniket-Engg:


apps/remix-ide/src/blockchain/blockchain.js
```diff
@@ -501,7 +501,7 @@ class Blockchain extends Plugin {
       if (isVM) {
         const hhlogs = await this.web3().eth.getHHLogsForTx(txResult.transactionHash)
         if (hhlogs && hhlogs.length) {
-          let finalLogs = 'console.log:\n'
+          let finalLogs = '<b>console.log:</b>\n'
```
yann300, line 504: would you do the same for the unit tests log?

yann300 (approved):

apps/remix-ide/src/blockchain/blockchain.js
Aniket-Engg: It is done there before here: https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/tabs/test-tab.js#L184

Issue #1016, opened 2021-03-23
Activate react production build when deploying to remix_live


ioedeveloper:

HarikrishnanBalagopal:
If you have react developer tools extension installed then you will see this message
![image](https://user-images.githubusercontent.com/20921177/117576631-3c9f7200-b104-11eb-8a36-34dd3ee3a390.png)

PR #1503, opened 2021-08-20
Enable react production build

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/1016

### DIFF
--- a/apps/debugger/webpack.config.js
+++ b/apps/debugger/webpack.config.js
@@ -3,18 +3,8 @@ const TerserPlugin = require('terser-webpack-plugin')
 
 module.exports = config => {
   const nxWebpackConfig = nxWebpack(config)
-  const productionConfig = process.env.NODE_ENV === 'production' ? {
-    mode: 'production',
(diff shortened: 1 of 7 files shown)


apps/debugger/webpack.config.js
```diff
@@ -1,17 +1,27 @@
...
-            fs: 'empty',
-            tls: 'empty',
-            readline: 'empty',
-            net: 'empty',
-            module: 'empty',
-            child_process: 'empty'
-        }
+  return {
+    ...nxWebpackConfig,
+    ...productionConfig,
```
Aniket-Engg, line 17: Should we put a check on return object to return `productionConfig` only if environment is production? That can be better from returning even `{}`

ioedeveloper: Ok. I'll make the change.

Mehdish1414:
Be l to

Aniket-Engg (approved):

yann300 (approved):

PR #1506, opened 2021-08-23
solidity default version updated to 0.8.7

Aniket-Engg:


yann300 (approved):

LianaHus (changes requested):
shouldn't you also update the local version?

Aniket-Engg:
> shouldn't you also update the local version?

Local solc version is dynamically placed in `assets` using https://github.com/ethereum/remix-project/blob/master/.circleci/config.yml#L241

LianaHus (approved):

PR #1507, opened 2021-08-23
Sol errors

LianaHus:


yann300:
does this concern compiler warning, static analysis warning or both?

yann300:
ok thanks @ioedeveloper cause I was checking my PR https://github.com/ethereum/remix-project/pull/1509 (it concern static analysis but also span to the compiler warning), there was an error when clicking on a warning that link to an imported source file, I was hoping this PR fixes this so I can double check with my PR.

libs/remix-ui/renderer/src/lib/renderer.tsx
```diff
@@ -56,11 +50,14 @@ export const Renderer = ({ message, opt = {}, plugin }: RendererProps) => {
     setClose(false)
   }, [message, opt])
 
-  const getPositionDetails = (msg: any) => {
+  const getPositionDetails = (msg: string) => {
     const result = { } as Record<string, number | string>
 
     // To handle some compiler warning without location like SPDX license warning etc
-    if (!msg.includes(':')) return { errLine: -1, errCol: -1, errFile: msg }
+    if (!msg.includes(':')) return { errLine: -1, errCol: -1, errFile: '' }
```
Aniket-Engg, line 57: Please check if a case if possible where msg includes `-->`, but not `:`. In that case, it will be returned from here itself

libs/remix-ui/renderer/src/lib/renderer.tsx
```diff
@@ -56,11 +50,14 @@ export const Renderer = ({ message, opt = {}, plugin }: RendererProps) => {
...
+
+    if (!msg.includes('-->')) return { errLine: -1, errCol: -1, errFile: '' }
```
Aniket-Engg, line 59: Not sure if this statement can be combined with the above one using `||`

Aniket-Engg: Also, I think for old compiler, msg will not be including `-->`

yann300:
thanks for that peace of code @LianaHus , it is back to https://github.com/ethereum/remix-project/pull/1509 please review it

Issue #1393, opened 2021-07-19
Location and code highlighting for imported contract in Solidity Static Analysis with localhost workspace


Aniket-Engg:
- [ ] For Remix Analyzer, it doesn't refer the file in .deps or node_modules directory, instead it create a folder and file in root of project and then refers to highlight the code
- [ ] For Slither, for imported contract, location is `not available`

Aniket-Engg:
This is an attempt: https://github.com/ethereum/remix-project/pull/1383

Issue #1408, opened 2021-07-27
Solidity Compiler gives wrong source location


LianaHus:
In order to reproduce:
- put a solidity version 0.8.4
- compile import "https://github.com/aave/protocol-v2/blob/ice/mainnet-deployment-03-12-2020/contracts/protocol/tokenization/StableDebtToken.sol";

go to the Compiler plugin and click on pragma error, for highlighting the StableDebtToken.sol should be opened.

similar to https://github.com/ethereum/remix-project/issues/846

PR #1509, opened 2021-08-23
Fix source location in npm

yann300:
fix https://github.com/ethereum/remix-project/issues/1393


LianaHus:
can we have any e2e to check this?

yann300:
@LianaHus This PR also handles https://github.com/ethereum/remix-project/pull/1507, I had to get it over to my pr and needed to fix an issue regarding how your PR handled solidity errors , afaik it doesn't seem to work with previous compilers (those that aren't having an `--->`)

Aniket-Engg (commented):
Regarding https://github.com/ethereum/remix-project/issues/1408 , I see, on clicking `pragma error`, it opens the `StableDebtToken.sol` file but highlights line 1. When click on it again, it highlights line 2 having pragma statement which is fine.

Ideally it should highlight line 2 at first attempt itself.

Mehdish1414:
Mehdish14

Aniket-Engg (commented):
Similar for #1393 , on first attempt it highlights first line and then the correct location.

Also, it shows lots of errors in browser console, it would be good if they can be avoided

extraConsoles

apps/remix-ide-e2e/src/tests/remixd.test.ts
```diff
@@ -71,6 +78,20 @@ module.exports = {
...
       .testContracts('test_import_node_modules_with_github_import.sol', sources[4]['test_import_node_modules_with_github_import.sol'], ['ERC20', 'test11'])
   },
+  'Static Analysis run with remixd': function (browser) {
+    browser.testContracts('test_static_analysis_with_remixd_and_hardhat.sol', sources[5]['test_static_analysis_with_remixd_and_hardhat.sol'], ['test5'])
+      .clickLaunchIcon('solidityStaticAnalysis')
+      .click('#staticanalysisButton button')
+      .waitForElementPresent('#staticanalysisresult .warning', 2000, true, function () {
+        browser
+          .click('[data-id="staticAnalysisModuleMiscellaneous1"')
+          .getEditorValue((content) => {
```
Aniket-Engg, line 88: Does it also check if the required location is highlighted?

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -115,66 +116,58 @@ export class CompilerImports extends Plugin {
...
-                      if (error) return reject(error)
-                      resolve(content)
-                    })
-                  }
-                  resolve(result)
-                })
-              }
-              this.importExternal(url, targetPath, (error, content) => {
+        if (exist) {
+          const content = await (() => {
```
Aniket-Engg, line 136: This seems a bit complex code. May be it can be simplified

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -88,21 +87,23 @@ export class CompilerImports extends Plugin {
...
-        try {
-          const provider = await this.call('fileManager', 'getProviderOf', null)
-          const path = targetPath || type + '/' + cleanUrl
-          if (provider) provider.addExternal('.deps/' + path, content, url)
-        } catch (err) {
-
-        }
-        cb(null, content)
-      }, null)
+  importExternal (url, targetPath) {
```
Aniket-Engg, line 90: A clean approach may be making this function itself `async`

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -115,66 +116,58 @@ export class CompilerImports extends Plugin {
...
             })
-          }).catch(error => {
-            return reject(error)
-          })
+          })()
+          return content
+        } else {
+          const localhostProvider = await this.call('fileManager', 'getProviderByName', 'localhost')
+          if (localhostProvider.isConnected()) {
+            var splitted = /([^/]+)\/(.*)$/g.exec(url)
```
Aniket-Engg, line 148: use `const`

libs/remix-ui/renderer/src/lib/renderer.tsx
```diff
@@ -29,25 +29,19 @@ export const Renderer = ({ message, opt = {}, plugin }: RendererProps) => {
     // ^ e.g:
     // browser/gm.sol: Warning: Source file does not specify required compiler version! Consider adding "pragma solidity ^0.6.12
     // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v3.2.0/contracts/introspection/IERC1820Registry.sol:3:1: ParserError: Source file requires different compiler version (current compiler is 0.7.4+commit.3f05b770.Emscripten.clang) - note that nightly builds are considered to be strictly less than the released version
-    let positionDetails = getPositionDetails(text)
+    const positionDetails = getPositionDetails(text)
     const options = opt
```
Aniket-Engg, line 33: Is this assignment/copy required?

apps/remix-ide-e2e/src/tests/remixd.test.ts
yann300: this should be done now

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
yann300: The issue here is that the provider doesn't return a promise but uses callback, but we need this call to `await`...

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
yann300: same issue as above, `import` rely mostly on a callback. This can certainly be done but not sure we have place for that in this PR.

apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -170,9 +170,14 @@ class FileManager extends Plugin {
   async open (path) {
     try {
       path = this.limitPluginScope(path)
+      try {
+        path = this.getPathFromUrl(path).file
+      } catch (e) {
+        return console.error(e)
```
ioedeveloper, line 176: throw new Error(e.message)

apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -538,6 +543,36 @@ class FileManager extends Plugin {
...
   }
 
+  /**
+   * Try to resolve the given file path (the actual path in the file system)
+   * e.g if it's specified a github link, npm library, or any external content,
+   * it returns the actual path where the content can be found.
+   * @param {string} file url we are trying to resolve
+   * @returns {{ string, provider }} file path resolved and its provider.
+   */
+  getPathFromUrl (file) {
```
ioedeveloper, line 553: should remixdProvider also have an implementation for this?

apps/remix-ide/src/app/files/fileProvider.js
```diff
@@ -35,7 +36,12 @@ class FileProvider {
   }
 
   getPathFromUrl (url) {
-    return this.providerExternalsStorage.get('reverse-' + url)
+    return this.providerExternalsStorage.get(this.reverseKey + url)
+  }
+
+  getUrlFromPath (path) {
```
ioedeveloper, line 42: should remixdProvider also have an implementation for this?

yann300: it already has, remixdProvider has fileProvider has base class.

ioedeveloper (approved):

PR #1517, opened 2021-08-25
check if remixd shared folder exists

Aniket-Engg:


libs/remixd/src/bin/remixd.ts
```diff
@@ -83,7 +83,7 @@ function errorHandler (error: any, service: string) {
     console.log('\x1b[33m%s\x1b[0m', '[WARN] You may now only use IDE at ' + program.remixIde + ' to connect to that instance')
   }
 
-  if (program.sharedFolder) {
+  if (program.sharedFolder && existsSync(absolutePath('./', program.sharedFolder))) {
```
yann300, line 86: `sharedFolder` can be either absolute path or relative path, is this test taking this in account?

libs/remixd/src/bin/remixd.ts
```diff
@@ -102,7 +102,7 @@ function errorHandler (error: any, service: string) {
       })
       // Run hardhat service if a hardhat project is shared as folder
       const hardhatConfigFilePath = absolutePath('./', program.sharedFolder) + '/hardhat.config.js'
```
ioedeveloper, line 104: Is hardhat path limited to relative path?

libs/remixd/src/bin/remixd.ts
bunsenstraat: I have tested this and it seems fine to me

bunsenstraat (approved):
I ran some tests with it, relative, absolute. seems fine to me.

Aniket-Engg:
> I ran some tests with it, relative, absolute. seems fine to me.

Cool Thanks

PR #1519, opened 2021-08-25
ipfs pull fallbacks in dgit provider

bunsenstraat:
it's unclear sometimes which IPFS node will response with data. I've seen 'too many requests' errors on ipfs.io and so on... 
so this picks the first one that has the files. sometimes the hash is not available yet on any of the nodes. this can still happen if you push to pinata. it takes a few minutes before public nodes have the hash. it also adds a timeout for the request. 
it also returns a warning when the local storage is full, which happens more quickly than you'd think.

### DIFF
--- a/apps/remix-ide/src/app/files/dgitProvider.js
+++ b/apps/remix-ide/src/app/files/dgitProvider.js
@@ -38,6 +38,13 @@ class DGitProvider extends Plugin {
       protocol: 'https',
       ipfsurl: 'https://ipfs.io/ipfs/'
     }
+    this.remixIPFS = {
+      host: 'ipfs.remixproject.org',
+      port: 443,
+      protocol: 'https',
+      ipfsurl: 'https://ipfs.remixproject.org/ipfs/'
+    }
+    this.ipfsSources = [this.remixIPFS, this.globalIPFSConfig, this.ipfsconfig]
   }
 
   async getGitConfig () {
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/files/dgitProvider.js
```diff
@@ -38,6 +38,13 @@ class DGitProvider extends Plugin {
       protocol: 'https',
       ipfsurl: 'https://ipfs.io/ipfs/'
     }
+    this.remixIPFS = {
```
yann300, line 41: `ipfsconfig` already uses the same URL, is that wanted?

apps/remix-ide/src/app/files/dgitProvider.js
```diff
@@ -287,36 +294,58 @@ class DGitProvider extends Plugin {
     }
   };
 
+  async importIPFSFiles (config, cid, workspace) {
+    const ipfs = IpfsHttpClient(config)
+    let result = false
+    try {
+      const data = ipfs.get(cid, { timeout: 2000 })
```
yann300, line 301: `2000` for the timeout might not be enough in some cases where the content hasn't been queried since long

apps/remix-ide/src/app/files/dgitProvider.js
bunsenstraat: ipfsconfig is set by the dgit client. so it can be anything. so it's not a duplicate per se.

apps/remix-ide/src/app/files/dgitProvider.js
bunsenstraat: true. what would you advise?

yann300: if the UX allows the user to wait for the response, i'd say 10000 is fine..

yann300 (approved):

PR #1533, opened 2021-08-30
Fix loading compiler from unit testing

yann300:


Aniket-Engg (commented):
It looks working fine. Still what do you think about adding e2e tests about this?

Aniket-Engg (approved):

Issue #1431, opened 2021-08-02
JSON parsing error in Yul metadata field in Compilation Details


cameel:
Somewhat related to #122.

When I compile a Yul snippet and click `Compilation Details`, I see the following error message in the `METADATA` section:

```
Unable to display "$metadata": $JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

Other fields representing missing compilation artifacts are empty and this one should be too.

I tried it specifically with the code from #122 and Solidity 0.8.6 but I've seen this behavior with other code/versions too.

By the way, the work on making it possible to output all the same information in Yul mode is currently ongoing and in the near future it will be possible to store all the extra info in the IR and get the same output as for Solidity (https://github.com/ethereum/solidity/issues/11494).

LianaHus:
https://github.com/ethereum/remix-project/issues/122

PR #1557, opened 2021-09-03
do not show undefined property in details

LianaHus:
fix https://github.com/ethereum/remix-project/issues/1431


yann300 (approved):

ryestew (approved):

ryestew:
for testing use a .yul file like 
https://gist.github.com/chriseth/841869c4c5c5a792d8bb956d684b85d1 
Change compiler to yul in solidity compiler
go to compilation details and make sure that nothing and that it doesn't say Undefined property in the metadata field.

PR #1580, opened 2021-09-09
fix whitespaces and canactivate

bunsenstraat:
methods entered with whitespaces would be stored/used with whitespaces. 

to test: set methods or canactivate with spaces ( method1, method2 ) on local plugin, look at local storage for the localplugin item and see if it does not have whitespaces


libs/remix-ui/plugin-manager/src/types.d.ts
```diff
@@ -180,6 +180,7 @@ export interface FormStateProps {
   hash: string
   methods: any
```
joeizang, line 181: Almost everything looks good to me. Just it would be nice to avoid any as much as possible. methods seems to be string[] and canActivate a state dispatch. If we can be explicit with the types, it would be super! But everything else looks honky dory!

libs/remix-ui/plugin-manager/src/types.d.ts
```diff
@@ -180,6 +180,7 @@ export interface FormStateProps {
...
   location: string
+  canActivate: any
```
joeizang, line 183: Same as above

joeizang (commented):
If we can address the types that would be great for any future changes.

ioedeveloper (approved):

PR #1617, opened 2021-09-21
show hh logs for EVM revert

### LINKED ISSUE #1615: console.log is not printed for test if transaction fails with revert in SUT plugin

Aniket-Engg:
fixes #1615

### DIFF
--- a/libs/remix-tests/src/testRunner.ts
+++ b/libs/remix-tests/src/testRunner.ts
@@ -232,6 +232,7 @@ export function runTest (testName: string, testObject: any, contractDetails: Com
   testCallback(undefined, resp)
   async.eachOfLimit(runList, 1, function (func, index, next) {
     let sender: string | null = null
+    let hhLogs
     if (func.signature) {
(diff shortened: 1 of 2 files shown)


libs/remix-tests/src/testRunner.ts
```diff
@@ -377,6 +377,11 @@ export function runTest (testName: string, testObject: any, contractDetails: Com
           context: testName,
           web3
         }
+        if (err.message.includes('Transaction has been reverted by the EVM')) {
```
yann300, line 380: even if the transaction failed, you should be able to get a receipt.. it looks like this will break as soon as the error label change. but does seem we can do better..

yann300 (approved):

libs/remix-tests/src/testRunner.ts
Aniket-Engg: Yes, according to current web3 implementation, we don't get tx hash on error

PR #1625, opened 2021-09-24
Plugin manager improvements

bunsenstraat:
- It deactivates a plugin first before activating, that way forcing a plugin that could not be activated to retry connecting if possible. Resolves the 'Plugin is already rendered' issue.
- Allows replacing a plugin:
Screenshot 2021-09-24 at 13 33 44
Allowing replacing makes dev much easier, you can have an existing plugin replaced leaving intact for example context menu actions and so on. You can't do that if you need to give your plugin a new name all the time. This is more realistic. 
- add method to app manager to remove profile (should maybe go in plugin)

TODO:

- [x] e2e


apps/remix-ide-e2e/src/tests/pluginManager.spec.ts
```diff
@@ -127,7 +127,7 @@ module.exports = {
       // .waitForElementVisible('*[data-id="pluginManagerComponentDeactivateButtonremixIde"]', 60000)
   },
 
-  'Should display error message for creating already existing plugin': function (browser: NightwatchBrowser) {
+  /*   'Should display error message for creating already existing plugin': function (browser: NightwatchBrowser) {
```
LianaHus, line 130: please either remove or leave a comment with

PR #1634, opened 2021-09-28
Autoactivate

LianaHus:


apps/remix-ide/src/app/tabs/compile-tab.js
```diff
@@ -124,7 +124,11 @@ class CompileTab extends CompilerApiMixin(ViewPlugin) { // implements ICompilerA
       path: [],
       pattern: []
     })
-    this.currentFile = await this.call('fileManager', 'file')
+    try {
+      this.currentFile = await this.call('fileManager', 'file')
+    } catch (e) {
+      throw new Error(e)
```
yann300, line 130: do you really want to throw here?

LianaHus: this will be fixed by Filips PR. If we still have it we need to fix it. but I can also rm if you want

yann300: just that I m not sure what happen if there's a throw here, how does the remix-plugin code will react?
could that cause other plugin to not activate?
I think @bunsenstraat may be able to help here.

PR #1635, opened 2021-09-28
track compile event only once per autocompile session

LianaHus:


libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -281,7 +287,14 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
     compileIcon.current.setAttribute('title', 'idle')
     compileIcon.current.classList.remove('remixui_spinningIcon')
     compileIcon.current.classList.remove('remixui_bouncingIcon')
-    _paq.push(['trackEvent', 'compiler', 'compiled_with_version', _retrieveVersion()])
+    if (!state.autoCompile || (state.autoCompile && state.matomoAutocompileOnce)) {
```
yann300, line 290: this doesn't seem to push to matomo when the button 'compile' is clicked

PR #1638, opened 2021-09-29
fix event remove annotations on compile file

### LINKED ISSUE #1636: Error not cleared when compiling via right click -> compile
when using the compile button in the "solidity compiler" section errors are cleared after fixing them as expected:

https://user-images.githubusercontent.com/111600/135151019-13f17a5b-ec2c-43a8-bdc0-f94e035bfd2e.mov

but when using right-click -> compile the error persists even after fixing it:

https://user-images.githubusercontent.com/111600/135151162-9194e825-24d4-4b98-b1e3-f07137812723.mov

Josel95:
Fix #1636

The event `removeAnnotations` was moved just before the compile method as when compiling from right click menu it doesn't call `runCompiler` method.


yann300 (approved):

yann300:
thanks for this PR  @Josel95, can you come and say hello (https://gitter.im/ethereum/remix-dev), we can continue these contributions if you are interested

PR #1640, opened 2021-09-29
fix no current file error

bunsenstraat:


apps/remix-ide/src/app/tabs/compile-tab.js
```diff
@@ -124,7 +124,9 @@ class CompileTab extends CompilerApiMixin(ViewPlugin) { // implements ICompilerA
       path: [],
       pattern: []
     })
-    this.currentFile = await this.call('fileManager', 'file')
+    try {
+      this.currentFile = await this.call('fileManager', 'file')
+    } catch (error) {}
```
Aniket-Engg, line 129: So If I am right, It will never show error for any failure on `this.call('fileManager', 'file')`

bunsenstraat: well it does this on activation, a lot of the times there is no file selected when compiler is activated. it's not user action.

LianaHus: should we assign an empty string in catch? @bunsenstraat

bunsenstraat: no, it's already set by the init of the base class
https://github.com/ethereum/remix-project/blob/a11fe6b8d81cd68edff88b5ab32be6074077f721/apps/solidity-compiler/src/app/compiler-api.ts#L44

PR #1650, opened 2021-10-04
refactor: upgrade swarm support to use bee-js library

### LINKED ISSUE #1610: Open a discussion with the JS person from Swarm
We have some swarmgw commands and we had a publish to swarm button that we had taken down - we need to open a discussion with them about making a plugin for Remix and to explore other ways Swarm can be used with Remix.

agazso:
fix #1610 
Changed the deprecated `swarmgw` dependency to the `bee-js` library. Updated the code to upload to several hosts, including the gateway and a locally running node, similarly how the IPFS publisher works. Open issues:

- [x] there is a hardcoded single gateway node address, because the API has not been properly exposed and load-balanced on the gateway. This needs some discussion with the Swarm team for proposing a solution.
- [ ] the code still uses the deprecated `bzz-raw` protocol. The functionality works correctly and the generated hash matches, but this protocol is no longer available in Bee, instead there is only a simple `bzz` protocol. This needs some discussion with the Swarm team for proposing a solution.
- [x] there is a hard coded postage stamp id in the code. Ideally this could come from a setting (along with a configuration of the local node)
- [x] the `publishOnSwarm.js` file (in `apps/remix-ide/src/lib`) has not been updated yet. However it seems to be working correctly, so the question arises what that file is used for?
- [x] it seems that `eslint` with the current settings does not recognize `import type` statements correctly, so I had to disable it for one line, otherwise I get an error: `'UploadResult' is defined but never used` @yann300 
- [x] for now I kept the `swarmgw` dependency because removing it breaks the build with an error: `(node:36900) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'type' of undefined` @yann300


ryestew:
fix https://github.com/ethereum/remix-project/issues/1610

agazso:
> could you explain how to test it and verify it's actually published

The easiest way to test it is with using the public gateway only (which is the default case). In that case you just compile a smart contract and press the `Publish on Swarm` button. Then it will show a modal saying `Published Storage's Metadata` along with swarm references in the form of `bzz-raw://<reference>`.

Then you can use the `reference` and try to access it through the gateway API at `https://api.gateway.ethswarm.org/bytes/<reference>`. This is true for all swarm references by the way. Please note that when publishing the code also verifies the integrity of the file, because the solidity compiler also calculates the swarm hash of the file (== reference) and during the publishing it is checked if the hashes match. In case there is a mismatch you would see an error here, also if the upload fails for some reason.

There is also an option to upload to a private swarm node as well. In order to do that you have to provide the API address of a running [Bee node](https://docs.ethswarm.org/docs/introduction/terminology#bee) and a Postage Stamp ID. Then when publishing it also uploads the metadata to your local node, besides the public gateway.

Currently the code is forgiving in a sense that if there are multiple nodes to upload but at least one of them succeeds then it considers the publishing successful.

Hope that clears things up, let me know if you have more questions!

yann300:
I successfully uploaded 
```
{"compiler":{"version":"0.8.7+commit.e28d00a7"},"language":"Solidity","output":{"abi":[],"devdoc":{"kind":"dev","methods":{},"version":1},"userdoc":{"kind":"user","methods":{},"version":1}},"settings":{"compilationTarget":{"contracts/2_Owner.sol":"A"},"evmVersion":"london","libraries":{},"metadata":{"bytecodeHash":"ipfs"},"optimizer":{"enabled":false,"runs":200},"remappings":[]},"sources":{"contracts/2_Owner.sol":{"keccak256":"0x69059a7d8630f13673194f42fa799b1a0d8187cfed19a37368ace169c285b293","license":"GPL-3.0","urls":["bzz-raw://a1bcf445feaed2df40e7d2bc01a7f20eb9eaa68e8a6f9cde913d975acc4b3408","dweb:/ipfs/QmUKdGkFkfMkvyQR2fmmELuHicSdPb5QqSZimYcgzivBYr"]},"hardhat/console.sol":{"keccak256":"0x72b6a1d297cd3b033d7c2e4a7e7864934bb767db6453623f1c3082c6534547f4","license":"MIT","urls":["bzz-raw://a8cb8681076e765c214e0d51cac989325f6b98e315eaae06ee0cbd5va9f084763","dweb:/ipfs/QmNWGHi4zmjxQTYN3NMGnJd49jBT5dE4bxTdWEaDuJrC6N"]}},"version":1}
```

Is the source code uploaded too?
cause I can't reach https://api.gateway.ethswarm.org/bytes/a8cb8681076e765c214e0d51cac989325f6b98e315eaae06ee0cbd5va9f084763

yann300:
the metadata above is at https://api.gateway.ethswarm.org/bytes/de14ece87faf86a722512688ff672df010dc254784abc3a92089c73b130cf75e

yann300 (approved):

Issue #1333, opened 2021-06-25
Move vertical icon bar to Reactjs


ioedeveloper:
Vertical icon bar should have a scroll if needed

ryestew:
I don't see the arrows and I'm not sure why the top border is thicker than the bottom one.  And I think the problem is the thickness of the top border and not the bottom border...

Here it is in the dark theme:
<img width="52" alt="image" src="https://user-images.githubusercontent.com/6050001/137566033-f7e76ebd-de81-4db4-a556-e8f1a951ece1.png">

ryestew:
In the dark theme (at least on IPFS) the settings gear is alway in black -which does not seem correct.

ryestew:
The other thing is that when there are not too many plugins I see the thick border line under the plugin manager and it is not clear why the boder is there.

oh wait - I just tried to deactive all the plugins and I get this:
<img width="394" alt="image" src="https://user-images.githubusercontent.com/6050001/137566204-7515a09e-5275-4966-8bbf-934445fe3ef8.png">

ryestew:
I"ll try this later not on IPFS.

ryestew:
I reloaded and the plugins that I had deactivated are not in the icon panel - but the thick line is there
<img width="68" alt="image" src="https://user-images.githubusercontent.com/6050001/137566333-29a9c45e-3c9c-4edd-81ba-168d23dcf4ec.png">

ryestew:
The border lines probably should not be there unless the scroll is happening - because they don't serve any purpose except to show the scrolling area.

joeizang:
yeah I fixed those issues and would ask that you take a look.

vermouth22:
<img width="169" alt="Screen Shot 2022-01-19 at 1 54 09 PM" src="https://user-images.githubusercontent.com/35277634/150216568-4d396181-1b31-447f-ba1a-5e2bd23ca07d.png">
<img width="185" alt="Screen Shot 2022-01-19 at 1 55 09 PM" src="https://user-images.githubusercontent.com/35277634/150216573-9581ca6f-8811-4c40-98f2-f51d3e802aeb.png">
Rollover glow shadow (at Cairo Compiler) looks very subtle on a 13 inch Macbook Pro.

PR #1671, opened 2021-10-11
create vertical icons panel react lib

joeizang:
fix https://github.com/ethereum/remix-project/issues/1333
fix https://github.com/ethereum/remix-project/issues/133


ryestew:
I'm still having troubles getting the scroll to work properly.  Also - and this may just have to do wtih IPFS - when I try to load vyper - I see the icon in the icon panel but it does move to the active list in the plugin manager.

joeizang:
> I'm still having troubles getting the scroll to work properly. Also - and this may just have to do wtih IPFS - when I try to load vyper - I see the icon in the icon panel but it does move to the active list in the plugin manager.

I experienced both also but only on IPFS. I concluded that its an ipfs thing with plugins not working properly. This was an issue @bunsenstraat mentioned before and alluded to the fact that our plugins need to be updated somehow.

LianaHus:
sorry but I found a problem in the PR. The 'loading' badge on solidity icon is shifted. try to load a compiler to see the problem
@joeizang

Aniket-Engg:
image

I still see this on IPFS link.

LianaHus:
also I think we should discuss the position of plugin manager. doesn't look right in the middle of icons. I would suggest to

- either put it after the FE icon
- have a fine line after it to show that all native plugins are above it
- always show it at the end, before the settings icon(as we have now) even when we have the view with separators.

let's discuss this after the standup today

Aniket-Engg (commented):
There are multiple logs in browser with text `new status for solidity is :` and `from reducer of solidity :`

joeizang:
> There are multiple logs in browser with text `new status for solidity is :` and `from reducer of solidity :`

they have since been removed.

joeizang:
> also I think we should discuss the position of plugin manager. doesn't look right in the middle of icons. I would suggest to
> 
> * either put it after the FE icon
> * have a fine line after it to show that all native plugins are above it
> * always show it at the end, before the settings icon(as we have now) even when we have the view with separators.
> 
> Screenshot from 2021-11-09 10-58-20
> 
> let's discuss this after the standup today

This is done.

joeizang:
> sorry but I found a problem in the PR. The 'loading' badge on solidity icon is shifted. try to load a compiler to see the problem @joeizang

This also has been fixed.

apps/remix-ide-e2e/src/helpers/init.ts
```diff
@@ -5,7 +5,7 @@ require('dotenv').config()
 export default function (browser: NightwatchBrowser, callback: VoidFunction, url?: string, preloadPlugins = true): void {
   browser
     .url(url || 'http://127.0.0.1:8080')
-    .pause(5000)
+    .pause(10000)
```
ioedeveloper, line 8: `init.ts` is used in all test files. A 5sec increase in delay will increase circleci time significantly.

apps/remix-ide-e2e/src/tests/plugin_api.ts
```diff
@@ -112,6 +112,7 @@ const clickAndCheckLog = async (browser: NightwatchBrowser, buttonText: string,
   if (methodResult && typeof methodResult !== 'string') { methodResult = JSON.stringify(methodResult) }
   if (eventResult && typeof eventResult !== 'string') { eventResult = JSON.stringify(eventResult) }
   if (buttonText) {
+    console.log('this is buttonText in this test', buttonText)
```
ioedeveloper, line 115: Is this log needed? If not, please remove.

package.json
```diff
@@ -155,6 +155,7 @@
     "chokidar": "^2.1.8",
     "color-support": "^1.1.3",
     "commander": "^2.20.3",
+    "core-js": "^3.6.5",
```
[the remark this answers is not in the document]
ioedeveloper: I still see `core-js` listed as a dependency.

workspace.json
```diff
@@ -1035,72 +1101,36 @@
...
+    "@remix-ui-verticaliconspanel": {
+      "root": "libs/@remix-ui/verticaliconspanel",
+      "sourceRoot": "libs/@remix-ui/verticaliconspanel/src",
+      "projectType": "library",
+      "architect": {
+        "lint": {
+          "builder": "@nrwl/linter:eslint",
+          "options": {
+            "lintFilePatterns": [
+              "libs/@remix-ui/verticaliconspanel/**/*.{ts,tsx,js,jsx}"
```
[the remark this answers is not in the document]
ioedeveloper: I still see 2 entries for `@remix-ui-verticaliconspanel` in workspace.json

ioedeveloper (changes requested):

apps/remix-ide-e2e/src/tests/plugin_api.ts
joeizang: sorry left this while working here.

apps/remix-ide-e2e/src/helpers/init.ts
joeizang: I have made this value 6000, I hope this is an acceptable value.

workspace.json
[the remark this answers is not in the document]
joeizang: I have cleaned these up now.

libs/remix-ui/vertical-icons-panel/src/lib/components/RequiredSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+  itemContextAction: (e: any, name: string, documentation: string) => Promise<void>
+  addActive: (name: string) => void
+  removeActive: () => void
+  scrollableRef: MutableRefObject<any>
+}
+
+function RequiredSection ({ verticalIconsPlugin, itemContextAction, addActive, removeActive, scrollableRef }: RequiredSectionProps) {
+  return (
+    <Fragment>
+      <FilePanel
```
ioedeveloper, line 23: As an improvement, we can refactor the required section to reuse a single Icon component with changing props.

ioedeveloper (approved):

ryestew (approved):

PR #1703, opened 2021-10-26
do not throw error if storage is not available

### LINKED ISSUE #1690: unable to retrieve storage while debugging
It seems to work but the browser console says on failed and successful tests:
```
Error: unable to retrieve storage 0x972c1f8efdbd6566f3850ee80770b8d8e39f00b547f9752454c31485daeb6165 0xd8b934580fcE35a11B58C6D73aDeE468a2833fa8
```

Aniket-Engg:
fixes #1690


yann300 (approved):

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -117,7 +118,18 @@ export class CompilerImports extends Plugin {
     * @returns {Promise} - string content
     */
   async resolveAndSave (url, targetPath) {
-    if (url.indexOf('remix_tests.sol') !== -1) return remixTests.assertLibCode
+    if (url.indexOf('remix_tests.sol') !== -1 || url.indexOf('remix_accounts.sol') !== -1) {
+      let content
+      if (url === 'remix_tests.sol') content = assertLibCode
+      else if (url === 'remix_accounts.sol') {
+        this.testRunner = new UnitTestRunner()
```
yann300, line 125: getting a static content from remix-tests in that case is fine, but instantiating `new UnitTestRunner()` and re-executing `init` is not optimal _at all_. 
Moreover `init` include the instantiation of a simulator `this.createWeb3Provider(web3)`, which won't be used in that case, but will instanciate a full js-vm, that looks like wasting resources.

instead of that, would this work to have `await this.call('activate', 'solidityUnitTesting)'`? please check the syntax.

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -116,8 +116,7 @@ export class CompilerImports extends Plugin {
     * @param {String} targetPath - (optional) internal path where the content should be saved to
     * @returns {Promise} - string content
     */
-  async resolveAndSave (url, targetPath) {
-    if (url.indexOf('remix_tests.sol') !== -1) return remixTests.assertLibCode
+  async resolveAndSave (url, targetPath, testRunner = null) {
```
yann300, line 119: this can be rollbacked (and above also)

Aniket-Engg: ah yes. but `if (url.indexOf('remix_tests.sol') !== -1) return remixTests.assertLibCode` is not needed anymore i think as we create files in FE

apps/remix-ide/src/app/tabs/test-tab.js
```diff
@@ -549,7 +567,7 @@ module.exports = class TestTab extends ViewPlugin {
           this.updateFinalResult(error, result, testFilePath)
           callback(error)
         }, (url, cb) => {
-          return this.contentImport.resolveAndSave(url).then((result) => cb(null, result)).catch((error) => cb(error.message))
+          return this.contentImport.resolveAndSave(url, null, this.testRunner).then((result) => cb(null, result)).catch((error) => cb(error.message))
```
yann300, line 570: this need to be removed as well `this.testRunner`

yann300 (approved):

PR #1707, opened 2021-10-27
handle test files

Aniket-Engg:
Remix test files that are `remix_tests.sol` and `remix_accounts.sol` will be stored in file explorer inside `.deps/remix-tests` once someone runs unit tests in SUT plugin. To be exact they are created during compilation of a contract importing them.

Update:

Files will also be created when SUT plugin is activated.

### DIFF
--- a/apps/remix-ide/src/app/tabs/test-tab.js
+++ b/apps/remix-ide/src/app/tabs/test-tab.js
@@ -42,7 +42,7 @@ module.exports = class TestTab extends ViewPlugin {
     this.areTestsRunning = false
     this.defaultPath = 'tests'
     this.offsetToLineColumnConverter = offsetToLineColumnConverter
-    this.allFilesInvolved = []
+    this.allFilesInvolved = ['.deps/remix-tests/remix_tests.sol', '.deps/remix-tests/remix_accounts.sol']
     this.isDebugging = false
     this.currentErrors = []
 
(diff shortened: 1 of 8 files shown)


apps/remix-ide/src/app/tabs/test-tab.js
```diff
@@ -74,11 +74,21 @@ module.exports = class TestTab extends ViewPlugin {
...
+    return { assertLibCode, accountsLibCode: this.testRunner.accountsLibCode }
+  }
+
   async onActivation () {
     const isSolidityActive = await this.call('manager', 'isActive', 'solidity')
     if (!isSolidityActive) {
       await this.call('manager', 'activatePlugin', 'solidity')
     }
+    await this.testRunner.init()
+    const provider = await this.fileManager.getProviderOf()
```
yann300, line 87: can you use `currentFileProvider()` instead?

libs/remix-tests/src/compiler.ts
```diff
@@ -172,14 +172,7 @@ export function compileFileOrFiles (filename: string, isDirectory: boolean, opts
  */
 export function compileContractSources (sources: SrcIfc, compilerConfig: CompilerConfiguration, importFileCb: any, opts: any, cb): void {
   let compiler
-  const accounts: string[] = opts.accounts || []
   const filepath = opts.testFilePath || ''
```
yann300, line 175: you don't need anymore to inject the sources?

yann300 (approved):

libs/remix-tests/src/compiler.ts
Aniket-Engg: I think no, because they are received with content-import dynamically so we don't need to add them dynamically

apps/remix-ide/src/app/tabs/test-tab.js
Aniket-Engg: yes I can

PR #1727, opened 2021-11-03
Added Etherscan Verify Button

markusbug:
To be connected to the Etherscan Plugin


Th1983 (commented):
Run

apps/remix-ide/src/app/ui/universal-dapp-ui.js
```diff
@@ -156,6 +156,38 @@ UniversalDAppUI.prototype.renderInstanceFromABI = function (contractABI, address
     </div>
   `
 
+  // Verify smart contract on Etherscan UI
+  const verifySourceCodeOnEtherscan = yo`
```
LianaHus, line 160: we do not use yoyo anymore

PR #1742, opened 2021-11-10
Editor colors

LianaHus:
fix https://github.com/ethereum/remix-project/issues/1748


ryestew (commented):
DARK THEME
With old editor
image

With Monaco before this PR
image

With this PR
image

Open Zeppelin Wizard
image

LIGHT THEME
Old editor
image

Monaco before this PR
image

This PR
image

ryestew:
So if it is possible, it would be good if the keyword word **memory** is a bit easier to read in the dark theme without making it hard to read in the light theme.

ryestew:
In the light theme - the highlight color is good.
But in the light theme - the color palate seems like too much water got mixed in - they need more saturation - so they can stand out from the white-ish background.  Also there is no black color or very dark gray or dark blue.  

image

ryestew:
My suggestion is that the color for the variable names is a bit light and should be made slightly darker to
565B68
5B4343
363C49

See the light theme section below for the suggestions
remix-editor-colors

LianaHus:
If we change colors, you suggest this is changing all color schemes of the light theme.  info, success danger.. those are being used and not hardcoded colors. Let's merge this and create another issue to rethink the color scheme for the light theme.

ryestew (approved):

PR #1746, opened 2021-11-10
fixed the scroll when zoomed in

LianaHus:
fix https://github.com/ethereum/remix-project/issues/1743


ryestew (changes requested):
The scroll is not working for me.

ryestew:
When I horizontally scroll with the trackpad... it works!  When I use my mouse wheel, it does not work.  Also I thought that the horizontal scroll icon at the right of the screen only appears when there is something to horizontally scroll.  I see it all the time now.

ryestew:
It scrolls to the right with the mouse wheel but not to the left.

ryestew (approved):

PR #1756, opened 2021-11-11
add method

bunsenstraat:
this allows the FE to refresh after cloning in dgit


apps/remix-ide/src/app/files/fileManager.js
```diff
@@ -140,7 +140,7 @@ class FileManager extends Plugin {
   refresh () {
     const provider = this.fileProviderOf('/')
     // emit rootFolderChanged so that File Explorer reloads the file tree
-    provider.event.emit('rootFolderChanged', provider.workspace || '/')
```
yann300, line 143: we are listening on the `rootFolderChanged` of the provider here too: https://github.com/ethereum/remix-project/blob/9238ca9dfdc4649845f1902ba271b2cd2ff36324/apps/solidity-compiler/src/app/compiler-api.ts#L227

might worth keeping it

bunsenstraat: nobody else uses this method but dgit and dgit only works without remixd on

yann300 (approved):

PR #1783, opened 2021-11-23
Bump dev version

LianaHus:


yann300:
seems like not needed anymore

PR #1813, opened 2021-11-29
run monaco from assets

bunsenstraat:
**Fixes monaco editor in Remix Desktop**

The monaco editor sources from cdn were loading but caused errors. So loading them locally seemed the way to go.
Electron doesn't seem to like the minified sources from monaco so I had to load the full dev package in assets to get it to work. using the minified didn't do it.

so the path being copied from node_modules is node_modules/monaco-editor/dev using webpack
```
      new CopyWebpackPlugin({
        patterns: [
          { from: "../../../node_modules/monaco-editor/dev/vs/", to: "assets/js/monaco-editor/dev/vs" },
        ].filter(Boolean)
      })
```


libs/remix-ui/editor/src/lib/remix-ui-editor.tsx
```diff
@@ -288,7 +290,7 @@ export const EditorUI = (props: EditorUIProps) => {
 
   props.editorAPI.getFontSize = () => {
     if (!editorRef.current) return
-    return editorRef.current.getOption(42).fontSize
+    return editorRef.current.getOption(43).fontSize
```
ioedeveloper, line 293: What does 43 mean?

ioedeveloper (approved):

libs/remix-ui/editor/src/lib/remix-ui-editor.tsx
bunsenstraat: is the option for the fontsize. because I am using a later version of monaco the option nr changed

bunsenstraat: now we can use the enum from types... I will change it

PR #1822, opened 2021-12-01
fix desktop workspace issues

bunsenstraat:
- when switching to localhost it tries to merge the files with the previous state, effectively merging the file tree from the previous state with the new one, that seems to be not correct 
- when changing the state to localhost on desktop the file tree wasn't fetched.


ioedeveloper (commented):
Why is solidityImport_group1.spec.ts file deleted?

ioedeveloper (commented):
What change was made to react-tabs? I see it listed as changes.

PR #1831, opened 2021-12-06
terminl fix

tizah:


libs/remix-ui/terminal/src/lib/remix-ui-terminal.css
```diff
@@ -197,6 +197,7 @@ element.style {
     bottom           : 1em;
     border-width     : 4px;
     left             : 2em;
+    overflow-y: scroll;
```
LianaHus, line 200: Please keep the same spacing.

tizah: change implemented

yann300 (approved):

PR #1840, opened 2021-12-15
App-react

bunsenstraat:


yann300:
should remix app be created under `remix-ui` or rather directly under `apps/` ?

yann300:
can you confirm that there isn't changes in the app.component.js ? cause there's no diff...

ioedeveloper (approved):

yann300 (approved):

PR #1848, opened 2021-12-18
editor context view in react

yann300:
fix https://github.com/ethereum/remix-project/issues/1718

This move the context view in react. 
The context view is the component located between the editor and the terminal.
When the user click on a specific piece of code (function, variable,...) this panel gives more information about the current solidity "item". 
it allows to loop around all the references of the current item, to jump to the declaration.
In case of a "function" it display gas estimation value.

In order to use the panel, the code needs first to be compiled.


bunsenstraat (commented):
don't remove from nx.json
```
  },
  "targetDependencies": {
    "build": [
      {
        "target": "build",
        "projects": "dependencies"
      }
    ]
  }
```

libs/remix-ui/editor-context-view/src/lib/remix-ui-editor-context-view.tsx
```diff
@@ -0,0 +1,178 @@
...
+  /*
+    gotoLineDisableRef is used to temporarily disable the update of the view.
+    e.g when the user ask the component to "gotoLine" we don't want to rerender the component (but just to put the mouse on the desired line)
+  */
+  const referencesRef = useRef([])
+  const activeHighlightsRef = useRef([])
+  const currentNodeRef = useRef(null as nullableAstNode)
+  const gasEstimationRef = useRef({} as gasEstimationType)
+  const gotoLineDisableRef = useRef(false)
+  const [, setNode] = useState<Array<astNode>>([])
```
ioedeveloper, line 63: Why is `setNode` needed when the value is not being used?

yann300: yes I m trying to improve that now

yann300: should be better now

ioedeveloper (approved):

Issue #1290, opened 2021-06-16
Move Run and Deploy Tab to React.


ioedeveloper:
The run and deploy tab, currently built with yo-yo framework should be re-written in Reactjs.

### Additional Improvements
- [ ] https://github.com/ethereum/remix-project/issues/46

### Related issues
- [ ] Issue 1

PR #1850, opened 2021-12-20
Move Run and Deploy to Reactjs

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/1290


libs/remix-ui/run-tab/src/lib/actions/index.ts
```diff
@@ -0,0 +1,540 @@
...
+    const amount = plugin.blockchain.fromWei(tx.value, true, 'ether')
+    const content = confirmDialogContent(tx, network, amount, gasEstimation, plugin.blockchain.determineGasFees(tx), plugin.blockchain.determineGasPrice.bind(plugin.blockchain))
+
+    dispatch(displayNotification('Confirm transaction', content, 'Confirm', 'Cancel', () => {
+      plugin.blockchain.config.setUnpersistedProperty('doNotShowTransactionConfirmationAgain', plugin.REACT_API.confirmSettings)
+      // TODO: check if this is check is still valid given the refactor
+      if (!plugin.REACT_API.gasPriceStatus) {
+        cancelCb('Given transaction fee is not correct')
+      } else {
+        continueTxExecution({ maxFee: plugin.REACT_API.maxFee, maxPriorityFee: plugin.REACT_API.maxPriorityFee, baseFeePerGas: plugin.REACT_API.baseFeePerGas, gasPrice: plugin.REACT_API.gasPrice })
```
ioedeveloper, line 407: @yann300 Suggested including address as a field passed in to `continueTxExecution` to fix mainnet deployment error.

apps/remix-ide-e2e/src/tests/recorder.test.ts
```diff
@@ -73,11 +75,16 @@ module.exports = {
...
+        const modalOk = document.querySelector('[data-id="udappNotify-modal-footer-ok-react"]') as any
+
+        modalOk.click()
+      })
       .click('*[data-id="deployAndRunClearInstances"]') // clear udapp
       .click('*[data-id="terminalClearConsole"]') // clear terminal
-      .click('#runTabView .runtransaction')
-      .clickInstance(1)
+      .click('[data-id="runtransaction"]')
+      .clickInstance(2)
```
yann300, line 87: the input param for `clickInstance` doesn't seem to have changed. why do you need to put `2` instead of `1`

ioedeveloper: @yann300 The reason is that previously we increment the index passed into clickInstance command. Now we do not so the increased index is passed in now. https://github.com/ethereum/remix-project/pull/1850/files/58d1638b115c5102d126aabb4fb33af1722a7eba#diff-9d4fe3681c2fef5cd6e7c9fc9f52c230952964cad3b68cabbd938395c9bd1716L6

yann300: yes, the reason i ask that is that I see the use of  `clickInstance` hasn't been updated in some other places although I supposed it should have.

yann300 (approved):

PR #1856, opened 2021-12-23
Configplugin & registry 

bunsenstraat:
- 'config' is a plugin which can be called
- fixed async await issues when calling plugin
- use async in useeffect to get the variables
- replaced registry by TS singleton

### DIFF
--- a/apps/remix-ide/src/app/editor/contextView.js
+++ b/apps/remix-ide/src/app/editor/contextView.js
@@ -15,7 +15,7 @@ const css = require('./styles/contextView-styles')
 class ContextView {
   constructor (opts, localRegistry) {
     this._components = {}
-    this._components.registry = localRegistry || Registry.getInstance()
+    this._components.registry = Registry.getInstance()
     this.contextualListener = opts.contextualListener
     this.editor = opts.editor
     this._deps = {
--- a/apps/remix-ide/src/app/editor/contextView.js
+++ b/apps/remix-ide/src/app/editor/contextView.js
@@ -13,7 +13,7 @@ const css = require('./styles/contextView-styles')
    - rename declaration/references
 */
 class ContextView {
-  constructor (opts, localRegistry) {
+  constructor (opts) {
     this._components = {}
     this._components.registry = Registry.getInstance()
     this.contextualListener = opts.contextualListener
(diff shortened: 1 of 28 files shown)


apps/remix-ide/src/app/tabs/compile-tab.js
```diff
@@ -144,17 +144,12 @@ class CompileTab extends CompilerApiMixin(ViewPlugin) { // implements ICompilerA
...
   }
 
-  getAppParameter (name) {
-    // first look in the URL params then in the local storage
-    const params = this.queryParams.get()
-    const param = params[name] ? params[name] : this.config.get(name)
-    if (param === 'true') return true
-    if (param === 'false') return false
-    return param
+  async getAppParameter (name) {
```
yann300, line 147: are you certain that all the occurrences of `getAppParameter` use `await` now?

apps/remix-ide/src/app/editor/contextView.js
```diff
@@ -15,7 +15,7 @@ const css = require('./styles/contextView-styles')
 class ContextView {
   constructor (opts, localRegistry) {
     this._components = {}
-    this._components.registry = localRegistry || globalRegistry
+    this._components.registry = localRegistry || Registry.getInstance()
```
yann300, line 18: I guess you can now remove `localRegistry`  everywhere (this is basically the same as the global one)

apps/remix-ide/src/app/tabs/compile-tab.js
bunsenstraat: Can you find a place where it doesn't use it?

yann300: I think this is fine.
btw maybe the`getAppParameter` from  `compiler.ts` needs to use the new config plugin then, or is this not part of this PR and can be done afterwards ?

bunsenstraat: I saw that too, it confused me. The other one uses localstorage directly?

yann300: yes, the impl of the one in the plugin, is too use url param if any and fallback to the localstorage.
I think it's fine if the version from `compiler.ts` uses that version

bunsenstraat: That's better than I can change the api ts definition too.

yann300 (approved):

PR #1861, opened 2021-12-28
Fix vertical icons badge

joeizang:
Fix bug where dgit doesn't show any badge when others show. Example solidity compiler and static analysis show badges to mention a few.


yann300:
seems like there is no commit in this PR.

ryestew:
Is this PR addressing:
1. making the icons scroll when the space is filled up?  - if so - it took about 30 seconds or a minute  for the icons to become scroll-able.  I think I had that problem before.  
2.  making the chevrons clickable to do the scroll with a click - I don't think it is part of this PR (clicking the cevrons doesn't do anything)

joeizang:
> Is this PR addressing:
> 
> 1. making the icons scroll when the space is filled up?  - if so - it took about 30 seconds or a minute  for the icons to become scroll-able.  I think I had that problem before.
> 2. making the chevrons clickable to do the scroll with a click - I don't think it is part of this PR (clicking the cevrons doesn't do anything)

No, I think I will work as we agreed where if this is satisfactory for @bunsenstraat then I will move on to these issues. Especially point 2. I created a fix that might have fixed point 1. here [https://remix-alpha.ethereum.org/]

ryestew:
this works for me!

bunsenstraat (changes requested):
I tried it, but it seems if there are two plugins both should have badges.
for example try dgit and solhint. compile a file. @ryestew that's how you can test.
if you compile a file solhint should have badges and dgit should show badges too.

joeizang:
> I tried it, but it seems if there are two plugins both should have badges. for example try dgit and solhint. compile a file. @ryestew that's how you can test. if you compile a file solhint should have badges and dgit should show badges too.

I am looking at it now.

joeizang:
> I reviewed the PR. when I start to add plugins first there is a gap after Plugin manager, after a couple of icons, this gap is being filled. I would suggest fixing this and instead of adding space or gab after Plugin manager to have a fine line.

Ok...but that will have to be in a separate PR as this addresses a specific issue. I will combine that with the next fix I will be working on which will be right after I close this with @bunsenstraat. But thank you @LianaHus for taking the time to review it.

bunsenstraat (changes requested):
If you add plugin get badges, then deactivate the plugin and re-activate the plugin the badges are not there anymore.

PR #1876, opened 2022-01-04
do not merge files for localhost workspace

### LINKED ISSUE #1863: remixd creating empty default folders for shared folder
Screenshot 2021-12-31 at 6 29 58 PM

Directories `contracts, scripts, tests,.deps` are created with no files in there.

Aniket-Engg:
fixes #1863

### DIFF
--- a/libs/remix-ui/workspace/src/lib/reducers/workspace.ts
+++ b/libs/remix-ui/workspace/src/lib/reducers/workspace.ts
@@ -673,7 +673,7 @@ const fetchDirectoryContent = (state: BrowserState, payload: { fileTree, path: s
     }
   } else {
     if (payload.path === state.mode || payload.path === '/') {
-      let files = normalize(payload.fileTree, payload.path, payload.type)
+      const files = normalize(payload.fileTree, payload.path, payload.type)
       if (deletePath) delete files[deletePath]
       return { [state.mode]: files }
     } else {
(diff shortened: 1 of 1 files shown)


yann300 (approved):

libs/remix-ui/workspace/src/lib/reducers/workspace.ts
```diff
@@ -673,9 +673,7 @@ const fetchDirectoryContent = (state: BrowserState, payload: { fileTree, path: s
     }
   } else {
     if (payload.path === state.mode || payload.path === '/') {
-      let files = normalize(payload.fileTree, payload.path, payload.type)
-
-      files = _.merge(files, state[state.mode].files[state.mode])
+      const files = normalize(payload.fileTree, payload.path, payload.type)
```
ioedeveloper, line 676: We merge because this creates a bug in remix-desktop.

ioedeveloper (changes requested):

libs/remix-ui/workspace/src/lib/reducers/workspace.ts
bunsenstraat: @Aniket-Engg  if the publish version is the one you want  … I tested it on Desktop and @ioedeveloper is correct. if the tree is expanded and you add something in root the tree will collapse.

ioedeveloper (approved):

PR #1877, opened 2022-01-04
types in SUT component

Aniket-Engg:


libs/remix-ui/solidity-unit-testing/src/lib/logic/testTabLogic.ts
```diff
@@ -35,19 +36,20 @@ export class TestTabLogic {
...
         if (!path || !(/\S/.test(path))) return
         const fileProvider = this.fileManager.fileProviderOf(path.split('/')[0])
-        const res = await fileProvider.exists(path, (e: any, res: any) => { return res })
+        const res = await fileProvider.exists(path, (e: Error, res: boolean) => { return res })
         return res
     }
 
-    generateTestFile (errorCb: any) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    generateTestFile (errorCb:any) {
```
yann300, line 44: `errorCb` should be a `Function`

libs/remix-ui/solidity-unit-testing/src/lib/logic/testTabLogic.ts
```diff
@@ -72,7 +74,7 @@ export class TestTabLogic {
         let files = []
         try {
             if (await this.fileManager.exists(this.currentPath)) files = await this.fileManager.readdir(this.currentPath)
-        } catch (e: any) {
+        } catch (e: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
```
yann300, line 77: This can also be of type `Error`?

libs/remix-ui/solidity-unit-testing/src/lib/logic/testTabLogic.ts
Aniket-Engg: This shows linting error as: 
```
42:32  error  Don't use `Function` as a type. The `Function` type accepts any function-like value.
It provides no type safety when calling the function, which can be a common source of bugs.
It also accepts things like class declarations, which will throw at runtime as they will not be called with `new`.
If you are expecting the function to accept certain arguments, you should explicitly define the function shape  @typescript-eslint/ban-types
```

libs/remix-ui/solidity-unit-testing/src/lib/logic/testTabLogic.ts
Aniket-Engg: This shows build error as:
```
TS1196: Catch clause variable type annotation must be 'any' or 'unknown' if specified.
```

PR #1879, opened 2022-01-04
Gist handler in react

yann300:
fix https://github.com/ethereum/remix-project/issues/1869


libs/remix-core-plugin/src/lib/gist-handler.ts
```diff
@@ -0,0 +1,134 @@
...
+            id: 'gisthandler',
+            title: 'Gist load error',
+            message: data.message,
+            modalType: 'alert'
+          }
+          await this.call('modal', 'modal', modalContent)
+          return
+        }
+      } catch (e: any) {
+        const modalContent = {
```
bunsenstraat, line 100: it's easier to add type, in this case AppModal

yann300: yes but I can't import `AppModal` without breaking the build..

ioedeveloper (approved):

PR #1892, opened 2022-01-06
local compiler should be loaded without worker

Aniket-Engg:


Aniket-Engg:
With https://github.com/ethereum/solc-js/issues/581 , it is a bit unclear how a solc compiler can be loaded without worker, if possible.

PR #1901, opened 2022-01-09
Await async modals

bunsenstraat:


apps/remix-ide/src/app/plugins/remixd-handle.tsx
```diff
@@ -115,18 +117,10 @@ export class RemixdHandle extends WebsocketPlugin {
...
-          await this.canceled()
-        },
-        okLabel: 'Connect',
-        cancelLabel: 'Cancel',
-        hideFn: async () => {
-          if (this.state === State.new) await this.canceled()
-        }
       }
-      await this.call('modal', 'modal', mod)
+      else {
```
yann300, line 121: there isn't a `hide` handle?

bunsenstraat: There is no need. If return is undefined it is hide, if false it is cancel but in this case they both cause the plug-in to cancel

yann300 (approved):

PR #1913, opened 2022-01-11
encode arguments only on click

### LINKED ISSUE #1912: cannot encode arguments while deployment
Screenshot 2022-01-10 at 10 33 36 PM

Aniket-Engg:
fixes #1912

### DIFF
--- a/libs/remix-ui/clipboard/src/lib/copy-to-clipboard/copy-to-clipboard.tsx
+++ b/libs/remix-ui/clipboard/src/lib/copy-to-clipboard/copy-to-clipboard.tsx
@@ -12,10 +12,11 @@ interface ICopyToClipboard {
   direction?: Placement,
   className?: string,
   title?: string,
-  children?: JSX.Element
+  children?: JSX.Element,
+  onmousedown?: any
 }
 export const CopyToClipboard = (props: ICopyToClipboard) => {
-  let { content, tip = 'Copy', icon = 'fa-copy', direction = 'right', children, ...otherProps } = props
+  let { content, tip = 'Copy', icon = 'fa-copy', direction = 'right', children, onmousedown, ...otherProps } = props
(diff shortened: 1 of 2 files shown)


libs/remix-ui/clipboard/src/lib/copy-to-clipboard/copy-to-clipboard.tsx
```diff
@@ -6,30 +6,41 @@ import { Placement } from 'react-bootstrap/esm/Overlay'
 import './copy-to-clipboard.css'
 
 interface ICopyToClipboard {
-  content: any,
+  content?: any,
```
ioedeveloper, line 9: I think content should be of type `string` and not `any`.

libs/remix-ui/clipboard/src/lib/copy-to-clipboard/copy-to-clipboard.tsx
```diff
@@ -6,30 +6,41 @@ import { Placement } from 'react-bootstrap/esm/Overlay'
...
+  const handleClick = (e) => {
+    if (content) { // module `copy` keeps last copied thing in the memory, so don't show tooltip if nothing is copied, because nothing was added to memory
+      copyData()
     } else {
-      setMessage('Cannot copy empty content!')
+      content = getContent && getContent()
+      copyData()
     }
     e.preventDefault()
     return false
```
ioedeveloper, line 46: Can you remove think line? I don't think it is needed anymore.

ioedeveloper (approved):

libs/remix-ui/clipboard/src/lib/copy-to-clipboard/copy-to-clipboard.tsx
Aniket-Engg: done.

PR #1915, opened 2022-01-11
Permsreact

bunsenstraat:


apps/remix-ide/src/app/plugins/permission-handler-plugin.tsx
```diff
@@ -0,0 +1,126 @@
+import React from 'react' // eslint-disable-line
```
ioedeveloper, line 1: We can probably maintain `.ts` extension for plugin/service class and `.tsx` for react components. So it is easier to filter/differentiate.

ioedeveloper (approved):

Issue #1893, opened 2022-01-06
Remove Optimism plugin.


LianaHus:
We need to decide what we want to put on hometab instead.

PR #1942, opened 2022-01-13
dragbar z-index fixed, removed Optimism, added Cairo

LianaHus:
fix https://github.com/ethereum/remix-project/issues/1893


ioedeveloper:
When cairo compiler is activated from the hometab, the icon displayed in the vertical icons panel is just a blank circle and then when you switch to the plugin it then shows the icon properly. I just want to know if this is the expected behaviour?
Screen Shot 2022-01-14 at 11 02 31
Screen Shot 2022-01-14 at 11 05 54

ioedeveloper (approved):

PR #1984, opened 2022-01-20
solc version update for remix libs

Aniket-Engg:


yann300:
shouldn't we update solc version everywhere?

yann300:
@Aniket-Engg little bump for this PR

PR #1997, opened 2022-01-25
added Scam alert to home

LianaHus:


ryestew (commented):
The text is fine with me.
All the text is underlined -but is not a link - I mean only the "learn more" is a link - the rest is needlessly underlined (on hover).  Well this is just part of text-danger. Also it is all red and that seems like it is too much red - but we don't want people to miss this - so its OK.
image

And it looks like the Learn More is smaller than the rest of the text and it looks narrower.  But it looks like both are .8rem.

ryestew (approved):

PR #2048, opened 2022-02-05
Removed unnecessary (from React 17.0) React imports

glemiron:
- Removed unnecessary (from React 17.0) React imports
- Fragment replaced with shorter syntax <></>
- Removed eslint useless suppresses


glemiron:
I'll fix conflict a bit later

glemiron:
Here you can find some details why config was changed
https://stackoverflow.com/questions/64656055/react-refers-to-a-umd-global-but-the-current-file-is-a-module

glemiron:
I don't understand what can be wrong with failed test, but I'll try to fix it later

PR #2054, opened 2022-02-07
Import solidity standard json files via url param

kvhnuke:
This Pull request will allow users to import solidity standard files using the url param and set the workspace accordingly

Example: 
`http://localhost:8080/#optimize=false&runs=200&evmVersion=null&version=soljson-v0.6.12+commit.27d51765.js&url=https://raw.githubusercontent.com/EthVM/evm-source-verification/main/contracts/1/0x011e5846975c6463a8c6337eecf3cbf64e328884/input.json`

Please let me know if you'd like to add any additional features before merging

Ref #148


yann300:
Thanks for this PR!
could you make sure only the relevant source code is being updated? the diff is at the moment a bit complicated to apprehend.
Thanks very much.
adding the current release manager to the thread @ryestew

kvhnuke:
> Thanks for this PR! could you make sure only the relevant source code is being updated? the diff is at the moment a bit complicated to apprehend. Thanks very much. adding the current release manager to the thread @ryestew

will do, for some reason current `npm run lint` lead to all those changes

kvhnuke:
should be cleaner now, Thanks :)

kvhnuke:
@yann300 let me know if you need any other changes

libs/remix-ui/workspace/src/lib/actions/workspace.ts
```diff
@@ -75,29 +75,40 @@ export const loadWorkspacePreset = async (template: 'gist-template' | 'code-temp
...
+        path = 'contract-' + hash.replace('0x', '').substring(0, 10) + '.sol'
+        content = atob(params.code)
+        workspaceProvider.set(path, content)
+      }
+      if (params.url) {
+        const data = await plugin.call('contentImport', 'resolve', params.url)
+
+        path = data.cleanUrl
+        content = data.content
+        if (typeof content === 'object') {
```
ioedeveloper, line 94: `null` will pass this condition since it is also a typeof `'object'`.

libs/remix-ui/workspace/src/lib/actions/workspace.ts
```diff
@@ -75,29 +75,40 @@ export const loadWorkspacePreset = async (template: 'gist-template' | 'code-temp
...
+          const standardInput = content as JSONStandardInput
+          if (standardInput.language && standardInput.language === "Solidity" && standardInput.sources) {
+            const obj = standardInput.sources as any
```
ioedeveloper, line 97: `obj` is defined here, but i can't find where it is being used. You can remove this variable declaration if it is not needed.

ioedeveloper (changes requested):

libs/remix-ui/workspace/src/lib/actions/workspace.ts
kvhnuke: Thanks, fixed!

ioedeveloper (approved):

ioedeveloper:
@kvhnuke Can you rebase your PR? Thanks.

kvhnuke:
> 

done :)

Issue #1970, opened 2022-01-19
issue with url params in compiler


Aniket-Engg:
If value for `evmVersion` is given as something non-existing in URL params, on UI it sets `compiler default` as evm version but shows error on compilation
<img width="783" alt="Screenshot 2022-01-19 at 4 29 22 PM" src="https://user-images.githubusercontent.com/30843294/150117903-058b4457-2c95-4994-aa72-f515a6fdcd39.png">

PR #2062, opened 2022-02-09
fix providing wrong evm version

yann300:
fix https://github.com/ethereum/remix-project/issues/1970

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
+++ b/libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
@@ -17,10 +17,12 @@ export class CompileTabLogic {
   public language: string
   public compilerImport
   public event
+  public evmVersions: Array<string>
 
   constructor (public api: ICompilerApi, public contentImport) {
     this.event = new EventEmitter()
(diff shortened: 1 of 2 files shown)


libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
```diff
@@ -19,10 +19,12 @@ export class CompileTabLogic {
   public language: string
   public compilerImport
   public event
+  public evmVersions: Array<string>
 
   constructor (public api: ICompilerApi, public contentImport) {
     this.event = new EventEmitter()
     this.compiler = new Compiler((url, cb) => api.resolveContentAndSave(url).then((result) => cb(null, result)).catch((error) => cb(error.message)))
+    this.evmVersions = ['default', 'london', 'istanbul', 'petersburg', 'constantinople', 'byzantium', 'spuriousDragon', 'tangerineWhistle', 'homestead']
```
Aniket-Engg, line 27: `berlin` is missing in the list

yann300: that's fixed

Aniket-Engg (commented):
With IPFS env, it just doesn't set evm version from url params but just set `null` in url and `default` in the list

Aniket-Engg (approved):
Looks fine to me!

Issue #2024, opened 2022-01-31
Switching file tab clears code in solidity code on previous file tab: observed on Chrome


KennyDxsale:

ioedeveloper:
I can't reproduce this. Can you include a description of how to reproduce this?

KennyDxsale:
To reproduce the issue, I suggest you add in lots of scripts to the IDE.
After certain number of scripts added, the next ones will not save the the solidity script content, which disappears after switching tabs.

PR #2072, opened 2022-02-11
Show error message when localstorage is full and restore default file content.

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/2024


bunsenstraat (commented):
so you're saying the content is written although there's an error? and then it clips it or something?

ioedeveloper:
> so you're saying the content is written although there's an error? and then it clips it or something?

Yes, it overrides the existing content.

bunsenstraat (approved):

PR #2073, opened 2022-02-11
Run with mocha option in context menu

### LINKED ISSUE #2056: Support js unit tests for contracts

Aniket-Engg:
fixes #2056 

Related to https://github.com/yann300/remix-script-runner/pull/20


apps/remix-ide/src/remixAppManager.js
```diff
@@ -115,6 +115,7 @@ export class RemixAppManager extends PluginManager {
       const res = await fetch(this.pluginsDirectory)
       plugins = await res.json()
       plugins = plugins.filter((plugin) => {
+        if (plugin.name === 'scriptRunner') plugin.url = 'http://127.0.0.1:8081'
```
yann300, line 118: this should be removed

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -72,6 +72,16 @@ export class CompilerArtefacts extends Plugin {
     return contractsData
   }
 
+  getArtefactsByContractName (contractName) {
+    const contractsDataByFilename = this.getAllContractDatas()
```
yann300, line 76: it's maybe better to start checking the very last compilation result `getLastCompilationResult`.

apps/remix-ide/src/remixAppManager.js
Aniket-Engg: Yes, It will be

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
Aniket-Engg: current logic covers all compilations

yann300: yes that's right, but if multiple contracts with the same name were compiled, by using `getAllContractDatas` you are unsure to get the last compiled one... which is the one that it's likely users want to interact with.
that might not happen often but still...

ioedeveloper (approved):

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -72,6 +72,53 @@ export class CompilerArtefacts extends Plugin {
...
+        const fileList = await this.call('fileManager', 'fileList', path + '/artifacts')
+        const artefactsFilePaths = fileList.filter(filePath => {
+          const filenameArr = filePath.split('/')
+          const filename = filenameArr[filenameArr.length - 1]
+          if (filename === `${contractName}.json` || filename === `${contractName}_metadata.json`) return true
+        })
+        if (artefactsFilePaths && artefactsFilePaths.length) {
+          const content = await this.call('fileManager', 'readFile', artefactsFilePaths[1])
+          const artifacts = JSON.parse(content)
+          return { abi: artifacts.abi, bytecode: artifacts.data.bytecode.object }
```
yann300, line 88: what's the reason why you can't just return the compilation result? is this that the json files from the file system don't exactly match with the raw compilation result?

yann300: in that case, we should maybe think of replacing the <contract>_metadata.json by the actual compilation result

Aniket-Engg: The file named as <contractName>.json does not store bytecode. We can think of putting the compilation result in a file that can be done in a different PR

Aniket-Engg: Implemented by https://github.com/ethereum/remix-project/pull/2081

yann300 (approved):

PR #2081, opened 2022-02-18
save compilation result in file

Aniket-Engg:
- compiler input and version is also emitted with `compilationFinished` event
- build info is generated same as in hardhat
- Shows error while accessing the artefacts using contractName in case of same contract name in more than one files
- Allow accessing artefacts using fully qualified name which is `<filepath>:<contractName>`

For more info on compiler input and output: https://docs.soliditylang.org/en/v0.8.12/using-the-compiler.html#compiler-input-and-output-json-description

### DIFF
--- a/apps/solidity-compiler/src/app/compiler-api.ts
+++ b/apps/solidity-compiler/src/app/compiler-api.ts
@@ -261,11 +261,11 @@ export const CompilerApiMixin = (Base) => class extends Base {
 
     this.on('fileManager', 'fileClosed', this.data.eventHandlers.onFileClosed)
 
-    this.data.eventHandlers.onCompilationFinished = (success, data, source) => {
+    this.data.eventHandlers.onCompilationFinished = (success, data, source, input, version) => {
       this.compileErrors = data
       if (success) {
         // forwarding the event to the appManager infra
-        this.emit('compilationFinished', source.target, source, 'soljson', data)
+        this.emit('compilationFinished', source.target, source, 'soljson', data, input, version)
         if (data.errors && data.errors.length > 0) {
           this.statusChanged({
             key: data.errors.length,
--- a/libs/remix-core-plugin/src/lib/compiler-artefacts.ts
+++ b/libs/remix-core-plugin/src/lib/compiler-artefacts.ts
@@ -100,20 +100,33 @@ export class CompilerArtefacts extends Plugin {
 
   async getArtefactsByContractName (contractName) {
     const contractsDataByFilename = this.getAllContractDatas()
-    // let contractArtefacts
-    const  contractArtefacts = this.getAllContractArtefactsfromOutput(contractsDataByFilename, contractName)
-    let keys = Object.keys(contractArtefacts)
-    if (!keys.length) {
-      await this.getArtefactsFromFE ('contracts', contractName, contractArtefacts)
-      keys = Object.keys(contractArtefacts)
+    if (contractName.includes(':')) {
+      const nameArr = contractName.split(':')
+      const filename = nameArr[0]
+      const contract = nameArr[1]
+      if(Object.keys(contractsDataByFilename).includes(filename) && contractsDataByFilename[filename][contract]) 
+        return contractsDataByFilename[filename][contract]
+      else {
+        let allContractsData = {}
+        await this.getArtefactsFromFE ('contracts', contract, allContractsData)
+        if(allContractsData[contractName]) return allContractsData[contractName]
+        else throw new Error(`Could not find artifacts for ${contractName}. Compile contract to generate artifacts.`)
+      }
+    } else {
+      const contractArtefacts = this.getAllContractArtefactsfromOutput(contractsDataByFilename, contractName)
+      let keys = Object.keys(contractArtefacts)
+      if (!keys.length) {
+        await this.getArtefactsFromFE ('contracts', contractName, contractArtefacts)
+        keys = Object.keys(contractArtefacts)
+      }
+      if (keys.length === 1) return contractArtefacts[keys[0]]
+      else if (keys.length > 1) {
+        throw new Error(`There are multiple artifacts for contract "${contractName}", please use a fully qualified name.\n
+          Please replace ${contractName} for one of these options wherever you are trying to read its artifact: \n
+          ${keys.join()}\n
+          OR just compile the required contract again`)
+      } else throw new Error(`Could not find artifacts for ${contractName}. Compile contract to generate artifacts.`)
     }
-    if (keys.length === 1) return contractArtefacts[keys[0]]
-    else if (keys.length > 1) {
-      throw new Error(`There are multiple artifacts for contract "${contractName}", please use a fully qualified name.\n
-        Please replace ${contractName} for one of these options wherever you are trying to read its artifact: \n
-        ${keys.join()}\n
-        OR just compile the required contract again`)
-    } else throw new Error(`Could not find artifacts for ${contractName}. Compile contract to generate artifacts.`)
   }
 
   getCompilerAbstract (file) {
--- a/libs/remix-core-plugin/src/lib/compiler-artefacts.ts
+++ b/libs/remix-core-plugin/src/lib/compiler-artefacts.ts
@@ -80,12 +80,12 @@ export class CompilerArtefacts extends Plugin {
         const artefactsFilePaths = fileList.filter(filePath => {
           const filenameArr = filePath.split('/')
           const filename = filenameArr[filenameArr.length - 1]
-          if (filename === `${contractName}.json` || filename === `${contractName}_metadata.json`) return true
+          if (filename === `${contractName}_fullOP.json`) return true
         })
         if (artefactsFilePaths && artefactsFilePaths.length) {
-          const content = await this.call('fileManager', 'readFile', artefactsFilePaths[1])
+          const content = await this.call('fileManager', 'readFile', artefactsFilePaths[0])
           const artifacts = JSON.parse(content)
-          return { abi: artifacts.abi, bytecode: artifacts.data.bytecode.object }
+          return artifacts
         } else {
           for (const dirPath of dirList) {
             const result = await this.getArtefactsFromFE (dirPath, contractName)
@@ -106,7 +106,7 @@ export class CompilerArtefacts extends Plugin {
     const contractsData = Object.values(contractsDataByFilename)
     if (contractsData && contractsData.length) {
       const index = contractsData.findIndex((contractsObj) => Object.keys(contractsObj).includes(contractName))
-      if (index !== -1) return { abi: contractsData[index][contractName].abi, bytecode: contractsData[index][contractName].evm.bytecode.object }
+      if (index !== -1) return contractsData[index][contractName]
       else {
         const result = await this.getArtefactsFromFE ('contracts', contractName)
         if (result) return result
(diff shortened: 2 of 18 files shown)


bunsenstraat (approved):

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -80,12 +80,12 @@ export class CompilerArtefacts extends Plugin {
         const artefactsFilePaths = fileList.filter(filePath => {
           const filenameArr = filePath.split('/')
           const filename = filenameArr[filenameArr.length - 1]
-          if (filename === `${contractName}.json` || filename === `${contractName}_metadata.json`) return true
+          if (filename === `${contractName}_fullOP.json`) return true
```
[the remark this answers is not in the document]
Aniket-Engg: output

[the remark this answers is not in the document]
yann300: imo just `_fullOP` is fine

libs/remix-core-plugin/src/lib/compiler-metadata.ts
```diff
@@ -26,6 +26,10 @@ export class CompilerMetadata extends Plugin {
     return this.joinPath(path, this.innerPath, contractName + '_metadata.json')
   }
 
+  _OutputFileName (path, contractName) {
+    return this.joinPath(path, this.innerPath, contractName + '_fullOP.json')
```
yann300, line 30: hmm ok, this will be the compilation output per contract.
What do you think of having:
 - <contract_name>.json for the deployment configuration.
 - <contract_name>_metadata.json compilation metadata for this contract.
 - output.json the full compilation result regardless of the contract.

Aniket-Engg: if it will be `regardless of the contract.` how will we uniquely identify it?

yann300: the full compilation result contains the data for all the contracts, we can easily get data for a specific contract, and we store in a filesystem an artefact that is actually pretty common in the ecosystem.

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
yann300, line 83: this can be discarded right?

Aniket-Engg: yes, this is still to update once we are done with the file name and content stuff

libs/remix-core-plugin/src/lib/compiler-metadata.ts
```diff
@@ -26,12 +26,20 @@ export class CompilerMetadata extends Plugin {
...
+  }
+
   onActivation () {
     const self = this
     this.on('solidity', 'compilationFinished', async (file, source, languageVersion, data) => {
       if (!await this.call('settings', 'get', 'settings/generate-contract-metadata')) return
       const compiler = new CompilerAbstract(languageVersion, data, source)
       const path = self._extractPathOf(source.target)
+      const outputFileName = this._OutputFileName(path, source.target)
+      await this.call('fileManager', 'writeFile', outputFileName, JSON.stringify(data, null, '\t'))
```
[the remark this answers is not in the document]
Aniket-Engg: yes, right. will do that.

libs/remix-solidity/src/compiler/compiler.ts
```diff
@@ -133,7 +134,7 @@ export class Compiler {
    * @param source Source
    */
 
-  onCompilationFinished (data: CompilationResult, missingInputs?: string[], source?: SourceWithTarget): void {
+  onCompilationFinished (data: CompilationResult, missingInputs?: string[], source?: SourceWithTarget, input?, version?): void {
```
yann300, line 137: can this be typed?

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -72,51 +72,61 @@ export class CompilerArtefacts extends Plugin {
...
-  async getArtefactsFromFE (path, contractName) {
+  getAllContractArtefactsfromOutput (contractsOutput, contractName) {
+    const contractArtefacts = {}
+    for (const filename in contractsOutput) {
+      if(Object.keys(contractsOutput[filename]).includes(contractName)) contractArtefacts[filename + ':' + contractName] = contractsOutput[filename][contractName]
+    }
+    return contractArtefacts
+  }
+
+  async getArtefactsFromFE (path, contractName, contractArtefacts) {
```
yann300, line 83: could you set it as private if it's meant to be used only in that class?

yann300: same above

libs/remix-solidity/src/compiler/compiler.ts
Aniket-Engg: done

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
Aniket-Engg: done

libs/remix-solidity/src/compiler/compiler.ts
```diff
@@ -159,7 +161,7 @@ export class Compiler {
           source: source
         }
       }
-      this.event.trigger('compilationFinished', [true, data, source])
+      this.event.trigger('compilationFinished', [true, data, source, input, version])
```
yann300, line 164: see line 151,  we should add `intput` and `version ` there too.

apps/solidity-compiler/src/app/compiler-api.ts
```diff
@@ -249,11 +249,11 @@ export const CompilerApiMixin = (Base) => class extends Base {
     }
     this.on('fileManager', 'noFileSelected', this.data.eventHandlers.onNoFileSelected)
 
-    this.data.eventHandlers.onCompilationFinished = (success, data, source) => {
+    this.data.eventHandlers.onCompilationFinished = (success, data, source, input, version) => {
       this.compileErrors = data
       if (success) {
         // forwarding the event to the appManager infra
-        this.emit('compilationFinished', source.target, source, 'soljson', data)
+        this.emit('compilationFinished', source.target, source, 'soljson', data, input, version)
```
yann300, line 256: updating the signature of `compilationResult` is a big changes.. 
i.e we can keep it internal, but there's a tons of places where we are using the previous event signature.

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -72,51 +72,61 @@ export class CompilerArtefacts extends Plugin {
     return contractsData
   }
 
-  async getArtefactsFromFE (path, contractName) {
+  _getAllContractArtefactsfromOutput (contractsOutput, contractName) {
```
yann300, line 75: line 31 , probably that we should keep track of the `input` here too, so it can be used throughout the entire app.

apps/solidity-compiler/src/app/compiler-api.ts
Aniket-Engg: But this should be backward compatible I think

libs/remix-solidity/src/compiler/compiler.ts
Aniket-Engg: done

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
Aniket-Engg: this is done

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
Aniket-Engg: this is updated.

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -72,51 +72,61 @@ export class CompilerArtefacts extends Plugin {
...
+      const nameArr = contractName.split(':')
+      const filename = nameArr[0]
+      const contract = nameArr[1]
+      if(Object.keys(contractsDataByFilename).includes(filename) && contractsDataByFilename[filename][contract]) 
+        return contractsDataByFilename[filename][contract]
       else {
-        const result = await this.getArtefactsFromFE ('contracts', contractName)
-        if (result) return result
+        const allContractsData = {}
+        await this._getArtefactsFromFE ('contracts', contract, allContractsData)
```
yann300, line 111: having a return value isn't possible in that case? using the spread feature

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -72,51 +72,61 @@ export class CompilerArtefacts extends Plugin {
...
+        if(allContractsData[contractName]) return allContractsData[contractName]
```
yann300, line 112: in that case you don't check with the `filename`?

yann300: cause build-info folder maps with filename.
If you can't find the right contract at the given filename, it's better to throw imo.

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
```diff
@@ -72,51 +72,61 @@ export class CompilerArtefacts extends Plugin {
...
         else throw new Error(`Could not find artifacts for ${contractName}. Compile contract to generate artifacts.`)
       }
     } else {
-      const result = await this.getArtefactsFromFE ('contracts', contractName)
-      if (result) return result
-      else throw new Error(`Could not find artifacts for ${contractName}. Compile contract to generate artifacts.`)
-    }  
+      const contractArtefacts = this._getAllContractArtefactsfromOutput(contractsDataByFilename, contractName)
```
yann300, line 116: `_getAllContractArtefactsfromOutput` is used to get all the duplicates for a given contract? in that case it might be better to name it `_getDuplicateArtefacts` or something like that.

Aniket-Engg: Its not necessary that there will be duplicates

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
Aniket-Engg: next line is `else` throwing error. Let me know if that is something you meant

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
Aniket-Engg: Complete data is populated in `allContractsData`

yann300: yes I see, I am just wondering why you do like this instead of having `_getArtefactsFromFE` to return something.
but ok I guess that is fine like that, could you just rename the function `_populateContractsArtefactsFromFE`

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
yann300: I mean that in this scenario, the user has provided a filename, imo the function should be using that filename to return, and not only the `contractName`

yann300: nevermind

Aniket-Engg: In this case, `contractName ` is `<filename>:<contractName>`

libs/remix-core-plugin/src/lib/compiler-artefacts.ts
Aniket-Engg: done

yann300 (approved):

PR #2083, opened 2022-02-18
Fix retrieving events abi

yann300:
`eventsABI` is a mapping of contract name as key, and the values are mapping of event.

This function loops over all the contracts and all the events hashes to find the one given as parameter.
But there are duplicates..
The bug is that the current contract was ERC721 and event `Transfer` but the one returned by the function was the `Transfer` of an ERC20 (which is part of the compilation result too).

So the PR makes sure we are looking at the right contract.

### DIFF
--- a/libs/remix-lib/src/execution/eventsDecoder.ts
+++ b/libs/remix-lib/src/execution/eventsDecoder.ts
@@ -56,18 +56,19 @@ export class EventsDecoder {
     return eventsABI
   }
 
-  _event (hash, eventsABI) {
-    for (const k in eventsABI) {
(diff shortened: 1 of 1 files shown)


Aniket-Engg:
Can you share the steps/example to reproduce the issue without this PR?

yann300:
for reproducing:

 - compile https://github.com/Matlemad/PlaylistNFT/blob/main/PlaylistToken.sol
 - deploy playlist token contract
 - call the safemint function
 - an error should have thrown in the browser console and the tx result is never displayed in the remix console.

PR #2093, opened 2022-02-21
Clear error message when switching tabs

ioedeveloper:
Similar changes will also be needed for https://github.com/ethereum/remix-project/issues/2097


apps/remix-ide/src/app/tabs/compile-tab.js
```diff
@@ -82,6 +86,7 @@ class CompileTab extends CompilerApiMixin(ViewPlugin) { // implements ICompilerA
   }
 
   getFileManagerMode () {
+    this.emit()
```
yann300, line 89: what are you emitting?

libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
```diff
@@ -40,6 +42,14 @@ export const SolidityCompiler = (props: SolidityCompilerProps) => {
     })()
   }, [])
 
+  useEffect(() => {
+    if (badgeStatus[currentFile]) {
+      api.emit('statusChanged', badgeStatus[currentFile])
```
yann300, line 47: can't you use the `statusChanged` function too here

apps/remix-ide/src/app/tabs/compile-tab.js
ioedeveloper: This was a mistake. I will take it out.

libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
ioedeveloper: `statusChanged` function wouldn't work here. A plugin emitted event is needed.

yann300 (approved):

PR #2101, opened 2022-02-24
fix async method in remixdprovider

bunsenstraat:


apps/remix-ide/src/app/files/remixDProvider.js
```diff
@@ -98,20 +98,19 @@ module.exports = class RemixDProvider extends FileProvider {
...
-        if (error) console.log(error)
-        // display the last known content.
-        // TODO should perhaps better warn the user that the file is not synced.
-        return cb(null, this.filesContent[path])
-      })
+    try{
+      const file = await this._appManager.call('remixd', 'get', { path: unprefixedpath })
+      this.filesContent[path] = file.content
+      if (file.readonly) { this._readOnlyFiles[path] = 1 }
+      if(cb) return cb(null, file.content)
```
yann300, line 108: `if (cb)` you should probably not return but let line 109 to return

yann300 (approved):

Issue #1962, opened 2022-01-18
Hardhat Provider issues


Aniket-Engg:
- [x] Text in modal is coming in one line
<img width="630" alt="Screenshot 2022-01-18 at 6 28 59 PM" src="https://user-images.githubusercontent.com/30843294/149941824-bfeb8c7d-8b2a-45b2-9854-afa783680721.png">

- [x] `retainable.js:20 Unhandled promise rejection [TIMEOUT] Timeout for call modal from hardhat-provider` appears in browser console even after successful connection

- [x] tx hash appears twice in tx details
<img width="960" alt="Screenshot 2022-01-18 at 6 26 10 PM" src="https://user-images.githubusercontent.com/30843294/149941980-3b98cf29-329b-4f92-9ef9-a7b2bb25fd18.png">

- [x] `open in etherscan` appears without any link

<img width="954" alt="Screenshot 2022-01-18 at 6 26 24 PM" src="https://user-images.githubusercontent.com/30843294/149942046-847102ad-1c00-4243-8978-588d65284839.png">

- [x]  I think `hash` should be removed for calls on UI

<img width="1072" alt="Screenshot 2022-01-18 at 6 26 37 PM" src="https://user-images.githubusercontent.com/30843294/149942169-9368cf4d-cc8c-42e6-83a8-64a9c6e9772e.png">

- [x] Once I disconnect hardhat node locally, it doesn't go back to VM by default. Also, if I try to connect again without rerunning hardhat node. On clicking `OK` on modal, it again shows that screen with empty side panel

<img width="1271" alt="Screenshot 2022-01-18 at 6 27 51 PM" src="https://user-images.githubusercontent.com/30843294/149942438-21d87e1c-71c7-43d9-818a-00c45ec70c3d.png">

- [x] In above image, in remix terminal, txs are labelled as `[vm]` similar like https://github.com/ethereum/remix-project/issues/1937

- [ ] If `Hardhat provider` is selected in environment and workspace is changed, environment is set as `Javascript VM (London)`

PR #2119, opened 2022-03-01
Hardhat Provider Fixes

ioedeveloper:
Closes https://github.com/ethereum/remix-project/issues/1962


libs/remix-ui/terminal/src/lib/components/Table.tsx
```diff
@@ -158,21 +158,6 @@ const showTable = (opts, showTableHash) => {
             </td>
           </tr>
         ) : null}
-        {opts.hash ? (
```
yann300, line 161: this should be removed only if this isn't a call, you can use`opts.isCalll`

ioedeveloper: Do we need this displayed on the UI? We already display `transaction hash` in terminal output. It is a duplicate.

yann300 (approved):

PR #2138, opened 2022-03-04
update resolver

bunsenstraat:
update resolver not to return JSON
axios by default returns a JSON object if the data is JSON
preventing it from transforming makes it whatever it is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "@remixproject/plugin-ws": "^0.3.28",
     "ansi-gray": "^0.1.1",
     "async": "^2.6.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -161,7 +161,7 @@
     "@remixproject/plugin-ws": "^0.3.28",
     "ansi-gray": "^0.1.1",
     "async": "^2.6.2",
-    "axios": ">=0.21.1",
+    "axios": ">=0.26.0",
```
Aniket-Engg, line 164: May be this should also updated package-lock.json

Aniket-Engg (approved):

PR #2140, opened 2022-03-05
Auto exec

yann300:
This PR add menu item in the file explorer to set a script to run after each compilation (when ctrl+r is pressed).
 in order to run it:

 - add these files https://gist.github.com/yann300/04fb95555dbaa67ab1ea3a23959d4fdc
 - right click on the js file and click 'run after compilation'
 - select the 1_Storage.sol file, ctrl + s does a normal compilation (this is what we already have).
 - hit ctrl + r: it compiles, removes the current instances and run the script.
 - alternatively,:select the js file, hit ctrl+r. it runs the script


Hbutlercapone (approved):
Merge #2140

Hbutlercapone (approved):

Issue #1968, opened 2022-01-19
gist-handler issues


Aniket-Engg:
- [x] One first time load, on clicking on `GIST` from home tab, modal doesn't appear. These errors appear in browser console:
```
retainable.js:20 Unhandled promise rejection [TIMEOUT] Timeout for call load from home
retainable.js:20 Unhandled promise rejection [TIMEOUT] Timeout for call modal from gistHandler
```

- [x]  path for gist loaded file is shown with gist id in file explorer and with `gist-` prefix everywhere else. This creates ambiguity and makes it difficult to use file across the plugins. See screenshots:
<img width="894" alt="Screenshot 2022-01-19 at 2 18 15 PM" src="https://user-images.githubusercontent.com/30843294/150097077-fad11118-f2bd-46e6-a6ed-68855e68fc4a.png">
<img width="320" alt="Screenshot 2022-01-19 at 2 23 59 PM" src="https://user-images.githubusercontent.com/30843294/150097145-8adfef8f-a332-4fd9-91c9-f7076ff825c8.png">
<img width="298" alt="Screenshot 2022-01-19 at 2 25 50 PM" src="https://user-images.githubusercontent.com/30843294/150097170-162a60a9-f51a-461c-9f7d-18dde9e2f3b7.png">

PR #2143, opened 2022-03-07
readd gist prefix

yann300:
fix https://github.com/ethereum/remix-project/issues/1968

This PR handles the second item from the issue.


Aniket-Engg (commented):
As soon as I activate SUT plugin after a gist download, it shows error in browser console:
Screenshot 2022-03-08 at 6 16 16 PM

When I try to edit SUT plugin test directory, it show error in browser console and do not make any changes in test directory textbox value:

Screenshot 2022-03-08 at 6 16 33 PM

LianaHus:
the file name is not aligned with the icon. is this a new issue?

Aniket-Engg (commented):
Functionality wise this looks fine to me.

But for the gist folder name, it would be better if it can appear in one line

LianaHus (approved):

PR #2150, opened 2022-03-08
don't use look behinds

bunsenstraat:
safari doesn't support look behinds


yann300:
which issue is that fixing?

libs/remix-ui/search/src/lib/context/context.tsx
```diff
@@ -208,11 +208,11 @@ export const SearchProvider = ({
         const files = await getDirectory('/', plugin)
         const pathFilter: any = {}
         if (state.include){
-          const includeWithGlobalExpression = state.include.replaceAll(/(?<!\/)(\*\.)/g, '**/*.')
+          const includeWithGlobalExpression = state.include.replaceAll(/(\*\.)/g, '**/*.')
```
ioedeveloper, line 211: Wouldn't this change affect the correctness of the matching pattern?

bunsenstraat: I've changed it

Issue #2147, opened 2022-03-08
debugger update - is not saying that it did not find the file in Etherscan


ryestew:
on Ropsten with txn hash: 0x318d4ab00af73f6c8fa4a1d361c0992a853f45143a3f7ad1ad9d4a8937215377 of a Etherscan verfied contract, I see that it is not in sourcify from the toaster.  No toaster to say whether it is in etherscan or not.  But I do get this toaster
<img width="480" alt="image" src="https://user-images.githubusercontent.com/6050001/157154691-df5c3bb9-4740-4653-ab52-169afd2763bc.png">

With this toaster - I think the source code has been fetched.  But in the debugger when I pull the slider around nothing happens.  When I use the stepper I get this warning:

<img width="353" alt="image" src="https://user-images.githubusercontent.com/6050001/157155121-d1213b13-2651-4537-ad1b-2037ea666018.png">

Probs:
* did it retrieve it or not from etherscan - the toasters don't say
* this txn is in etherscan and is a verfied contract: https://ropsten.etherscan.io/address/0x7B91bCBb35450812689F9569D0Db9d40582392B9#code
* the slider didn't bring up the error
* the error came up with the stepper - but then why did the toaster come up saying that the contrat is being recompililed?

PR #2156, opened 2022-03-09
Fix solidity version while debugging

yann300:
fix https://github.com/ethereum/remix-project/issues/2147

 - add a warning when the debugger is trying to jump to a breakpoint.
 - bugfix: actual breakpoing was put in the line below the clicked line.

### DIFF
--- a/libs/remix-debug/src/debugger/debugger.ts
+++ b/libs/remix-debug/src/debugger/debugger.ts
@@ -41,11 +41,6 @@ export class Debugger {
       }
     })
 
-    this.breakPointManager.event.register('managersChanged', () => {
-      const { traceManager, callTree, solidityProxy } = this.debugger
-      this.breakPointManager.setManagers({ traceManager, callTree, solidityProxy })
-    })
-
     this.breakPointManager.event.register('breakpointStep', (step) => {
       this.event.trigger('breakpointStep', [])
       this.step_manager.jumpTo(step)
--- a/libs/remix-debug/src/debugger/debugger.ts
+++ b/libs/remix-debug/src/debugger/debugger.ts
@@ -47,7 +47,16 @@ export class Debugger {
     })
 
     this.breakPointManager.event.register('breakpointStep', (step) => {
+      this.event.trigger('breakpointStep', [])
       this.step_manager.jumpTo(step)
+    })
+
+    this.breakPointManager.event.register('noBreakpointHit', (step) => {
+      this.event.trigger('noBreakpointHit', [])
+    })
+
+    this.breakPointManager.event.register('locatingBreakpoint', () => {
+      this.event.trigger('locatingBreakpoint', [])
     })
 
     this.debugger.setBreakpointManager(this.breakPointManager)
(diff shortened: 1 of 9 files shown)


libs/remix-debug/test/debugger.ts
```diff
@@ -277,12 +277,7 @@ function testDebugging (debugManager) {
...
-    breakPointManager.event.register('managersChanged', () => {
-      const {traceManager, callTree, solidityProxy} = debugManager
-      breakPointManager.setManagers({traceManager, callTree, solidityProxy})
-    })
-
-    breakPointManager.add({fileName: 'test.sol', row: 38})
+    breakPointManager.add({fileName: 'test.sol', row: 39})
 
     breakPointManager.event.register('breakpointHit', function (sourceLocation, step) {
       console.log('breakpointHit')
```
ioedeveloper, line 283: It is safe to remove this line.

libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
```diff
@@ -87,6 +87,18 @@ export const DebuggerUI = (props: DebuggerUIProps) => {
       })
     })
 
+    debuggerInstance.event.register('locatingBreakpoint', async (isActive) => {
+      setState(prevState => {
+        return { ...prevState, sourceLocationStatus: 'Locating breakpoint, this might take a while...' }
```
ioedeveloper, line 92: I'm not sure if this setState will work as expected in a call back. You can define a new `useState` variable for `sourceLocationStatus` and update it directly.

libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
```diff
@@ -87,6 +87,18 @@ export const DebuggerUI = (props: DebuggerUIProps) => {
...
+      })
+    })
+
+    debuggerInstance.event.register('noBreakpointHit', async (isActive) => {
+      setState(prevState => {
+        return { ...prevState, sourceLocationStatus: '' }
```
ioedeveloper, line 98: See comment above.

libs/remix-debug/src/debugger/debugger.ts
```diff
@@ -41,15 +41,19 @@ export class Debugger {
       }
     })
 
-    this.breakPointManager.event.register('managersChanged', () => {
-      const { traceManager, callTree, solidityProxy } = this.debugger
-      this.breakPointManager.setManagers({ traceManager, callTree, solidityProxy })
-    })
-
     this.breakPointManager.event.register('breakpointStep', (step) => {
+      this.event.trigger('breakpointStep', [])
```
ioedeveloper, line 45: Is this a recursive call? Because i don't see anywhere else that `breakpointStep` event is registered. If not, then i'm not sure how useful this line is.

libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
yann300: This works, I used the same method as line 106. let me know still if there's a better way to handle this warning message

PR #2163, opened 2022-03-10
Set annotations from the compiler api

yann300:
fix annotation:

 - compile with an error, annotations are shown.
 - fix the error and recompile, the annotation remain.

### DIFF
--- a/apps/solidity-compiler/src/app/compiler-api.ts
+++ b/apps/solidity-compiler/src/app/compiler-api.ts
@@ -293,7 +293,7 @@ export const CompilerApiMixin = (Base) => class extends Base {
       if (this.onCompilationFinished) this.onCompilationFinished(this.compilationDetails)
       // set annotations
       if (data.errors) {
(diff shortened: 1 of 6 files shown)


ioedeveloper:
@yann300 Shouldn't `Renderer.tsx` be updated also? It performs similar task.

yann300:
I refactored out the `getPositionDetails`. So the issue here was that:
 - `onCompilationFinished` is triggerred only one time.
 - but the react ui is updating the state using several `setState` , which leads the Renderer.tsx to rerender several time.
 - which is maybe ok (cause the renderer is self sufficient).
 - but the renderer calls then the editor api to set the annotations several times (due to the several render)
 - It would have been possible to double check in the api id the annotation is already set, but that creates another issue....
 - so that PR move the annotations to a place, where we are sure it's being called 1 time for a single compilation result.

ioedeveloper (approved):

Issue #2158, opened 2022-03-10
Debugger step over back


ryestew:
Take this mainnet contract to debug:
0x5191d0053c670cd53eafff4c92dcc095823a7dae3238344be72daea2c4b935fa

Click the **step over forward** button many times until the slider jumps to the middle.  Then click the **step over back** button.  Nothing happens - it doesn't go back.

PR #2166, opened 2022-03-10
Fix debugger step

yann300:
fix https://github.com/ethereum/remix-project/issues/2158

 - don't update heavy data at every single step changed.
 - fix loading state and locals with timeout.
 - `getValidSourceLocationFromVMTraceIndex` don't call it so often.

### DIFF
--- a/libs/remix-core-plugin/src/lib/helpers/fetch-etherscan.ts
+++ b/libs/remix-core-plugin/src/lib/helpers/fetch-etherscan.ts
@@ -10,7 +10,7 @@ export const fetchContractFromEtherscan = async (plugin, network, contractAddres
     // etherscan api doc https://docs.etherscan.io/api-endpoints/contracts
     if (data.message === 'OK' && data.status === "1") {
       if (data.result.length) {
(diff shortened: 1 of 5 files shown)


ryestew (commented):
This fix is not working for this problem (below).  But this problem might be specific to this contract that is importing an unverified contract.

Try debugging this txn on mainnet: 0x5191d0053c670cd53eafff4c92dcc095823a7dae3238344be72daea2c4b935fa

Click the step over forward buttong 45 times until you get to step 0128 JUMP - if you click step over forward again - nothing happens and if you click any other type of step button, nothing happens.  You can use the slider.  But this is not an issue of clicking too fast ( I clicked slowly enough to see the opcodes change and I didn't click again until they changed.

Clearing this problem is easy - just by dragging the slider to the beginning. 

If  I use the step into button 45x to get to step 0128 JUMP - I can go to the next step and I can go backwards and when I get to the next step I can then use the step over forward button.

yann300:
@ryestew can we still merge it? I feel this can improve UX still

ryestew (approved):

PR #2187, opened 2022-03-15
truffle integration

Aniket-Engg:


libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
```diff
@@ -133,6 +139,12 @@ export class CompileTabLogic {
             this.api.logToTerminal({ type: 'error', value: error })
           })
         }
+      } else if (externalCompType === 'truffle') {
```
yann300, line 142: `this.api.getFileManagerMode() === 'localhost'` should be in the check too.

yann300:
could you rollback the changes to package-lock.json

libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
Aniket-Engg: ah yes

Aniket-Engg: updated

Aniket-Engg:
> could you rollback the changes to package-lock.json

done

yann300 (approved):

Issue #2177, opened 2022-03-14
Recursion in copy past


LianaHus:
my steps to reproduce.
copy `contracts` folder
past in `contracts` folder
past in `artifacts` folder of newly pasted one
recursive pasting happens...

remix-alpha and remix-beta affected but not remix@ethereum.org.

LianaHus:
I still see it on live.

yann300:
@LianaHus could you check in again please?

Issue #2153, opened 2022-03-08
Copying and pasting a file into a sub directory, then not being able to close it


cjxe:
### Steps to reproduce
1- create a new workspace
2- create sub directory under `contracts/` named `test`
2- open (double click on)  `2_Owner.sol`
3- right click on `./contracts/2_Owner.sol` > Copy
4- right click on sub dir `test/` > Paste
5- open `./contracts/test/Copy_2_Owner.sol`
6- rename `./contracts/test/Copy_2_Owner.sol` -> `./contracts/test/2_Owner.sol`
7- delete `./contracts/2_Owner.sol`
8- click on `test/2_Owner.sol` at the top bar
9- close `2_Owner.sol` 
10- on the top bar `test/2_Owner.sol` renames automatically to `2_Owner.sol` and can't be closed

### Steps to solve
1- Refresh the  #page

**Browser:** Version 1.36.109 Chromium: 99.0.4844.51 (Official Build) (64-bit)

LianaHus:
@cjxe hi. thanks for reporting this.i also discovered the bug couple of days ago but it was on remix-alpha and remix-beta. remix.ethereum.org was fine.
Therefore I should ask on which version it was reported. 
- [ ] remix.ethereum org
- [ ] remix-alpha.ethereum org
- [ ] remix-beta.ethereum org

cjxe:
The bug occurs in `v0.22.0-dev` and on the following URLs.

- [ ] remix.ethereum org
- [x] remix-alpha.ethereum org
- [x] remix-beta.ethereum org

PR #2193, opened 2022-03-16
Fix copy

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/2153
Fixes https://github.com/ethereum/remix-project/issues/2177

### DIFF
--- a/apps/remix-ide/src/app/files/remixDProvider.js
+++ b/apps/remix-ide/src/app/files/remixDProvider.js
@@ -1,5 +1,6 @@
 'use strict'
 const FileProvider = require('./fileProvider')
+const pathModule = require('path')
 
 module.exports = class RemixDProvider extends FileProvider {
   constructor (appManager) {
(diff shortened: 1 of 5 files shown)


apps/remix-ide/src/app/files/remixDProvider.js
```diff
@@ -207,4 +208,12 @@ module.exports = class RemixDProvider extends FileProvider {
     if (!this._isReady) throw new Error('provider not ready')
     return await this._appManager.call('remixd', 'isFile', { path: unprefixedpath })
   }
+
+  isSubDirectory (parent, child) {
```
yann300, line 212: don't you need to add `isSubDirectory` to all the provider?

yann300: you should perhaps then add  `isSubDirectory` to `FileProvider` instead of `RemixDProvider`

yann300 (approved):

PR #2197, opened 2022-03-17
Search3

bunsenstraat:
- use enter to search/also on include/exclude, do not update on typing
- search can be stopped with a stop button
- async plugin calls to files are canceled when search params changes
- search is throttled so not all files get hit one after another, so there is a delay between files
- file names that are searched are being displayed

### DIFF
--- a/apps/remix-ide-e2e/src/tests/search.test.ts
+++ b/apps/remix-ide-e2e/src/tests/search.test.ts
@@ -4,7 +4,6 @@ import { NightwatchBrowser } from 'nightwatch'
 import init from '../helpers/init'
 
 module.exports = {
-    '@disabled': true,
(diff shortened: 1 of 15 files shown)


yann300:
@bunsenstraat could you shortly describe how you did approach the changes?

bunsenstraat:
@yann300  fixed
- clear highlights when search changes
- bug : when cloning it didn't index the files

LianaHus:
could you please add some spacing at the bottom of search result

bunsenstraat:
> could you please add some spacing at the bottom of search result

Where exactly and how much?

LianaHus:
`pb-4` to search_plugin_search_tab :)

yann300 (approved):

PR #2202, opened 2022-03-17
update process

yann300:


LianaHus (changes requested):
approved but plz merge after release

Issue #1767, opened 2021-11-17
Solidity compiler: add all missing flags for compiler


LianaHus:
there should be a way to provide json file and also check from UI

LianaHus:
![Screenshot from 2022-03-21 11-03-41](https://user-images.githubusercontent.com/9685356/159240049-4dff0f7b-fdf4-4d8c-ad22-7b7965efa4af.png)

PR #2249, opened 2022-03-29
Compiler config

LianaHus:
fix https://github.com/ethereum/remix-project/issues/1767


libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -8,6 +8,8 @@ import { compilerReducer, compilerInitialState } from './reducers/compiler'
 import { resetEditorMode, listenToEvents } from './actions/compiler'
 import { OverlayTrigger, Tooltip } from 'react-bootstrap' // eslint-disable-line
 import { getValidLanguage } from '@remix-project/remix-solidity'
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
+import * as json_config from '../../../../../apps/remix-ide/contracts/solidity_compiler_config.json'
```
yann300, line 12: could you put this file in the `solidity-compiler` react plugin or in the remix-solidity module?

apps/solidity-compiler/src/app/compiler.ts
```diff
@@ -48,7 +44,9 @@ export class CompilerClientApi extends CompilerApiMixin(PluginClient) implements
       optimize: localStorage.getItem('optimize') === 'true',
       version: localStorage.getItem('version') || defaultCompilerParameters.version,
       evmVersion: localStorage.getItem('evmVersion') || defaultCompilerParameters.evmVersion, // default
-      language: localStorage.getItem('language') || defaultCompilerParameters.language
+      language: localStorage.getItem('language') || defaultCompilerParameters.language,
+      useFileConfiguration: localStorage.getItem('useFileConfiguration') == 'true',
```
ioedeveloper, line 48: Why `==` instead of `===` like in line 44?

libs/remix-lib/src/types/ICompilerApi.ts
```diff
@@ -24,7 +24,7 @@ export interface ICompilerApi {
 
     onCurrentFileChanged: (fileName: string) => void
     // onResetResults: () => void,
-    onSetWorkspace: (workspace: any) => void
+    onSetWorkspace: (isLocalhost: boolean, workspaceName: string) => void
```
ioedeveloper, line 27: The old argument contains `isLocalhost` and `workspaceName` as properties already. Is there any reason why you seperated them into individual arguments?

libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
```diff
@@ -46,18 +48,31 @@ export class CompileTabLogic {
     this.api.setCompilerParameters({ evmVersion: this.evmVersion })
     this.compiler.set('evmVersion', this.evmVersion)
 
+    this.useFileConfiguration = this.api.getCompilerParameters().useFileConfiguration
+    console.log("this.configFilePath ", this.configFilePath)
```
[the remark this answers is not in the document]
LianaHus: it's in wip :)  sure

libs/remix-lib/src/types/ICompilerApi.ts
LianaHus: I don't remember details but there was. maybe because the name localhost doesn't mean it is. i should look closer

LianaHus: besides i always pass as little as is necessary. :))

bunsenstraat:
- when you click set new config file, you can't go out of it by escape or something, the input field always stays open
- option to use a file is not remembered on reload
- file name can be clickable so that it opens
- it seems to try to compile the json file every time it's saved, even though auto compile is off
- we could have a right click in FE to set the config file, if you've created one before you can't select it now in the advanced configuration. maybe we can store the config files a dedicated dir? just an idea

LianaHus:
> * when you click set new config file, you can't go out of it by escape or something, the input field always stays open
+
> * option to use a file is not remembered on reload
will check
> * file name can be clickable so that it opens
+
> * it seems to try to compile the json file every time it's saved, even though auto compile is off
it shouldn't compile json file at all. if it is a case please add a separate issue
> * we could have a right click in FE to set the config file, if you've created one before you can't select it now in the advanced configuration. maybe we can store the config files a dedicated dir? just an idea
right click sound good but this can be a separate PR

bunsenstraat (commented):
lock file should probably not be included?

Aniket-Engg (commented):
1. This UI looks a bit inconsistent. Not sure if we should move `Include nightly builds` parellel to auto compile and Hide warnings

Screenshot 2022-05-23 at 7 02 58 PM


2. We can disassociate the config file creation with compiler plugin activate and associate it with the `Use configuration file` radio button click. In current scenario, there is a possibility that file stays in the workspace and used never.
3.  URL params are synced with the config provided through UI. Not sure if we should update the behaviour for configuration file.
4. Choice to provide configuration is preserved on new workspace creation. Considering other functionalities, we can reset it on workspace creation.
5. For `Set new config file` button, we can keep the button parellel to file name and label it with `Change`
Screenshot 2022-05-23 at 7 15 01 PM


6. When `Use configuration file` radio button is select, other UI is faded but checkbox for `Enable Optimization` doesn't
7. If we remove some of the fields from output selection arrays, compilation shows this in browser console
```
SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at CompilerMetadata._setArtefacts (compiler-metadata.ts:90)
    at compiler-metadata.ts:42
```
and `Compilation details` modal still shows those fields with blank value

ryestew:
The set new config file button in the light theme has an dark outline - that looks like a faux 3d effect that doesn't fit with the rest of the IDE - and its not visible in the dark theme.
image

LianaHus:
> configFilePath: "/compiler_config.json"



> The set new config file button in the light theme has an dark outline - that looks like a faux 3d effect that doesn't fit with the rest of the IDE - and its not visible in the dark theme. image

this is how our theme made. it's a standard look

LianaHus:
"change" instead of set...

libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
```diff
@@ -52,12 +54,22 @@ export class CompileTabLogic {
...
     this.compiler.set('optimize', this.optimize)
   }
 
+  setUseFileConfiguration (useFileConfiguration: boolean) {
+    this.useFileConfiguration = useFileConfiguration
+    this.compiler.set('useFileConfiguration', useFileConfiguration)
+  }
+
+  setConfigFilePath (path) {
+    console.log("path in logic = ", path)
```
Aniket-Engg, line 69: this can be removed

Aniket-Engg (approved):

PR #2258, opened 2022-04-03
Update README Setup Instructions

Miladinho:
I recently cloned this repo to make a change, I noticed that the setup instructions were wrong (at least for my machine). While I installed NX CLI i found that there were existing commands inside the package.json scripts that ran the `nx build` and `nx serve` commands and that it is better to use those so I updated the README accordingly.


Miladinho:
@Aniket-Engg need a CI re-run on this as well i believe

README.md
```diff
@@ -52,8 +52,8 @@ git clone https://github.com/ethereum/remix-project.git
 cd remix-project
 npm install
 npm run build:libs // Build remix libs
-nx build
-nx serve
+npm run build
+npm run serve
```
LianaHus, line 56: nx build remix-ide
nx serve

should work

Miladinho: sure it might work but isn't it better to use the local binaries and run the command from the script instead of requiring to download an external global library? It just seems like if the script is there we might as well use it... seems like a better practice to me

Miladinho:
@yann300 curious to get your feedback on this

README.md
ioedeveloper, line 56: `npm run build` builds the whole project and limits the available command/flags that a user can leverage when using `nx`.
Example: `nx build remix-simulator` or `nx build --skip-nx-cache`

`npm run build` will only be useful if we want to build the whole project. I think we should still keep the old changes or provide a better description to distinguish the usecase of having `nx` installed global compared to using the local `nx` within the `node_modules`.

ryestew:
The instructions have since been updated.

PR #2262, opened 2022-04-05
UI for auto exec

yann300:
- solidity plugin: add keybindings tooltips.
 - new file creation: default content for sol, ts and js file.
 - new file creation contains the `dev-run-script` tag


libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -621,12 +621,32 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
               </a>
             </div>
           }
-          <button id="compileBtn" data-id="compilerContainerCompileBtn" className="btn btn-primary btn-block remixui_disabled mt-3" title="Compile" onClick={compile} disabled={disableCompileButton}>
+          <OverlayTrigger overlay={
+              <Tooltip id="overlay-tooltip">
+                <div className="text-left">
+                <div>Ctr+s for compiling</div>
```
Aniket-Engg, line 627: typo

Aniket-Engg:
Feedback:

- remove `console.log(JSON.stringify(result, null, ' ')))` from deploy scripts as it prints a big output
- Rather than saying it on the fly execution, will be better to name it `CompileAndRun` feature which runs a script after compilation
-  if a script is not set for a contract and user presses CTRL+SHIFT+S, it should show a toaster like `You have not set a script for CompileAndRun. Set it with @dev-run-script natspec tag`

apps/remix-ide/src/app/tabs/compile-and-run.ts
```diff
@@ -50,7 +62,14 @@ export class IntelligentScriptExecutor extends Plugin {
       if (this.targetFileName === contract.file && contract.object && contract.object.devdoc['custom:dev-run-script']) {
         this.targetFileName = null
         const file = contract.object.devdoc['custom:dev-run-script']
-        if (file) this.runScript(file, true)
+        if (file) {
+          this.runScript(file, true)
+          _paq.push(['trackEvent', 'ScriptExecutor', 'run_script_after_compile'])
+        } else {
+          this.call('notification', 'toast', 'You have not set a script to run. Set it with @dev-run-script natspac tag.')
```
Aniket-Engg, line 69: typo. it is `NatSpec`

Aniket-Engg: tag name should be `@custom:dev-run-script`

apps/remix-ide/src/app/tabs/compile-and-run.ts
```diff
@@ -50,7 +62,14 @@ export class IntelligentScriptExecutor extends Plugin {
...
+        }
+      } else {
+        this.call('notification', 'toast', 'You have not set a script to run. Set it with @dev-run-script natspac tag.')
```
Aniket-Engg, line 72: same here

libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -621,12 +621,32 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
               </a>
             </div>
           }
-          <button id="compileBtn" data-id="compilerContainerCompileBtn" className="btn btn-primary btn-block remixui_disabled mt-3" title="Compile" onClick={compile} disabled={disableCompileButton}>
+          <OverlayTrigger overlay={
+              <Tooltip id="overlay-tooltip">
+                <div className="text-left">
+                <div>Ctrl+s for compiling</div>
+                <div>Ctrl+Shift+s for compiling and on the fly execution:</div>
```
Aniket-Engg, line 628: This can be: `Ctrl+Shift+s for compiling and script execution`

Aniket-Engg (commented):
- Clicking on `Compile And Run` spins the spinner on `Compile` button
- There is a small space between i and copy icon which does n't show any tooltip but copies the tag
- If script doesn't exist, it first says `running <scriptName>` and then shows the error, it would be better to show error before that

Aniket-Engg (approved):

PR #2274, opened 2022-04-10
invalidate cache

bunsenstraat:
When you build a version.json file is created in assets containing the version in package, timestamp & mode.
Webpack writes the chunks as file.${version}.${timestamp}.js like app.0.23.0-dev.1649658546818.js
Loader.js loads the version json and writes script tags loading the correct version files


bunsenstraat:
- TERSERPLUGIN doesn't generate the correct chunk names for polyfills/main.. need a solution for that otherwise it won't load on live/alpha

yann300 (approved):

PR #2275, opened 2022-04-10
catch filepanel activation

bunsenstraat:


libs/remix-ui/search/src/lib/context/context.tsx
```diff
@@ -336,9 +336,12 @@ export const SearchProvider = ({
...
     async function fetchWorkspace() {
-      const workspace = await plugin.call('filePanel', 'getCurrentWorkspace')
-      if(workspace) value.setCurrentWorkspace(workspace.name)
-      setFiles(await getDirectory('/', plugin))
+      try {
+        const workspace = await plugin.call('filePanel', 'getCurrentWorkspace')
+        if (workspace) value.setCurrentWorkspace(workspace.name)
+        setFiles(await getDirectory('/', plugin))
+      } catch (e) {
+      }
```
Aniket-Engg, line 344: I think you should suppress the errors based on certain error message rather than all.

PR #2312, opened 2022-04-15
Allow users to specify gasLimit instead of forcing the estimate

### LINKED ISSUE #2310: pop() or delete error: "The transaction ran out of gas"
The pop and delete methods always give "VM error: out of gas.", 

The code is very very simple, it's just a test:

```
contract Arrays {
    uint [] public myArray;

   function pushToArray(uint number) public {
        myArray.push(number);
    }

    function popArray() public {
        myArray.pop();
    }
}
```

I tried to increase the gas limit, but nothing changes.

The Remix version is **0.23.0**

ernestognw:
Fixes #2310 

## Description

The gas estimation is used to determine if the `gasLimit` provided by the user is enough to execute the transaction, otherwise the UI shows an error regarding it (same for when the estimate exceeds block's `gasLimit`).

For some reason, the tx sent for any contract interaction was forced to use the estimate and fallback to the user's `gasLimit`. Remix should honor whatever the user sets in the `gasLimit` field.

Also [more context here](https://twitter.com/destructioneth/status/1515012325647990794)

### DIFF
--- a/libs/remix-lib/src/execution/txRunnerWeb3.ts
+++ b/libs/remix-lib/src/execution/txRunnerWeb3.ts
@@ -109,8 +109,8 @@ export class TxRunnerWeb3 {
         err = network.name === 'VM' ? null : err // just send the tx if "VM"
         gasEstimationForceSend(err, () => {
           // callback is called whenever no error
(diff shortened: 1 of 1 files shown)


Aniket-Engg:
I think this has been implemented and issue is already closed, so closing this

PR #2340, opened 2022-04-28
Better checks when saving a file

yann300:
- use `askUserPermission` when calling `resolveAndSave` and calling 'web3provider'.
 - always display modal when it's a `writeFile` api call (it isn't possible to remember permission anymore in that case).
 - display a warning when it's a `writeFile` api call.

### DIFF
--- a/apps/remix-ide/src/app/tabs/web3-provider.js
+++ b/apps/remix-ide/src/app/tabs/web3-provider.js
@@ -21,7 +21,6 @@ export class Web3ProviderModule extends Plugin {
     Should be taken carefully and probably not be release as it is now.
   */
   sendAsync (payload) {
-    console.log(payload)
     return this.askUserPermission('sendAsync', `Calling ${payload.method} with parameters ${JSON.stringify(payload.params)}`).then(
       async (result) => {
         if(result){
--- a/apps/remix-ide/src/app/tabs/web3-provider.js
+++ b/apps/remix-ide/src/app/tabs/web3-provider.js
@@ -11,7 +11,7 @@ export const profile = {
 }
 
 export class Web3ProviderModule extends Plugin {
-  constructor (blockchain) {
+  constructor(blockchain) {
     super(profile)
     this.blockchain = blockchain
   }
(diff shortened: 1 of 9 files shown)


libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -124,6 +124,10 @@ export class CompilerImports extends Plugin {
     */
   async resolveAndSave (url, targetPath) {
     try {
+      if (targetPath && this.currentRequest) {
+        const canCall = await this.askUserPermission('writeFile', 'This action will update the path ' + targetPath)
```
bunsenstraat, line 128: writeFile is not a method on this plugin so this will always return false.

bunsenstraat: when executed by scriptrunner it will always return true

apps/remix-ide/src/app/tabs/web3-provider.js
```diff
@@ -21,27 +21,38 @@ export class Web3ProviderModule extends Plugin {
...
-      provider[provider.sendAsync ? 'sendAsync' : 'send'](payload, async (error, message) => {
-        if (error) return reject(error)
-        if (payload.method === 'eth_sendTransaction') {
-          if (payload.params.length && !payload.params[0].to && message.result) {
-            setTimeout(async () => {
-              const receipt = await this.tryTillReceiptAvailable(message.result)
-              if (!receipt.contractAddress) {
-                console.log('receipt available but contract address not present', receipt)
-                return
+    return this.askUserPermission('sendAsync', `Calling ${payload.method} with parameters ${JSON.stringify(payload.params)}`).then(
```
yann300, line 24: it might be better readable to include `askUserPermission` in the promise callback.

yann300: are you sure this is returning a promise?

apps/remix-ide/src/app/tabs/web3-provider.js
```diff
@@ -20,31 +20,41 @@ export class Web3ProviderModule extends Plugin {
...
-      provider[provider.sendAsync ? 'sendAsync' : 'send'](payload, async (error, message) => {
-        if (error) return reject(error)
-        if (payload.method === 'eth_sendTransaction') {
-          if (payload.params.length && !payload.params[0].to && message.result) {
-            setTimeout(async () => {
-              const receipt = await this.tryTillReceiptAvailable(message.result)
-              if (!receipt.contractAddress) {
-                console.log('receipt available but contract address not present', receipt)
-                return
+      this.askUserPermission('sendAsync', `Calling ${payload.method} with parameters ${JSON.stringify(payload.params)}`).then(
```
yann300, line 26: could this be:

`JSON.stringify(payload.params, nul, '\t')` for better formatting

libs/remix-ui/permission-handler/src/lib/permission-dialog.tsx
```diff
@@ -52,12 +52,14 @@ const PermissionHandlerDialog = (props: PermissionHandlerProps) => {
       <h6>{to.displayName} :</h6>
       <p> {to.description || <i>No description Provided</i>}</p>
       {pluginMessage()}
+      { sensitiveCall ? <p className='text-warning'><i className="fas fa-exclamation-triangle mr-2" aria-hidden="true"></i>This action will update your files. Please make sure you trust this plugin.</p> : '' }
```
yann300, line 55: maybe the msg need to be more generic:
`You are going to process a sensitive call. Please make sure you trust this plugin.`

bunsenstraat (approved):

PR #2343, opened 2022-04-29
current file fixed after deletion

### LINKED ISSUE #2273: current file not updated once a file is deleted
Screenshot 2022-04-08 at 5 00 22 PM

Steps to reproduce:

1. Select any file e.g; `2_Owner.sol`
2. Now from FE, right click on any other file and delete
3. Check `Solidity Compiler` plugin, current file is not shown on `Compile` button even if it is selected in editor. It is updated when you open another file in editor

Aniket-Engg:
fixes #2273

### DIFF
--- a/apps/remix-ide/src/app/panels/tab-proxy.js
+++ b/apps/remix-ide/src/app/panels/tab-proxy.js
@@ -41,6 +41,13 @@ export class TabProxy extends Plugin {
 
       if (this.fileManager.mode === 'browser') {
         name = name.startsWith(workspace + '/') ? name : workspace + '/' + name
+        // If deleted file is not current file and not an active tab in editor,
+        // ensure current file is active in the editor
+        if (this.fileManager.currentFile() && name !== this.fileManager.currentFile()) {
+          const currentFile = this.fileManager.currentFile()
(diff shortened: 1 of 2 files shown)


apps/remix-ide/src/app/panels/tab-proxy.js
```diff
@@ -41,6 +41,13 @@ export class TabProxy extends Plugin {
 
       if (this.fileManager.mode === 'browser') {
         name = name.startsWith(workspace + '/') ? name : workspace + '/' + name
+        // If deleted file is not current file and not an active tab in editor,
+        // ensure current file is active in the editor
+        if (this.fileManager.currentFile() && name !== this.fileManager.currentFile()) {
+          const currentFile = this.fileManager.currentFile()
+          const currentFileTabPath = currentFile.startsWith(workspace + '/') ? currentFile : workspace + '/' + currentFile
+          this.removeTab(name, { name: currentFileTabPath })
```
yann300, line 49: you should probably return here, otherwise you'll call `removeTab` 2 times.

Aniket-Engg: fixed

yann300 (approved):

PR #2353, opened 2022-05-03
create workspace using templates

Aniket-Engg:
fixes #1102 

Templates for workspace introduced. A template will help user to decide the starting point for a project.

Existing default workspace will still be the default choice for template. 2 new templates have been introduced for now: Blank & ERC20.

PR #2367, opened 2022-05-06
Templates in dirs

Aniket-Engg:


Aniket-Engg:
closed in support of #2353 & #2368

PR #2409, opened 2022-05-18
oz721 template added

Aniket-Engg:


libs/remix-ws-templates/src/templates/ozerc20/scripts/ethers-lib.ts
```diff
@@ -1,4 +1,4 @@
-import { ethers } from 'ethers-lib'
+import { ethers } from 'ethers'
```
yann300, line 1: the name needs to be preffixed with `-lib`, same for web3.ts

yann300 (approved):

Issue #2385, opened 2022-05-11
color of mainpanel plugin icon should be changed with theme


LianaHus:

LianaHus:
it works on theme change but not on loading

PR #2412, opened 2022-05-18
fixed icons' colors for light theme

LianaHus:
fix https://github.com/ethereum/remix-project/issues/2385

### DIFF
--- a/apps/remix-ide/src/app/panels/tab-proxy.js
+++ b/apps/remix-ide/src/app/panels/tab-proxy.js
@@ -24,7 +24,7 @@ export class TabProxy extends Plugin {
     this.themeQuality = 'dark'
   }
 
-  onActivation () {
+  async onActivation () {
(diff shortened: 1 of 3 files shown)


bunsenstraat (approved):

Aniket-Engg (approved):

apps/remix-ide/src/app/panels/tab-proxy.js
```diff
@@ -169,6 +169,9 @@ export class TabProxy extends Plugin {
     this.on('manager', 'pluginDeactivated', (profile) => {
       this.removeTab(profile.name)
     })
+
+    this.themeQuality = (await this.call('theme', 'currentTheme') ).quality
```
ioedeveloper, line 173: Shouldn't this be in a try/catch? Because if the plugin call fails, it may break the UI.

Issue #2198, opened 2022-03-17
Improve Slither analisys UX


yann300:
- [ ] filter out solidity library import.
 - [ ] make it a stand alone plugin.
 - [ ] maybe improve remixd install / initialization.
 - [ ] make some promotion with @vermouth22 @ryestew. If there is not much people using it, set up a public api.

bunsenstraat:
https://github.com/ethereum/remix-project/pull/2450

Aniket-Engg:
We should make it more visible in some ways. Stand alone plugin can be the last option if we couldn't find anything specific.

PR #2450, opened 2022-05-25
Install slither cmd

Aniket-Engg:
Related to #2198 

Slither will be install along with remixd installation and there will be an option with remixd to install it like: `remixd --install slither`


libs/remixd/src/services/slitherClient.ts
```diff
@@ -25,6 +25,28 @@ export class SlitherClient extends PluginClient {
     this.currentSharedFolder = currentSharedFolder
   }
 
+  install (): void {
+    try {
+      console.log('\x1b[32m%s\x1b[0m', `[Slither Installation]: slither requires Python3.6+ (pip3) and solc, the Solidity compiler. They should be installed on your system`)
```
yann300, line 30: I don't know if this is possible but we could publish a docker that'd contains all the dependencies. And retrieve and run that docker from here.

yann300: mainly because we shouldn't ask the user to install manually pip3 and solc.

libs/remixd/src/bin/remixd.ts
```diff
@@ -63,6 +63,7 @@ function errorHandler (error: any, service: string) {
     .description('Establish a two-way websocket connection between the local computer and Remix IDE for a folder')
     .option('-u, --remix-ide  <url>', 'URL of remix instance allowed to connect')
     .option('-s, --shared-folder <path>', 'Folder to share with Remix IDE (Default: CWD)')
+    .option('-i, --install <name>', 'Module name to install locally')
```
yann300, line 66: coud you this be run be default in a post install script?

libs/remixd/src/services/slitherClient.ts
Aniket-Engg: solc requirement is removed

libs/remixd/src/services/slitherClient.ts
```diff
@@ -14,7 +14,7 @@ export class SlitherClient extends PluginClient {
 
   constructor (private readOnly = false) {
     super()
-    this.methods = ['analyse']
+    this.methods = ['analyse', 'install']
```
yann300, line 17: you don't need to add `install` here if you don't want to use it from the UI. I am not even sure it's necessary at all.

Aniket-Engg: it is removed

libs/remixd/src/bin/remixd.ts
Aniket-Engg: this is done now.

PR #2462, opened 2022-05-30
do not compile if  not sol files & fix typescript autocompletion

yann300:
- do not compile if not sol files 
 - fix typescript autocompletion

### DIFF
--- a/apps/remix-ide/src/app/editor/editor.js
+++ b/apps/remix-ide/src/app/editor/editor.js
@@ -1,5 +1,6 @@
 'use strict'
 import React from 'react' // eslint-disable-line
+import { resolve } from 'path'
 import { EditorUI } from '@remix-ui/editor' // eslint-disable-line
 import { Plugin } from '@remixproject/engine'
 import * as packageJson from '../../../../../package.json'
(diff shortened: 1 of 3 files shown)


apps/remix-ide/src/app/editor/editor.js
```diff
@@ -150,6 +151,25 @@ class Editor extends Plugin {
...
+        // so Monaco can provide auto completion
+        let content = await this.call('fileManager', 'readFile', name)
+        const paths = name.split('/')
+        paths.pop()
+        const fromPath = paths.join('/') // get current execution context path
+        for (const match of content.matchAll(/import\s+.*\s+from\s+(?:"(.*?)"|'(.*?)')/g)) {
+          let path = match[2]
+          if (path.startsWith('./') || path.startsWith('../')) path = resolve(fromPath, path)
+          if (path.startsWith('/')) path = path.substring(1)
+          if (!path.endsWith('.ts')) path = path + '.ts'
```
Aniket-Engg, line 167: This creates an issue for npm module import like ethers and web3 and shows errors:
Screenshot 2022-05-31 at 12 39 52 PM

libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
```diff
@@ -131,6 +131,11 @@ export class CompileTabLogic {
 
   runCompiler (externalCompType) {
     try {
+      const currentFile = this.api.currentFile
+      if (currentFile && (currentFile.endsWith('.ts') || currentFile.endsWith('.js') || currentFile.endsWith('.txt') || currentFile.endsWith('.log'))) {
```
Aniket-Engg, line 135: I think it is better to check files available for compilation like .sol, .vy etc as there are a great number of other extensions possible.

I can still compile the .json and .md file

yann300: yes I know, i'll find an alternative

apps/remix-ide/src/app/editor/editor.js
yann300: done

ioedeveloper (approved):

Aniket-Engg (commented):
Compilation is not triggered for non-sol files that is fine. But autocomplete is still not working for an lib imported in an imported file.

Aniket-Engg (approved):

Issue #2469, opened 2022-06-01
null workspace on loading Remix


Aniket-Engg:
![Screenshot 2022-06-01 at 6 28 48 PM](https://user-images.githubusercontent.com/30843294/171410284-fbdd12d4-c143-47c4-9a86-b826984fc5a8.png)

It is difficult to reproduce.

Source of this issue should be around this https://github.com/ethereum/remix-project/blob/master/libs/remix-ui/workspace/src/lib/actions/workspace.ts#L249

PR #2470, opened 2022-06-01
added Berlin, fixed config file logic

LianaHus:
Fixed https://github.com/ethereum/remix-project/issues/2469

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
+++ b/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
@@ -79,10 +79,12 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
   }, [workspaceName])
 
   useEffect(() => {
-    if (state.useFileConfiguration)
+    if (state.useFileConfiguration) {
       api.fileExists(defaultPath).then((exists) => {
         if (!exists || state.useFileConfiguration ) createNewConfigFile()
       })
+      setToggleExpander(true)
+    }
   }, [state.useFileConfiguration])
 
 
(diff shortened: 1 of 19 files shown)


libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -53,17 +66,27 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
...
-    api.setAppParameter('configFilePath', "/compiler_config.json")
+    api.setAppParameter('configFilePath', "compiler_config.json")
+
+    if (state.useFileConfiguration) {
+      api.fileExists("compiler_config.json").then((exists) => {
+        if (!exists || state.useFileConfiguration ) createNewConfigFile()
+      })
+    }
+
+    api.setAppParameter('configFilePath', "compiler_config.json")
```
yann300, line 77: you probably don't need to do it 2 times (see line 69)

libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -53,17 +66,27 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
...
     setShowFilePathInput(false)
   }, [workspaceName])
 
+  useEffect(() => {
+
```
yann300, line 82: linting

ryestew:
In this PR - but maybe also in beta - if you create a new file with the path - it will make a new file.  This info needs to be updated in the docs. (note to self)

ryestew:
and very important - are the matomo tags here?  Please make sure the button to open the advanced panel is added and how often people are using the .json file for compilation - and more too.

LianaHus:
> Here's a screenshot from beta - which does not have anything to do with this PR...maybe... image
> 
> not the 2 files named lala.json - beta asked me if I wanted to create a file named that when I put that in the path text box. So it hadn't seen the current file there.
> 
> When you try to delete one of the files - the FE selects both - but only 1 gets deleted - but the remaining can't be selected.

@ioedeveloper

yann300 (approved):

Issue #2502, opened 2022-06-07
Issue in changing compiler config file


Aniket-Engg:
Steps to reproduce:
1. Select `Use configuration file` option and click on `Change` button
2. Put some name eg: `asdf` in the textbox and click `Enter`
3. Click on `Cancel` on `New Configuration File` modal
4.  Create new workspace, it has `asdf.json` in the workspace and selected as configuration file in `Solidity Compiler`

![Screenshot 2022-06-07 at 1 07 41 PM](https://user-images.githubusercontent.com/30843294/172323245-28e1ba17-3652-4b54-8f36-9fe7676d932d.png)

PR #2530, opened 2022-06-13
fixing bug of saved path

LianaHus:
fix https://github.com/ethereum/remix-project/issues/2502


Aniket-Engg (commented):
This fixed the old bug.

New issue is that when I click on `change` and edit the file name to `compiler_c.json` , click enter and `Cancel` from modal. Further, when I again click on `change`, textbox is filled with `compiler_c.json` which should be `compiler_config.json`

LianaHus:
ok :)

Aniket-Engg (approved):

Issue #2446, opened 2022-05-24
tool tips- deploy & run


ryestew:
<img width="510" alt="image" src="https://user-images.githubusercontent.com/6050001/170132739-be829377-a591-48a9-a73a-cec98100192e.png">

current: ⚠ Selected *.abi file allows accessing contracts, select and compile *.sol file to deploy and access one.
suggested: 

other issues: if nothing is compiled and the select box is blank this tool tip doesn't really help.

Issue #2287, opened 2022-04-11
Check title properties


yann300:
Check all the titles and make sure they are up to date, easily undestable.

PR #2545, opened 2022-06-16
update titles

ryestew:
fix: https://github.com/ethereum/remix-project/issues/2446
fix: https://github.com/ethereum/remix-project/issues/2287


libs/remix-ui/run-tab/src/lib/components/contractDropdownUI.tsx
```diff
@@ -133,12 +133,12 @@ export function ContractDropdownUI (props: ContractDropdownProps) {
...
       setContractOptions({
         disabled: false,
-        title: 'Select contract for Deploy or At Address.'
+        title: 'Select compiled a contract to deploy or to use with At Address.'
       })
     } else {
       setContractOptions({
         disabled: true,
-        title: loadType === 'sol' ? '⚠ Select and compile *.sol file to deploy or access a contract.' : '⚠ Selected *.abi file allows accessing contracts, select and compile *.sol file to deploy and access one.'
+        title: loadType === 'sol' ? '⚠ Select and compile *.sol file to deploy or access a contract.' : '⚠ When there is a compiled .sol file, the choice of contracts to deploy or to use with AtAddress is made here'
```
Aniket-Engg, line 141: I could not find a way to see this title. 

Still, warning sign should be removed and title for second condition should be updated

libs/remix-ui/run-tab/src/lib/components/environment.tsx
```diff
@@ -29,7 +29,7 @@ export function EnvironmentUI (props: EnvironmentProps) {
             )
           }
         </select>
-        <a href="https://remix-ide.readthedocs.io/en/latest/run.html#run-setup" target="_blank" rel="noreferrer"><i className="udapp_infoDeployAction ml-2 fas fa-info" title="check out docs to setup Environment"></i></a>
+        <a href="https://remix-ide.readthedocs.io/en/latest/run.html#run-setup" target="_blank" rel="noreferrer"><i className="udapp_infoDeployAction ml-2 fas fa-info" title="Click for docs about Environment"></i></a>
```
Aniket-Engg, line 32: link can be updated to : https://remix-ide.readthedocs.io/en/latest/run.html#environment

libs/remix-ui/run-tab/src/lib/components/gasPrice.tsx
```diff
@@ -10,7 +10,7 @@ export function GasPriceUI (props: GasPriceProps) {
   return (
     <div className="udapp_crow">
       <label className="udapp_settingsLabel">Gas limit</label>
-      <input type="number" className="form-control udapp_gasNval udapp_col2" id="gasLimit" value={props.gasLimit} onChange={handleGasLimit} />
+      <input type="number" className="form-control udapp_gasNval udapp_col2" title="The default gas limit is 3M. This does not work for all transactions." id="gasLimit" value={props.gasLimit} onChange={handleGasLimit} />
```
Aniket-Engg, line 13: `This does not work for all transactions` looks confusing to me

libs/remix-ui/run-tab/src/lib/components/value.tsx
```diff
@@ -47,7 +47,7 @@ export function ValueUI (props: ValueProps) {
 
   return (
     <div className="udapp_crow">
-      <label className="udapp_settingsLabel" data-id="remixDRValueLabel">Value</label>
+      <label className="udapp_settingsLabel" data-id="remixDRValueLabel">Amount</label>
```
Aniket-Engg, line 50: I am not sure if this is a good idea as when we send tx using web3 libs, they call it `value` and even inside Solidity, it is accessed as `msg.value`

libs/remix-ui/run-tab/src/lib/components/value.tsx
```diff
@@ -58,7 +58,7 @@ export function ValueUI (props: ValueProps) {
           className="form-control udapp_gasNval udapp_col2"
           id="value"
           data-id="dandrValue"
-          title="Enter the value and choose the unit"
+          title="Enter an amount and choose its unit"
```
Aniket-Engg, line 61: It can be `Enter a value to send with tx`

libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
```diff
@@ -115,14 +115,14 @@ export const runTabInitialState: RunTabState = {
     providerList: [{
       id: 'vm-mode-london',
       dataId: 'settingsVMLondonMode',
-      title: 'Execution environment does not connect to any node, everything is local and in memory only.',
+      title: 'Execution environment does not connect to any node. All data is stored in browser memory only.',
```
Aniket-Engg, line 118: I think it can be like: `Environment local to Remix. Data is stored in browser and deployed contracts disappear on reload`

libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
```diff
@@ -115,14 +115,14 @@ export const runTabInitialState: RunTabState = {
...
       value: 'vm-london',
       fork: 'london',
       content: 'JavaScript VM (London)'
     }, {
       id: 'vm-mode-berlin',
       dataId: 'settingsVMBerlinMode',
-      title: 'Execution environment does not connect to any node, everything is local and in memory only.',
+      title: 'Execution environment does not connect to any node. All data is stored in browser memory only.',
```
Aniket-Engg, line 125: same here

libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
```diff
@@ -135,8 +135,7 @@ export const runTabInitialState: RunTabState = {
     }, {
       id: 'web3-mode',
       dataId: 'settingsWeb3Mode',
-      title: `Execution environment connects to node at localhost (or via IPC if available), transactions will be sent to the network and can cause loss of money or worse!
-      If this page is served via https and you access your node via http, it might not work. In this case, try cloning the repository and serving it via http.`,
+      title: `Execution environment connects to an external node. For security, only connect to trusted networks. If Remix is served via https and your node is accessed via http, it might not work. In this case, try cloning the repository and serving it via http.`,
```
Aniket-Engg, line 138: this is very long and difficult to read at once. `In this case, try cloning the repository and serving it via http.` can be removed

libs/remix-ui/run-tab/src/lib/components/contractDropdownUI.tsx
ryestew: This only comes up when there is no active .sol file in the editor nor an abi in the editor.
image

I removed the warning icons. I think the text is OK as I had updated it.  If you'd like an edit, please made a suggestion.

libs/remix-ui/run-tab/src/lib/components/value.tsx
ryestew: Yes but it is confusing for users.  A user wants to know where to send some amount of wei or gwei or eth or whatever to a function.

libs/remix-ui/run-tab/src/lib/components/value.tsx
ryestew: Same as above.  I think value is too general a concept.

libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
ryestew: I agree this is veering to more info than is appropriate for a tooltip - but it could still be helpful if someone is patient enough to re-activate the tooltip once it fades out.  Or if this is made into a custom tooltip - maybe it should have a longer live in the active state.  So I'd prefer to leave this edit for now.

libs/remix-ui/run-tab/src/lib/components/value.tsx
ryestew: Sending "value" doesn't make immediate sense.

Aniket-Engg (approved):

Issue #2377, opened 2022-05-09
More Descriptive and Dynamic Transaction Recording (enhancement)


RuiXi-Verilog:
## About Dynamic Transaction Recording
I find it really useful to quickly deploy contracts and run quick unit tests using the Transaction Recording feature. However, the Recording only replays exactly the same contract bytecode as I recorded, so every time I update my code, the old Recording deploys the old ones. I was wondering if it is possible to **make Transaction Recording dynamically link to the source files or the compiled artifacts**, and thus the Recording replay does an up-to-date deployment.

### Pros and Cons
- (+) Deploy everything up-to-date!
- (+) No need to store bytecode in `recording.json`
- (+) Pave the way to a fast-paced DevTestOps pipeline.
- (-) Rely on existing source code and/or artifacts. It means you cannot just upload a `recording.json` and run everything.

### One Possible Format
```json
{
  "timestamp": 1650162593412,
  "record": {
    "value": "0",
    "parameters": [
        "created{1650162569424}"
    ],
    "abi": "0x28e2ec147ccaac8c352ab53bd66b8187d2ab8c0dc4b881cec5748aa81d5051b5",
    "contractName": "Attacker1",
    "contractPath": "contracts/Attacker1.sol", 
    "artifactPath": "contracts/artifacts/Attacker1.json",
    "linkReferences": {},
    "name": "",
    "inputs": "(address)",
    "type": "constructor",
    "from": "account{1}"
  }
}
```

## About DescriptiveTransaction Recording
Also looking forward to a more **descriptive parameter list to function execution transactions**. And thus I can read `recording.json` directly to figure out what the payloads mean. Similar to previous cons, it requires source code or artifacts available to extract parameter names. 

### What it is now
```json
{
  "timestamp": 1650474814976,
  "record": {
    "value": "0",
    "parameters": [
      "created{1650474395581}",
      "5000000000000000000"
    ],
    "to": "created{1650474338320}",
    "abi": "0x3e91d4a4cb3d9af06253d1e2a6db8ea31b4981cfbf80b7c1a367ad75e462ea98",
    "name": "approve",
    "inputs": "(address,uint256)",
    "type": "function",
    "from": "account{0}"
  }
},
```

### One Possible Format
```json
{
  "timestamp": 1650474814976,
  "record": {
    "value": "0",
    "name": "approve",
    "parameters": [
      {
        "name": "spender", 
        "type": "address",
        "value": "created{1650474395581}"
      },
      {
[...]

PR #2571, opened 2022-06-27
Improve recorder

yann300:
fix https://github.com/ethereum/remix-project/issues/2377

in order to test this impl:

 - first step: grab this contract code : https://github.com/ethereum/remix-project/pull/2571/files#diff-b25cd6d5966988c80d845026ef2b3b83f3a23c931bab6ce2831c79210c09df4aR466
 - compile and deploy storage.
 - save a scenario
 - at the json root, set `live` to true. This means that when running the scenario, instead of getting the hardcoded bytecode from the json file, it will get the bytecode from the last compilation result.
 - run the scenario, `retrieve` should return 350.
 - second step: replace 350 by another number.
 - compile and reun the scenarion again.
 - `retrieve` should return this new number.

### DIFF
--- a/apps/remix-ide-e2e/src/tests/recorder.test.ts
+++ b/apps/remix-ide-e2e/src/tests/recorder.test.ts
@@ -414,7 +414,6 @@ const liveModeScenario = {
     "account{0}": "0x5B38Da6a701c568545dCfcB03FcB875f56beddC4"
   },
   "linkReferences": {},
-  "live": true,
   "transactions": [
     {
       "timestamp": 1656329164297,
--- a/libs/remix-ui/run-tab/src/lib/components/recorderCardUI.tsx
+++ b/libs/remix-ui/run-tab/src/lib/components/recorderCardUI.tsx
@@ -27,8 +27,7 @@ export function RecorderUI (props: RecorderProps) {
           <div className="ml-2 mb-2 badge badge-pill badge-primary" title="The number of recorded transactions">{props.count}</div>
         <OverlayTrigger placement={'right'} overlay={
           <Tooltip className="text-nowrap" id="info-recorder">
-            <span>Save transactions (deployed contracts and function executions) and replay then in an
-            another environment. <br/> e.g Transactions created in Javascript VM can be replayed in the Injected Web3.
+            <span>Save transactions (deployed contracts and function executions) and replay them in another environment. <br/> e.g Transactions created in Javascript VM can be replayed in the Injected Web3.
           </span>
           </Tooltip>
         }>
@@ -44,12 +43,12 @@ export function RecorderUI (props: RecorderProps) {
       <div className={`flex-column ${toggleExpander ? "d-flex" : "d-none"}`}>
         <div className="mb-1 mt-1 fmt-2 custom-control custom-checkbox mb-1">
           <input ref={inputLive} type="checkbox" id="livemode-recorder" className="custom-control-input custom-select" name="input-livemode"/>
-          <label className="form-check-label custom-control-label" data-id="runtabLivemodeInput" htmlFor="livemode-recorder">When running transactions, get the contracts bytecodes from the last compilation result.</label>
+          <label className="form-check-label custom-control-label" data-id="runtabLivemodeInput" htmlFor="livemode-recorder">Use live mode (Run transactions against latest compiled contracts).</label>
         </div>     
         <div className="mb-1 mt-1 udapp_transactionActions">
         <OverlayTrigger placement={'right'} overlay={
           <Tooltip className="text-nowrap" id="tooltip-save-recorder">
-            <span>Save {props.count} transaction(s) to JSON.
+            <span>Save {props.count} transaction(s) as scenario file.
           </span>
           </Tooltip>
         }>
@@ -57,7 +56,7 @@ export function RecorderUI (props: RecorderProps) {
         </OverlayTrigger>
         <OverlayTrigger placement={'right'} overlay={
           <Tooltip className="text-nowrap" id="tooltip-run-recorder">
-            <span>Run Transaction from the current scenario file.
+            <span>Run transaction(s) from the current scenario file.
           </span>
           </Tooltip>
         }>


apps/remix-ide/src/app/tabs/runTab/model/recorder.js
```diff
@@ -275,6 +302,7 @@ class Recorder {
       var options = json.options || {}
       var abis = json.abis || {}
       var linkReferences = json.linkReferences || {}
+      var live = json.live || false
```
Aniket-Engg, line 305: can be `const` and other variables above too

apps/remix-ide-e2e/src/tests/ballot_0_4_11.test.ts
```diff
@@ -78,7 +78,7 @@ module.exports = {
     browser
       .openFile('Untitled.sol')
       .clickLaunchIcon('udapp')
-      .click('*[data-id="settingsWeb3Mode"]')
+      .click('*[data-id="settingsWeb3Mode"]')
```
Aniket-Engg, line 81: this can be reverted

Aniket-Engg (commented):
Suggestions:

1. I think `live` can always be there in the scenario, set `false` by default. It will be easy for user to edit it
2. Not sure if there can be a checkbox or an additional button on the recorder UI to run the scenario with the latest compiled contract. That can be more user friendly.
3. In terminal, while running a scenario, it says: `Running 2 transaction(s) ...` , it can be updated as `Running 2 transaction(s) in live mode ...` for live mode only.

libs/remix-ui/run-tab/src/lib/components/recorderCardUI.tsx
```diff
@@ -22,24 +25,45 @@ export function RecorderUI (props: RecorderProps) {
         <div className="d-flex">
           <label className="mt-1 udapp_recorderSectionLabel">Transactions recorded</label>
           <div className="ml-2 mb-2 badge badge-pill badge-primary" title="The number of recorded transactions">{props.count}</div>
+        <OverlayTrigger placement={'right'} overlay={
+          <Tooltip className="text-nowrap" id="info-recorder">
+            <span>Save transactions (deployed contracts and function executions) and replay then in an
```
Aniket-Engg, line 30: `replay then in an another environment.` => `replay them in another environment.`

apps/remix-ide-e2e/src/tests/recorder.test.ts
```diff
@@ -364,3 +408,92 @@ const scenario = {
     ]
   }
 }
+
+const liveModeScenario = {
+  "accounts": {
+    "account{0}": "0x5B38Da6a701c568545dCfcB03FcB875f56beddC4"
+  },
+  "linkReferences": {},
+  "live": true,
```
Aniket-Engg, line 417: this can be removed

libs/remix-ui/run-tab/src/lib/components/recorderCardUI.tsx
```diff
@@ -22,24 +25,45 @@ export function RecorderUI (props: RecorderProps) {
...
+          <Tooltip className="text-nowrap" id="tooltip-save-recorder">
+            <span>Save {props.count} transaction(s) to JSON.
+          </span>
+          </Tooltip>
+        }>
+          <button className="btn btn-sm btn-info savetransaction udapp_recorder" onClick={triggerRecordButton}>Save</button>
+        </OverlayTrigger>
+        <OverlayTrigger placement={'right'} overlay={
+          <Tooltip className="text-nowrap" id="tooltip-run-recorder">
+            <span>Run Transaction from the current scenario file.
```
Aniket-Engg, line 60: It can be similar to `Save` button tooltip

`Run Transaction from` => `Run transaction(s) from the current scenario file`

libs/remix-ui/run-tab/src/lib/components/recorderCardUI.tsx
```diff
@@ -22,24 +25,45 @@ export function RecorderUI (props: RecorderProps) {
...
-        </div>
+      <div className={`flex-column ${toggleExpander ? "d-flex" : "d-none"}`}>
+        <div className="mb-1 mt-1 fmt-2 custom-control custom-checkbox mb-1">
+          <input ref={inputLive} type="checkbox" id="livemode-recorder" data-id="runtabLivemodeInput" className="custom-control-input custom-select" name="input-livemode"/>
+          <label className="form-check-label custom-control-label" htmlFor="livemode-recorder">When running transactions, get the contracts bytecodes from the last compilation result.</label>
+        </div>     
+        <div className="mb-1 mt-1 udapp_transactionActions">
+        <OverlayTrigger placement={'right'} overlay={
+          <Tooltip className="text-nowrap" id="tooltip-save-recorder">
+            <span>Save {props.count} transaction(s) to JSON.
```
Aniket-Engg, line 52: Either `to JSON` can be removed to make it `Save {props.count} transaction(s)` or
it can be `Save {props.count} transaction(s) as scenario file` to keep it similar to Run button tooltip

libs/remix-ui/run-tab/src/lib/components/recorderCardUI.tsx
```diff
@@ -22,24 +25,45 @@ export function RecorderUI (props: RecorderProps) {
...
-        <div className="mb-2 udapp_transactionActions">
-          <i className="fas fa-save savetransaction udapp_recorder udapp_icon"
-            onClick={triggerRecordButton} title="Save Transactions" aria-hidden="true">
-          </i>
-          <i className="fas fa-play runtransaction udapp_runTxs udapp_icon" title="Run Transactions" data-id="runtransaction" aria-hidden="true" onClick={handleClickRunButton}></i>
-        </div>
+      <div className={`flex-column ${toggleExpander ? "d-flex" : "d-none"}`}>
+        <div className="mb-1 mt-1 fmt-2 custom-control custom-checkbox mb-1">
+          <input ref={inputLive} type="checkbox" id="livemode-recorder" className="custom-control-input custom-select" name="input-livemode"/>
+          <label className="form-check-label custom-control-label" data-id="runtabLivemodeInput" htmlFor="livemode-recorder">When running transactions, get the contracts bytecodes from the last compilation result.</label>
```
Aniket-Engg, line 47: `get the contracts bytecodes from the last compilation result` - It seems me confusing from user point of view as terminology it used in Remix codebase and user can be unaware about it. We can make it like: `Use live mode (Run transactions against latest compiled copy of contract)`

Issue #2415, opened 2022-05-19
Implement upgrades option for proxy contracts (Openzeppelin UUPS pattern)


ioedeveloper:
A sub-issue of https://github.com/ethereum/remix-project/issues/2221

PR #2594, opened 2022-07-07
Support UUPS Contract Upgrade Via ERC1967 Proxy

ioedeveloper:
Closes https://github.com/ethereum/remix-project/issues/2415


libs/remix-ui/run-tab/src/lib/actions/deploy.ts
```diff
@@ -135,6 +135,7 @@ export const createInstance = async (
   args,
   deployMode: DeployMode[]) => {
   const isProxyDeployment = (deployMode || []).find(mode => mode === 'Deploy with Proxy')
+  const isContractUpgrade = (deployMode || []).find(mode => mode === 'Upgrade Contract')
```
yann300, line 138: maybe the label should be `Upgrade Proxy`

libs/remix-ui/run-tab/src/lib/components/contractGUI.tsx
```diff
@@ -215,90 +166,194 @@ export function ContractGUI (props: ContractGUIProps) {
...
+                  Use last deployed ERC1967 contract
+                </label>
+              </div>
+              {
+                !useLastProxy ? 
+                <div className="mb-2">
+                  <label className='mt-2 text-left d-block'>Proxy Address: </label>
+                  <input style={{ height: 32 }} className="form-control udapp_input" placeholder='proxy address' title='Enter previously deployed proxy address on the selected network' onChange={handleSetProxyAddress} />
+                </div> :
+                <span className='text-capitalize'>{ proxyAddress ? shortenAddress(proxyAddress) : 'No proxy address available' }</span>
```
yann300, line 351: I would maybe not shorten the address

yann300:
in both case (deploy proxy and upgrade proxy), 2 transactions are run, so I think it could make sense to deploy a validation modal before these txs happen. this modal will clearly explain what the txs are about. if you guys approve, this can be done in another PR.

apps/remix-ide/src/blockchain/blockchain.js
```diff
@@ -173,12 +175,33 @@ export class Blockchain extends Plugin {
   
         return this.call('terminal', 'logHtml', log)
       }
+      if (networkInfo.name === 'VM') this.config.set('vm/proxy', address)
```
yann300, line 178: the use case is a bit limited, mainly because it won't work when deploying more than 1 instance. we can improve that later on.
As I see we could create a JSON file insted of putting it in the localstorage.
The json file would save stuff like the deployment time, the contract name, the network, ...
Then the upgrade proxy would display a dropdown of just an input when the user would need to paste the correct address from the JSON file.
can be addressed separately of course

yann300 (approved):

yann300:
- add a UUPS contract and a simple contract like `contract test {}`
 - compile
 - switch to the uups contract, we can see the extra option
 - switch to `test` the contract option ar still visible..

yann300:
when the selected contract is not UUPS, the extra option shouldn't appear

yann300:
Compiling a UUPS contract doesn't fill the contract list in udapp. it works with normal contracts.

ioedeveloper:
> Compiling a UUPS contract doesn't fill the contract list in udapp. it works with normal contracts.

I can't reproduce this. Please provide a copy of the contract you're compiling and steps to reproduce it.

yann300:
> I can't reproduce this. Please provide a copy of the contract you're compiling and steps to reproduce it.

nevermind, it seems to work now

yann300:
> Compiling a UUPS contract doesn't fill the contract list in udapp. it works with normal contracts.

ah here it is:
udapp is not filled when compiling this contract.
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.4;

import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";

contract MyToken is Initializable, ERC20Upgradeable, OwnableUpgradeable, UUPSUpgradeable {
    /// @custom:oz-upgrades-unsafe-allow constructor
    constructor() {
        _disableInitializers();
    }

    function initialize() initializer public {
        __ERC20_init("MyToken", "MTK");
        __Ownable_init();
        __UUPSUpgradeable_init();
    }

    function _authorizeUpgrade(address newImplementation)
        internal
        onlyOwner
        override
    {}
}

contract test {

}

```

libs/remix-core-plugin/src/lib/openzeppelin-proxy.ts
```diff
@@ -28,16 +28,17 @@ export class OpenZeppelinProxy extends Plugin {
...
 
     if (this.kind === 'UUPS') {
       Object.keys(contracts).map(name => {
         const abi = contracts[name].abi
         const initializeInput = abi.find(node => node.name === 'initialize')
 
-        if (initializeInput) {
-          inputs[name] = {
+        inputs[name] = {
+          options: [{ title: 'Deploy with Proxy', active: false }, { title: 'Upgrade Contract', active: false }],
```
yann300, line 40: `Upgrade` Proxy`?

yann300 (approved):

PR #2549, opened 2022-06-17
renamed envs

LianaHus:
fix https://github.com/ethereum/remix-project/issues/2538
fix https://github.com/ethereum/remix-project/issues/2431

PR #2595, opened 2022-07-08
env name updated

Aniket-Engg:
part of #2549


vermouth22:
Shouldn't these also be changed in the readthedocs? "Web3" to "Provider" and the image that still says Javascript VM... 
image

PR #2602, opened 2022-07-10
add vyper plugin

yann300:
add vyper plugin to remix-project code base


Aniket-Engg (commented):
It would be great to add some e2e tests

Aniket-Engg (approved):

PR #2603, opened 2022-07-11
update package

bunsenstraat:


Aniket-Engg (commented):
I think yarn.lock should also be updated with deps

Aniket-Engg (approved):

PR #2646, opened 2022-07-18
Nightwatch update to 2.3

bunsenstraat:
This is required to allow certain tests to run for the autocomplete functionality coming up in another PR.
However there is 1 test which can not be run anymore, which is uploading files to the file explorer. Setvalue on the file input doesn't work anymore. Maybe there will be a fix in the future.


apps/remix-ide-e2e/src/tests/specialFunctions.test.ts
```diff
@@ -101,7 +101,7 @@ module.exports = {
             .journalLastChildIncludes('to: CheckSpecials.(receive)')
             .journalLastChildIncludes('value: 1 wei')
             .journalLastChildIncludes('data: 0x')
-            .perform(done)
+            .perform(done())
```
yann300, line 104: shouldn't this be `.perform(done)`
or
```
.perform(() => { done() })
```

yann300 (approved):

apps/remix-ide-e2e/src/tests/specialFunctions.test.ts
bunsenstraat: https://nightwatchjs.org/api/perform.html
It's a function passed nothing unless you want to pass and err
    performAsync(function(err) {
      if (err) {
        done(err);
        return;
      }
      // ...
    });

PR #2669, opened 2022-07-21
compiler trigger

bunsenstraat:
allows the compiler to stop the previous task when
compilerRetriggerMode: CompilerRetriggerMode.regtrigger
This will be used when the compiler is run continuously during the autocomplete checks in the editor.
After the first file is compiled and before an import is handled the process will be stopped.


libs/remix-solidity/src/compiler/compiler.ts
```diff
@@ -275,6 +277,9 @@ export class Compiler {
 
     this.state.worker.addEventListener('message', (msg: Record <'data', MessageFromWorker>) => {
       const data: MessageFromWorker = msg.data
+      if (this.state.compilerRetriggerMode == CompilerRetriggerMode.retrigger && data.timestamp !== this.state.compilationStartTime) {
```
yann300, line 280: see the function above `loadInternal` this is called when the compiler is loaded is a old fashionned way (using script tag), probably that this function should be updated too.

bunsenstraat: I assume that function doesn't do any import handling? in that case it would have no use.

yann300: you mean solidity import?
once a compiler is loaded with loadInternal (which should not actually happen a lot anymore), solidity import should still function..

Issue #688, opened 2020-12-21
Fix debugger navigation panel


Aniket-Engg:
Currently, if we want to see the details on the lower sections with each step, we move the step and scroll down to check the values, further scroll up and move to next step/breakpoint and scroll down again and so on.

By fixing the navigation panel, one can scroll to the related section and make it visible just below the navigation panel and quickly perform the debugging.

PR #2681, opened 2022-07-25
Debuger UI

LianaHus:
fix https://github.com/ethereum/remix-project/issues/688

### DIFF
--- a/libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
+++ b/libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
@@ -47,7 +47,7 @@ export const DebuggerUI = (props: DebuggerUIProps) => {
 
   useEffect(() => {
     handleResize()
-  }, [state.debugging, state.isActive])
+  }, [])
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
```diff
@@ -36,6 +36,28 @@ export const DebuggerUI = (props: DebuggerUIProps) => {
...
+  const panelsRef = useRef<HTMLDivElement>(null)
+  const debuggerTopRef = useRef(null)
+
+  const handleResize = () => {
+    if (panelsRef.current && debuggerTopRef.current) {
+      panelsRef.current.style.height = (window.innerHeight - debuggerTopRef.current.clientHeight) - debuggerTopRef.current.offsetTop - 7 +'px'
+    }
+  }
+
+  useEffect(() => {
```
ioedeveloper, line 48: It is not advice-able to use a useEffect without a dependency array, it can crash the app. Please provide a dependency array for this `useEffect`.

ioedeveloper (changes requested):

ioedeveloper (approved):

Issue #1729, opened 2021-11-04
Zoom in/Zoom out move to context menu


LianaHus:
We need to remove 2 buttons on the top of editor and add them to context menu

- Zoom In      ___      Ctr +
- Zoom Out    ___     Ctrl -

LianaHus:
cc @ryestew

PR #2697, opened 2022-07-27
added zoomin and zoomout to context menu of editor

LianaHus:
fix https://github.com/ethereum/remix-project/issues/1729


Aniket-Engg:
I see in the issue it is mentioned that  two buttons from editor will be removed, they are still there. Actually I think they too should be there.

Issue #2721, opened 2022-08-03
add e2e for 'calls' url Param


Aniket-Engg:
See: https://github.com/ethereum/remix-project/pull/2592

PR #2731, opened 2022-08-04
add e2e for calls

yann300:
fix https://github.com/ethereum/remix-project/issues/2721

### DIFF
--- a/apps/remix-ide/src/app.js
+++ b/apps/remix-ide/src/app.js
@@ -410,7 +410,7 @@ class AppComponent {
             if (params.call) {
               const callDetails = params.call.split('//')
               if (callDetails.length > 1) {
-                this.appManager.call('notification', 'toast', `initiating ${callDetails[0]} ...`)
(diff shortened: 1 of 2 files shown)


apps/remix-ide/src/app.js
```diff
@@ -410,7 +410,7 @@ class AppComponent {
             if (params.call) {
               const callDetails = params.call.split('//')
               if (callDetails.length > 1) {
-                this.appManager.call('notification', 'toast', `initiating ${callDetails[0]} ...`)
+                this.appManager.call('notification', 'toast', `initiating ${callDetails[0]} and calling "${callDetails[1]}" ...`)
```
bunsenstraat, line 413: so we limit it to two?

yann300: no there can be multiple calls.

yann300: here we are loging the name of the plugina dn the function call

bunsenstraat (approved):

PR #2745, opened 2022-08-09
fix display json rpc error

yann300:
This ensure that while calling the json rpc servic error are displayed correctly .

In order to test it:
 - clone https://github.com/yann300/remix-reward
 - compile the proxy contract
 - move to injected Optimism or Goerli (you should have some test eth)
 - open the script `mint`
 - put any addresses in the variable `to`
 - execute that script
 - terminal should log 2 errors rather than an empty object


Aniket-Engg (approved):
change set looks fine

PR #2752, opened 2022-08-10
fix ipfs publish

yann300:
This ensure routing is done using the hash parameter instead of the path itself.
this fix loading the app from an ipfs gateway, like https://url.gateway/ipfs/-hash-


LianaHus:
why are the indents done so badly in this file? is it our code?

bunsenstraat (approved):

yann300:
> why are the indents done so badly in this file? is it our code?

it wasn't orginally, it's come from MachinaLabs repo

Issue #2720, opened 2022-08-03
issues with vyper plugin


Aniket-Engg:
- [ ] Documentation and `Make an issue` link in plugin info and link on GitHub icon on top still refers to old repo i.e. https://github.com/GrandSchtroumpf/vyper-remix
- [x] When there is no file opened in editor, it shows issue in browser console
```
Error: Error from IDE : Error: No such file or directory No file selected
    at l.<anonymous> (index.js:345)
    at Object.<anonymous> (index.js:561)
    at l.emit (index.js:634)
    at index.js:1099
    at index.js:1011
    at Object.next (index.js:959)
    at i (index.js:912)
```

- [x]  This text `The url to your local compiler` seems unnecessary
- [x] When tried compiling a contract using `Remote compiler`, it shows no message on UI. On icon panel, it shows warning icon and in browser console error is:
```
compiler.tsx:46 POST https://vyper.live/compile 400 (Bad Request)
CompilerButton.tsx:64 Error: Vyper compilation failed: Bad Request
    at compiler.tsx:56
    at s (runtime.js:45)
    at Generator._invoke (runtime.js:264)
    at Generator.e.<computed> [as next] (runtime.js:98)
    at r (asyncToGenerator.js:3)
    at l (asyncToGenerator.js:25)
```

PatrickAlphaC:
Is this plugin supported by [this repo](https://github.com/GrandSchtroumpf/vyper-remix)? All the links are pointing to there, but it looks like this repo supports the vyper plugin?

PatrickAlphaC:
Thanks for this issue, I was just playing around with that repo trying to get stuff to work and then I found that this repo hosts all the code!

PatrickAlphaC:
Additionally, the plugin doesn't support 0.3^ editions of vyper, and the syntax highlighting is off. 


<img width="827" alt="Screen Shot 2022-08-16 at 8 19 36 AM" src="https://user-images.githubusercontent.com/54278053/184877973-b12fcdab-ba04-490d-8041-f4bccc99b0cc.png">

Aniket-Engg:
cc: @yann300

PatrickAlphaC:
Would love to see how we can help with this, it'll be really helpful for making vyper courses.

PR #2755, opened 2022-08-11
Misc Fix Vyper

yann300:
fix https://github.com/ethereum/remix-project/issues/2720

 - fix displaying the title in the status icon


Aniket-Engg (commented):
I see error on no file selected and on remote compiler compilation.
Screenshot 2022-08-16 at 4 42 41 PM

yann300:
The bad request is thrown when the compilation fails, there isn't much to do here.
For the `NoFileSelected`, we explicitly remove the log (it always throw when no file are selected)

apps/vyper/src/app/utils/remix-client.tsx
```diff
@@ -14,34 +15,70 @@ export class RemixClient extends PluginClient {
...
     }
   }
 
+  async cloneVyperRepo() {
+    try {
+      // @ts-ignore
+      this.call('notification', 'toast', 'cloning Vyper repository...')
+      await this.call('manager', 'activatePlugin', 'dGitProvider')
+      // @ts-ignore
+      await this.call('dGitProvider', 'clone', { url: 'https://github.com/vyperlang/vyper', token: null }, 'vyper-lang')
```
Aniket-Engg, line 46: You mentioned `Vyper` at other places but creating workspace with `vyper-lang` name

Aniket-Engg (approved):

apps/vyper/src/app/utils/remix-client.tsx
yann300: that's the issue I was discussing with with @bunsenstraat. we can't have nested folder with same name...

Aniket-Engg: ok

PatrickAlphaC:
Awesome! Looking forward to the release.

PR #2759, opened 2022-08-12
show compiler license

### LINKED ISSUE #247: Display compiler license somewhere (in an about window)
Use `solc.license()` to retrieve the text - it will be **very** long though.

Depends on https://github.com/ethereum/solc-js/pull/121.

Aniket-Engg:
fixes #247

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
+++ b/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
@@ -49,6 +49,7 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
     allversions: [],
     customVersions: [],
     compilerLicense: null,
+    loadedVersion: null,
     selectedVersion: null,
     defaultVersion: 'soljson-v0.8.7+commit.e28d00a7.js', // this default version is defined: in makeMockCompiler (for browser test)
     runs: '',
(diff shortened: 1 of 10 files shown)


libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -704,6 +720,7 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
             <label className="remixui_compilerLabel form-check-label" htmlFor="versionSelector">
               Compiler
               <button className="far fa-plus btn-light border-0 p-0 mx-2 btn-sm" onClick={promptCompiler} title="Add a custom compiler with URL"></button>
+              <button className="fa fa-file-text-o btn-light border-0 p-0 mx-2 btn-sm" onClick={showCompilerLicense} title="See compiler license"></button>
```
LianaHus, line 723: to make those look like icons on FE we would need to:
- put span instead of button (you would need to fix the action)
- remove btn-sm
- rm btn-light
- add ml-3 to the first span to move it away of the label

and please use mr-2 instead of mx-2 on the second icon because now we have 2+2 between them

LianaHus: Screenshot from 2022-08-15 10-43-59
Screenshot from 2022-08-15 10-42-36
Screenshot from 2022-08-15 10-42-13

LianaHus (changes requested):
Please make it look like FE icons (see the comment above)

libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
LianaHus: also I have "could not retreive license" in modal

Aniket-Engg: updated UI

LianaHus:
we had a call and decided to disable the button if the compiler is not loaded.. the tooltip can be updated as well

LianaHus (approved):

PR #2764, opened 2022-08-15
fix listen on transactions

yann300:


libs/remix-lib/src/execution/txListener.ts
```diff
@@ -164,26 +160,33 @@ export class TxListener {
...
-            current++
+      if (this._loopId === null) return
+      if (!lastSeenBlock) {
+        lastSeenBlock = this.executionContext.lastBlock?.number // trying to resynchronize
+        console.log('listen on blocks, resynchronising')
+        return
+      }
+      const current = this.executionContext.lastBlock?.number
+      if (!current) {
+        console.error(new Error('no last block found'))
```
bunsenstraat, line 175: does this mean we would see a console message on each loop when no block is found? Is this not something to show in the terminal itself? as a warning... or something.

yann300: this will be show in the browser terminal. not sure we should put it in the remix terminal, don't want to spam it...

yann300: but you can definetly make it a log or warnng

Aniket-Engg (approved):

PR #2765, opened 2022-08-15
Styling Issues with reserved keywords bug in black theme

### LINKED ISSUE #2661: Styling issues with reserved words, require and bytes, in Black theme
I noticed styling issues with 'require' and 'bytes' when using the Black (dark) theme.

These reserved words turn to black color when the Black theme is active, making them impossible to see unless you hover over them. Maybe there are more words with this problem.

I also noticed that the words public, view, and returns turn dark green and are hard to see.

joeizang:
Fixes #2661

### DIFF
--- a/apps/remix-ide/src/assets/css/themes/remix-black_undtds.css
+++ b/apps/remix-ide/src/assets/css/themes/remix-black_undtds.css
@@ -1,6 +1,6 @@
 @import url('https://fonts.googleapis.com/css?family=Nunito+Sans:400,600&display=swap');
 :root {
-  --blue: #568acf;
+  --blue: #90c3f6;
   --indigo: #6610f2;
   --purple: #9e77f6;
   --pink: #f38abb;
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/assets/css/themes/remix-black_undtds.css
```diff
@@ -1,6 +1,6 @@
 @import url('https://fonts.googleapis.com/css?family=Nunito+Sans:400,600&display=swap');
 :root {
-  --blue: #28282D;
+  --blue: #568acf;
```
LianaHus, line 3: could you please use #90c3f6; instead? the color you picked is too close to primary which is used for other keywords

joeizang: Thank you for pointing this out @LianaHus. I have made the change now.

Aniket-Engg (approved):

ryestew (approved):

Issue #2683, opened 2022-07-25
Double scroll bars in side panel plugins


ryestew:
Learneth has this problem and DGit.  https://github.com/bunsenstraat/remix-storage-plugin/issues/33
The problem is the same ( I believe).  The author of DGit & LearnEth, has asked us for assistance.

PR #2766, opened 2022-08-16
Update to prevent double vertical scrollbars

ryestew:
fix https://github.com/ethereum/remix-project/issues/2683

Test: 
You need the plugin to take up more than the visible height of the browser window - so that a vertical scrollbar appears. When testing :
with internal plugins: check Udapp with a bunch of deployed contracts
with external ones: DGIT & Sourcify (with the verify accordion open).

Check if a double nav occurs.  The problem this solves may only be on Mac OS.
@LianaHus - I switched this from pb-1 to pb-2 to make it work.


bunsenstraat:
it helps but we also now have a horizontal scrollbar on the bottom of the plugins.

LianaHus:
this is because of pb-2 I suppose. can you check with pb-1? I can't reproduce this on ubuntu

ryestew:
@bunsenstraat - which plugins now have the horizontal scroll - all plugins? Internal ones? I've now updated it with pb-1.  Do you still see the horizontal nav?

bunsenstraat (approved):

Issue #2712, opened 2022-08-01
Grid the debugger component instead of stacking vertically.


yann300:
In order to improve the general view, this issue would:
  - when a debug session starts, update the side panel width to at least 50% of the window size.
  - using a React grid to layout the components.

PR #2781, opened 2022-08-18
Grid debugger 2

yann300:
fixes https://github.com/ethereum/remix-project/issues/2712

### DIFF
--- a/apps/remix-ide/src/app/components/side-panel.tsx
+++ b/apps/remix-ide/src/app/components/side-panel.tsx
@@ -12,7 +12,7 @@ const sidePanel = {
   displayName: 'Side Panel',
   description: 'Remix IDE side panel',
   version: packageJson.version,
-  methods: ['addView', 'removeView']
+  methods: ['addView', 'removeView', 'currentFocus']
 }
 
 export class SidePanel extends AbstractPanel {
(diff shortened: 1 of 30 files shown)


apps/remix-ide/src/app/components/side-panel.tsx
```diff
@@ -12,7 +12,7 @@ const sidePanel = {
   displayName: 'Side Panel',
   description: 'Remix IDE side panel',
   version: packageJson.version,
-  methods: ['addView', 'removeView']
+  methods: ['addView', 'removeView', 'currentFocus']
```
bunsenstraat, line 15: is there an implementation for this in this file?

apps/remix-ide/src/app/panels/layout.ts
```diff
@@ -92,4 +102,16 @@ export class Layout extends Plugin {
     this.panels[name].minimized = minimized
     this.event.emit('change', null)
   }
+
+  async maximiseSidePanel () {
+    this.event.emit('maximisesidepanel')
+    const current = await this.call('sidePanel', 'currentFocus')
```
bunsenstraat, line 108: unclear what this does? where is the implementation?

Aniket-Engg: it is defined here: https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/components/panel.ts#L17

bunsenstraat (approved):

PR #2808, opened 2022-08-25
Replace deprecated .enable()

skimaharvey:


frozeman (approved):

frozeman:
This is a needed change, as `.enable()` was never standardised and is in MM only supported for legacy support. All browser extensions support `eth_requestAccounts`

yann300 (approved):

frozeman:
Nice!

PR #2809, opened 2022-08-25
when file change externally, always override with new content

yann300:
fix https://github.com/ethereum/remix-project/issues/2998

When using remixd or remix-desktop, it would warn the user if the file is externally changed.
This has some weird consequences, mostly because the way we track downstream update isn't optimal.

This PR would:
 - when a file is externally changed, it override the remix editor. This is the default behavior of VSCode when the file isn't changed in the editor).
 - try to improve the way downstrream update is being catched.

### DIFF
--- a/apps/remix-ide/src/app/editor/editor.js
+++ b/apps/remix-ide/src/app/editor/editor.js
@@ -13,7 +13,7 @@ const profile = {
   name: 'editor',
   description: 'service - editor',
   version: packageJson.version,
-  methods: ['highlight', 'discardHighlight', 'clearAnnotations', 'addLineText', 'discardLineTexts', 'addAnnotation', 'gotoLine', 'revealRange', 'getCursorPosition', 'open', 'addModel','addErrorMarker', 'clearErrorMarkers'],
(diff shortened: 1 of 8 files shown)


Aniket-Engg:
@yann300 Please fix the conflicts

apps/remix-ide/src/app/editor/editor.js
```diff
@@ -280,11 +280,23 @@ class Editor extends Plugin {
...
-      this.sessions[this.currentFile].setValue(text)
+  setText (url, text) {
+    if (this.sessions[url]) {
+      this.sessions[url].setValue(text)
+    }
+  }
+
+  /**
+   * Get the text in the current session, if any.
+   * @param {string} url Address of the text to replace.
```
Aniket-Engg, line 294: wrong description of param

apps/remix-ide/src/app/editor/editor.js
```diff
@@ -280,11 +280,23 @@ class Editor extends Plugin {
...
+   * @param {string} text New text to be place.
```
Aniket-Engg, line 295: extra param spec

PR #2811, opened 2022-08-26
add timestamp to url

bunsenstraat:
FF doesn't deal well with loading the compiler twice, it basically aborts the first load so nothing loads.


Hbutlercapone (commented):
Approve merge #2811

Aniket-Engg (approved):

PR #2814, opened 2022-08-26
updating profile of remixd

LianaHus:


apps/remix-ide/src/app/plugins/remixd-handle.tsx
```diff
@@ -18,7 +18,11 @@ const profile = {
   events: [],
   description: 'Using Remixd daemon, allow to access file system',
   kind: 'other',
-  version: packageJson.version
+  version: packageJson.version,
+  repo: "https://github.com/ethereum/remixd",
```
bunsenstraat, line 22: this is not maintained, it's archived

LianaHus: What should I put then? @bunsenstraat

bunsenstraat: I guess our monorepo? that's where the code is right.

Aniket-Engg: If specific link can be given, it should be : https://github.com/ethereum/remix-project/tree/master/libs/remixd

Aniket-Engg (approved):

Issue #2796, opened 2022-08-23
Cloning a github repo where name of the repo is same as one of the root folders


yann300:
- in File explorer and clone: https://github.com/bunsenstraat/empty
 - remix freezes
 - reload remix, file explorer never loads.

PR #2818, opened 2022-08-27
Maintain standard for root directory

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/2796


bunsenstraat:
the Filemanager refresh command doesn't work as expected anymore. that's why the git clone test fails @ioedeveloper 
but it clones it correctly.

bunsenstraat (approved):

PR #2819, opened 2022-08-28
Fix drag and drop

ioedeveloper:


yann300:
- this doesn't work with Brave.
 - it's also not possible to drag n drop multuple files.

ioedeveloper:
> * this doesn't work with Brave.

This is a problem from brave that has been fixed. Try updating your browser.

yann300 (approved):

PR #2822, opened 2022-08-29
Fix monaco diagnostic

yann300:
- add hardhat extension & mocha & chai
 - fix showing typescript error for existing dependencies


Aniket-Engg (approved):
LFTM!

PR #2825, opened 2022-08-30
side panel paddings

LianaHus:


libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -188,11 +188,11 @@ export function Workspace () {
   }
 
   return (
-    <div className='px-2 remixui_container'>
+    <div className='emixui_container'>
```
Aniket-Engg, line 191: looks like a typo in class name

Aniket-Engg (approved):

PR #2829, opened 2022-08-30
Vyperstyle

LianaHus:


apps/vyper/src/app/components/VyperResult.tsx
```diff
@@ -27,18 +26,20 @@ function VyperResult({ output }: VyperResultProps) {
...
-      </Button>
+      <p className="my-3">No contract compiled yet.</p>
+      
     </div>
   )
 
   if (isCompilationError(output)) {
     return (
-    <div id="result" className="error">
+    <div id="result" className="error" title={output.message}>
```
yann300, line 36: maybe it would have been bettere to use the react tool tip that we've added.

LianaHus: it is not possible to use, it removes the prestyle and needs more space than iframe has.

ryestew (approved):

Issue #2844, opened 2022-09-01
Click CTRL/CMD key moves cursor to definition in Editor


Aniket-Engg:
For example: using CTRL  + C, it moves to different line and copy that. This is getting quite inconvenient.

cc: @bunsenstraat

PR #2850, opened 2022-09-01
fix definition provider

bunsenstraat:
fixes: https://github.com/ethereum/remix-project/issues/2844
jump to definition 
peek definition
reveals range of files opened by double clicking in peeks

### DIFF
--- a/libs/remix-ui/editor/src/lib/providers/definitionProvider.ts
+++ b/libs/remix-ui/editor/src/lib/providers/definitionProvider.ts
@@ -39,10 +39,16 @@ export class RemixDefinitionProvider implements monaco.languages.DefinitionProvi
     */
     async jumpToPosition(position: any) {
         const jumpToLine = async (fileName: string, lineColumn: any) => {
+            const fileTarget = await this.props.plugin.call('fileManager', 'getPathFromUrl', fileName)
             if (fileName !== await this.props.plugin.call('fileManager', 'file')) {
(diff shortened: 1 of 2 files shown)


libs/remix-ui/editor/src/lib/providers/definitionProvider.ts
```diff
@@ -30,20 +39,33 @@ export class RemixDefinitionProvider implements monaco.languages.DefinitionProvi
     */
     async jumpToPosition(position: any) {
         const jumpToLine = async (fileName: string, lineColumn: any) => {
+            const fileTarget = await this.props.plugin.call('fileManager', 'getPathFromUrl', fileName)
             if (fileName !== await this.props.plugin.call('fileManager', 'file')) {
                 await this.props.plugin.call('contentImport', 'resolveAndSave', fileName, null)
-                await this.props.plugin.call('fileManager', 'open', fileName)
+                const fileContent = await this.props.plugin.call('fileManager', 'readFile', fileName)
+                try {
+                    await this.props.plugin.call('editor', 'addModel', fileTarget.file, fileContent)
```
yann300, line 47: you do that for making sure monaco editor can actually display the file?
could you add a comment for describing what this is about?

bunsenstraat: it's for peeking. 
we capture the open file action from the editor but that doesn't happen when it's being peeked at. so we need to model, just like I do in the referenceProvider.

yann300 (approved):

PR #2879, opened 2022-09-06
Switch and Checkout Branch

ioedeveloper:


yann300:
This probably needs some e2e tests

ryestew (approved):

ryestew:
I like the GIT stuff at the bottom of the FE.  And the widget for the branch selection is cute.  It works with branch names that are pretty long.  Ahhh having just approved - I see a slight problem - when making a new branch name that has a bit of length - it can make the width wider that the iframe like this:
image

PR #2897, opened 2022-09-08
Undo file actions

iamsethsamuel:
Last actions on files such as moving, rename, creating and deleting a file can now be undone with Ctrl+Z


apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -861,7 +885,7 @@ class FileManager extends Plugin {
    * @returns {void}
    */
   
-  async moveDir(src: string, dest: string) {
+   async moveDir(src: string, dest: string) {
```
LianaHus, line 888: no need for space here

apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -877,11 +901,93 @@ class FileManager extends Plugin {
...
+  /**
+   * Reverts the last action that was performed by the user
+   * @returns {void}
+  */
+
+  async revertFileAction(redo: boolean){
+    this.isFromLastAction = true
+
+    let lastAction = this.actions[this.actions.length - 1]
+    if(redo){
```
LianaHus, line 935: space needed after if. also below

apps/remix-ide/src/app/files/fileManager.ts
iamsethsamuel: I have removed it

apps/remix-ide/src/app/files/fileManager.ts
iamsethsamuel: I have added it

apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -396,14 +411,21 @@ class FileManager extends Plugin {
       path = this.limitPluginScope(path)
       await this._handleExists(path, `Cannot remove file or directory ${path}`)
       const provider = this.fileProviderOf(path)
+
+      if(await this.isDirectory(path)){
```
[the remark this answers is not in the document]
iamsethsamuel: I have done that

iamsethsamuel:
I have fixed the merge conflict

iamsethsamuel:
Please can someone merge this so that I can send the next PR?

bunsenstraat:
> Please can someone merge this so that I can send the next PR?

Yes, but we need to discuss some things about it.

bunsenstraat:
Two things
- when I switch workspace it should remove the recorded actions otherwise you'd be undoing it in another workspace
- the undo action should be available globally: so ie you are doing a file rename, and then click somewhere else it should still undo, even when the cursor is in the editor. so these things should be linked if we can do that... might be hard. or we have another way except ctrl-z. Macs btw don't use CTRL. 
- the undo actions should have a modal ( I think ) telling you you're undoing something.

iamsethsamuel:
> 

Okay, I modelled it after VSCode's implementation. Which I think most people are used to so if the behaviour is different here I guess it will create a confusion. I think the file explorer in windows and Linux also require focus to work. So imagine I working on the editor and want to undo my code and I undo a rename. That will be weird.

As for the dialog, you can always redo so I don't see the need for it.

The workspace is very important, let me see if I can implement it.

And I don't have a Mac so I don't know how things work there. If someone can help implement it that will be awesome.

Just my thoughts, but we can do it your way

bunsenstraat:
I think it would be good to have a call with the team and discuss how we could do it. Would that fit for you?

bunsenstraat:
Thank you for making this effort btw!! It's a great idea.

iamsethsamuel:
> I think it would be good to have a call with the team and discuss how we could do it. Would that fit for you?

Yes, when will you be free for the call

iamsethsamuel:
Please can someone review this?

ioedeveloper (approved):

Issue #2886, opened 2022-09-07
easier way to run script.


yann300:
There should probably be a "run" action somewhere that appear when the current file is typescript or javascript.

PR #2901, opened 2022-09-12
play btn to run scripts

LianaHus:
fix [#2886](https://github.com/ethereum/remix-project/issues/2886)

### DIFF
--- a/libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
+++ b/libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
@@ -24,15 +24,18 @@ export interface TabsUIApi {
 interface ITabsState {
   selectedIndex: number,
   fileDecorations: fileDecoration[],
+  currentExt: string
 }
 interface ITabsAction {
   type: string,
   payload: any,
+  ext?: string,
 }
 
 const initialTabsState: ITabsState = {
   selectedIndex: -1,
   fileDecorations: [],
+  currentExt: ''
 }
 
 const tabsReducer = (state: ITabsState, action: ITabsAction) => {
@@ -40,6 +43,7 @@ const tabsReducer = (state: ITabsState, action: ITabsAction) => {
     case 'SELECT_INDEX':
       return {
         ...state,
+        currentExt: action.ext,
         selectedIndex: action.payload,
       }
     case 'SET_FILE_DECORATIONS':
@@ -54,7 +58,6 @@ const tabsReducer = (state: ITabsState, action: ITabsAction) => {
 
 export const TabsUI = (props: TabsUIProps) => {
   const [tabsState, dispatch] = useReducer(tabsReducer, initialTabsState);
-  const [currentFileExt, setCurrentFileExt] = useState<string>('sol')
   const currentIndexRef = useRef(-1)
   const tabsRef = useRef({})
   const tabsElement = useRef(null)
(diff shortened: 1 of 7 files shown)


libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
```diff
@@ -135,10 +132,36 @@ export const TabsUI = (props: TabsUIProps) => {
...
+
   return (
     <div className="remix-ui-tabs d-flex justify-content-between border-0 header nav-tabs" data-id="tabs-component">
       <div className="d-flex flex-row" style={{ maxWidth: 'fit-content', width: '97%' }}>
-        <div className="d-flex flex-row justify-content-center align-items-center m-1 mt-2">
+        <div className="d-flex flex-row justify-content-center align-items-center m-1 mt-1">
+          <OverlayTrigger placement="bottom" overlay={
+            <Tooltip id="overlay-tooltip-run-script">
+              <span>
+                Run script
```
yann300, line 149: The message should also be displayed when it's disabled.
like i.e `Select a ts or js file and run it` and also with the key binding which is `CTRL + SHIFT + S`

libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
```diff
@@ -135,10 +132,36 @@ export const TabsUI = (props: TabsUIProps) => {
...
+              </span>
+            </Tooltip>
+          }>
+            <button
+              className="btn text-success py-0"
+              disabled={!(currentFileExt === 'js' || currentFileExt === 'ts')}
```
yann300, line 155: you don't compile if a `sol` file is selected? that maybe be better actuallly...

libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
LianaHus: well that is not working with this tooltip. I'll try to find a solution.

libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
LianaHus: to compile? I though to do it in a separate PR since it is not included in the issue. but I can do it

yann300:
I see the toast `tab is requiring to compile...` ...

libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
```diff
@@ -135,10 +133,41 @@ export const TabsUI = (props: TabsUIProps) => {
...
+        <div className="d-flex flex-row justify-content-center align-items-center m-1 mt-1">
+          <button
+            className="btn text-success py-0"
+            disabled={!(tabsState.currentExt === 'js' || tabsState.currentExt === 'ts' || tabsState.currentExt === 'sol')}
+            onClick={async () => {
+              const path = active().substr(active().indexOf('/') + 1, active().length)
+              const content = await props.plugin.call('fileManager', "readFile", path)
+              if (tabsState.currentExt === 'js' || tabsState.currentExt === 'ts') {
+                await props.plugin.call('scriptRunner', 'execute', content)
+              } else if (tabsState.currentExt === 'sol') {
```
yann300, line 155: should also works with the `yul` extension.

yann300 (approved):

libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
LianaHus: added

PR #2908, opened 2022-09-13
Parserfix

bunsenstraat:
some improvements have been made to the parser 
- there are two modes now, 
   - parse - which is used to parse snippets you're typing, that is done without a worker
   - parseBlock - a fast custom grammar to only extract contracts, functions & constructors to determine the block the cursor is in. Done without a worker
   - the worker is also parsing every 5 secs for:
      - testing if the file can be parsed fast enough for throttling
      - to build a node cache ast
- throttling:
When the parser gets slow ( after 2 times parsing ) it disables the parsing of blocks until it encounters in the worker that it's ok again and it will be turned on again. It does this after 2 times because the parser caches internally too, so the timing info after 2 is better.
If a file has a lot nodes with a similar name monaco can't render the list quick enough so we need to cap it, it's set at 100
- fixed some bugs

TODO for a later PR: put the parser code in the monorepo as a lib


apps/remix-ide/src/app/plugins/parser/code-parser.tsx
```diff
@@ -124,6 +124,10 @@ export class CodeParser extends Plugin {
 
         this.on('fileManager', 'currentFileChanged', async () => {
             await this.call('editor', 'discardLineTexts')
+            const completionSettings = await this.call('config', 'getAppParameter', 'auto-completion')
+            if (completionSettings) {
+                this.antlrService.getCurrentFileAST()
```
yann300, line 129: this doesn't return anything?

yann300:
just saw this toast: `This file is big so some autocomplete features will be disabled.` for the Storage.sol sample contract.

yann300:
remix
the warning is: `Warning: Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. `
it spams the code editor, it'd maybe be better to not underline the code for this specific issue.

apps/remix-ide/src/app/plugins/parser/code-parser.tsx
yann300, line 129: if it's supposed to return something anymore, renaming the function would be better

apps/remix-ide/src/app/plugins/parser/services/antlr-worker.ts
```diff
@@ -0,0 +1,54 @@
...
+                            timestamp: data.timestamp,
+                            ast,
+                            text: data.text,
+                            file: data.file,
+                            duration: endTime - startTime,
+                            blockDuration,
+                            blocks
+                        })
+                    } catch (e) {
+                        // do nothing
```
yann300, line 47: sometimes it's handy to log to the browser console. it helps remote debugging while doing community support

apps/remix-ide/src/app/plugins/parser/code-parser.tsx
bunsenstraat: no

bunsenstraat:
> remix the warning is: `Warning: Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. ` it spams the code editor, it'd maybe be better to not underline the code for this specific issue.

If error/warnings have no specific codes but are just strings this is hard to do. I could have a hide warnings option...

apps/remix-ide/src/app/plugins/parser/code-parser.tsx
bunsenstraat: I renamed it

yann300:
> If error/warnings have no specific codes but are just strings this is hard to do. I could have a hide warnings option...

that's fine.. but enabling optimization in the sol compiler would in most cases remove this warning. I still see the warning after enabling it.

yann300:
working with `UniswapV3Factory` and I didn't notice any slow-down

bunsenstraat:
> > If error/warnings have no specific codes but are just strings this is hard to do. I could have a hide warnings option...
> 
> that's fine.. but enabling optimization in the sol compiler would in most cases remove this warning. I still see the warning after enabling it.
> 
> is there an event for settings changing in the compiler? @yann300 ... then I could react to it

This is fixed by adding optimizer from the settings

apps/remix-ide/src/app/plugins/parser/services/code-parser-antlr-service.ts
```diff
@@ -27,18 +118,31 @@ export default class CodeParserAntlrService {
...
-            try {
-                const ast = await this.parseSolidity(fileContent)
-                this.plugin.antlrParserResult = ast
-            } catch (e) {
-                // do nothing
+    async setCurrentFileAST(text: string | null = null) {
+        try {
+            this.plugin.currentFile = await this.plugin.call('fileManager', 'file')
+            if (this.plugin.currentFile && this.plugin.currentFile.endsWith('.sol')) {
+                if (!this.plugin.currentFile) return
```
Aniket-Engg, line 125: parent `if` already check if `this.plugin.currentFile` exists

apps/remix-ide/src/app/plugins/parser/services/code-parser-antlr-service.ts
```diff
@@ -27,18 +118,31 @@ export default class CodeParserAntlrService {
...
+                        parsingEnabled: true
+                    }
+                }
+                if (this.cache[this.plugin.currentFile] && this.cache[this.plugin.currentFile].text !== fileContent) {
+                    try {
+                        await this.parseWithWorker(fileContent, this.plugin.currentFile)
+                    } catch (e) {
+                        // do nothing
+                    }
+                } else {
```
Aniket-Engg, line 140: not sure if this `else` can be removed

apps/remix-ide/src/app/plugins/parser/services/code-parser-antlr-service.ts
```diff
@@ -127,6 +232,30 @@ export default class CodeParserAntlrService {
...
+    */
+    async getCurrentFileBlocks(text: string | null = null) {
+        this.plugin.currentFile = await this.plugin.call('fileManager', 'file')
+        if (this.cache[this.plugin.currentFile]) {
+            if (!this.cache[this.plugin.currentFile].parsingEnabled) {
+                return
+            }
+        }
+        if (this.plugin.currentFile && this.plugin.currentFile.endsWith('.sol')) {
+            if (!this.plugin.currentFile) return
```
Aniket-Engg, line 246: same here

apps/remix-ide/src/app/plugins/parser/services/code-parser-antlr-service.ts
```diff
@@ -127,6 +232,30 @@ export default class CodeParserAntlrService {
...
+            const fileContent = text || await this.plugin.call('fileManager', 'readFile', this.plugin.currentFile)
+            try {
+                const startTime = Date.now()
+                const blocks = (SolidityParser as any).parseBlock(fileContent, { loc: true, range: true, tolerant: true })
+                this.setFileParsingState(this.plugin.currentFile, Date.now() - startTime)
+                if(blocks) this.cache[this.plugin.currentFile].blocks = blocks
```
Aniket-Engg, line 252: add a space between `if` and `(blocks)`

Aniket-Engg (approved):

PR #2915, opened 2022-09-14
add code format

bunsenstraat:
Adds code formatting 
for ts, sol, js & json

It can be use by SHIFT ALT F
or right click and use the context menu

Not exactly sure about tweaking the formatters ... need feedback

ps: the plugin for prettier could not be used completely because it required the antlr parser which is already present because it is being used by autocomplete. that caused conflicts because it loads itself in the incorrect way.


Aniket-Engg (approved):
LGTM!

apps/remix-ide/src/app/plugins/code-format.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export class CodeFormat extends Plugin {
+
+    constructor() {
+        super(profile)
+    }
+
+    async format() {
+        const file = await this.call('fileManager', 'getCurrentFile')
```
yann300, line 26: I feel `format` should accept a `path` param, instead of getting the current file, if for instance one want to prettier entire project, or just prettier multiple files.

apps/remix-ide/src/app/plugins/code-format.ts
```diff
@@ -0,0 +1,77 @@
...
+        const content = await this.call('fileManager', 'readFile', file)
+        let parserName = ''
+        let options: Options = {
+        }
+        switch (path.extname(file)) {
```
yann300, line 31: at best we should have configuration files (for sol, js, ts).
https://github.com/prettier-solidity/prettier-plugin-solidity#configuration-file
but that can certainly be done in an other iteration.

apps/remix-ide/src/app/plugins/code-format/parser.ts
```diff
@@ -0,0 +1,197 @@
...
+  if (node.type === 'BinaryOperation' && operators.includes(node.operator))
+    return {
+      type: 'TupleExpression',
+      components: [node],
+      isArray: false
+    };
+  return node;
+};
+
+export function parse(text, _parsers, options) {
```
yann300, line 17: this isn't running in a worker, is it still ok for now?

yann300:
We should add the new key binding to the list of key bindings that are shown when no file is selected.

apps/remix-ide/src/app/plugins/code-format.ts
bunsenstraat: sure

yann300 (approved):

PR #2917, opened 2022-09-14
add imports to auto-complete

bunsenstraat:
turn on autocomplete in the settings!
type import " and then you will get a list of local imports or packages
select a package ie openzeppelin en type "/" it will autocomplete 
you can either navigate by folder by selecting and typing '/' or just type anything and find the right file

also: improved error handling on imports. you will now see a squiggly line in the imports if the imported file contains an error. also improved jumping to imports. 
sometimes the compiler returns just a single error if the file is not found. also handling that.


Aniket-Engg (commented):
This works for me. It can be good to have options for contracts in workspace `contracts` directory

yann300 (approved):

PR #2936, opened 2022-09-22
Script to generate changelog

yann300:


yann300:
@Aniket-Engg the issue is with https://github.com/ethereum/remix-project/pull/2818 , which is present in my list but not in the `is:pr` view... it is part of the `done` column though..

Aniket-Engg:
> @Aniket-Engg the issue is with #2818 , which is present in my list but not in the `is:pr` view... it is part of the `done` column though..

Ok

Issue #2930, opened 2022-09-21
[bug] "Deployed Contracts" does not display contract instance correctly


zoey-t:
I found this weird bug.
Process to reproduce this bug:
1. deploy two difference contracts A and B
  
![image](https://user-images.githubusercontent.com/101599447/191574362-ac25256a-0c27-49fb-b717-88a60568f002.png)
2. then I delete the first contract A. contract B just become like this. The functions and variables are all become A's. If i click on these buttons, it will actually call these functions but with B's address. So I will just get errors in the console because those functions do not exist in B
 
![image](https://user-images.githubusercontent.com/101599447/191574507-0011aaf1-9a8f-48fa-8b4f-51c4d3f09ebd.png)
![image](https://user-images.githubusercontent.com/101599447/191574911-f6522776-27ce-4acb-96b6-f8a670c457ef.png)

PR #2937, opened 2022-09-22
fix issue 2930

drafish:
fix issue #2930  @Aniket-Engg @yann300

### DIFF
--- a/libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
+++ b/libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
@@ -24,7 +24,7 @@ export function InstanceContainerUI (props: InstanceContainerProps) {
       { instanceList.length > 0
         ? <div> { props.instances.instanceList.map((instance, index) => {
           return <UniversalDappUI
-            key={index}
+            key={instance.address}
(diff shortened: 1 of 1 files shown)


Aniket-Engg:
Thanks for the fix @drafish . Can you please add an explanation to the changes in this PR?

drafish:
This is caused by using index as key of the list. @zoey-t deployed contract A and contract B successively. The index of contract A is 0, and the index of contract B is 1. When he deleted contract A, the index of contract B became 0. However, react use key to identify items, it will think that 1 is the item that should be deleted. This is why contract B was deleted. 

React officially does not recommend using index as key. Key should be unique and stable, but index can only be unique, not stable. @Aniket-Engg You can see more here https://reactjs.org/docs/lists-and-keys.html#keys

I'v searched the whole project and I found 57 more results which use index as key. I will be happy to fix that. But I'm not sure, should I commit in this pr, or I should start an other pr.

Aniket-Engg:
Thanks for the explanation @drafish . I appreciate your efforts.

To fix it on other places, I would say it will be better to create an issue and open small PRs for tracking

Aniket-Engg (approved):

PR #2964, opened 2022-09-29
New home tab

LianaHus:
https://www.figma.com/file/TlNhECwnBowBvP6MU1odrS/Untitled?node-id=1%3A349


yann300:
some feedback:
 - the remix icon is probably to small
 - there are a lot of border
 - feature plugins and get started needs to be bigger I think
 - in "Files" maybe we can have a "Create new workspace"
 - "Connect to localhost" why no "Connect to Local files"
 - Features plugin and Get Started should probably be scrollable horizontaly
 - "Remix rewards" needs a link to the rewards UI
 - there are a lof of "margin top" when the terminal is minimized, this seems off

Aniket-Engg:
> "Connect to localhost" why no "Connect to Local files

Better to `Connect Local Project`

joeizang:
Marvellously done @LianaHus. Just some things I noticed.

image
+ The getting started and featured plugins section keeps scrolling when it gets to the end. I think it should stop when it gets to the last item.
+ The Remix Rewards UI image should have margins all round so the top and bottom don't touch the containing element.
+ Learn section headings should have heading text as white, indicating its active (I think)

vermouth22:
Hi @LianaHus I noticed in Mac OS Brave 1.42.97 "Remix" is rendering as Times Roman. Also, general comment: I wonder if the "New File" and "Open File" can be side-by-side instead of stacked, so that the LearnETh tutorial button can be seen above the fold without scrolling (on a 13 inch laptop screen)...
image

vermouth22:
And in Mac FF the images in the features carousel are displaying with distorted aspect ratio...
image

Aniket-Engg (commented):
Functionality wise, it looks fine. Only `Add a workspace` not showing modal to create workspace but creates a blank workspace

yann300:
- in the Figma titles like `Featured`, `Get Started` are bright, in the PR the font color is a bit different and  doesn't help to differentiate section.
- the "Remix" title is all uppercase in the design.
- the `get started` from the learn section doesn't look like it's using bootstrap or it's different from the design (see it with the black theme)
- in the pr, the remix rewards image are cropped in the bottom.
- the arrows to slide in the "get started" and "featured plugin", doesn't look centered.

yann300:
-  in the design, the twitter, youtube link, ... doesn't have border

yann300:
image cropped:
remix

PR #2977, opened 2022-10-03
handle connections in plugins

### LINKED ISSUE #2976: Clicking on Refresh icon shows ws errors in browser console

bunsenstraat:
fixes #2976 

before the plugin call happens check the connection. no use to SEND to the socket if it's not open. prevents timeouts and removes the need to throw errors in the console.

### DIFF
--- a/apps/remix-ide/src/app/files/foundry-handle.js
+++ b/apps/remix-ide/src/app/files/foundry-handle.js
@@ -15,4 +15,12 @@ export class FoundryHandle extends WebsocketPlugin {
   constructor () {
     super(profile)
   }
+
+  callPluginMethod(key, payload = []) {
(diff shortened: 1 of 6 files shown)


apps/remix-ide/src/app/files/foundry-handle.js
```diff
@@ -15,4 +15,12 @@ export class FoundryHandle extends WebsocketPlugin {
   constructor () {
     super(profile)
   }
+
```
yann300, line 18: might need to do the same for all the other WebsocketPlugin plugin (slither, remixd, ...)

bunsenstraat: ok

Aniket-Engg (approved):

PR #2983, opened 2022-10-04
Workspace dropdown menu

joeizang:


libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -292,91 +293,181 @@ export function Workspace () {
...
+        id='workspaceRename'
+        data-id='workspaceRename'
+        onClick={(e) => {
+          e.stopPropagation()
+          renameCurrentWorkspace()
+          _paq.push(['trackEvent', 'fileExplorer', 'workspaceMenu', 'workspaceRename'])
+          hideIconsMenu(!showIconsMenu)
+        }}
+        className='far fa-edit w-100'>
+          {'  Rename'}
```
LianaHus, line 316: IMHO it's better to style with css, For that please separate icon and the label into 2 separate spans and fix the spacing between them.

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -292,91 +293,181 @@ export function Workspace () {
...
+                    className='far fa-plus-square remixui_menuicon'
+                    >
+                  </span>
+                  </OverlayTrigger>
+                  <Dropdown id="workspacesMenuDropdown" data-id="workspacesMenuDropdown" onToggle={() => hideIconsMenu(!showIconsMenu)} show={showIconsMenu}>
+                    <Dropdown.Toggle
+                      as={CustomIconsToggle}
+                      mouseOverAction={(e: SyntheticEvent) => {
+                        hideIconsMenu(!showIconsMenu)
+                        console.log({ e })
```
LianaHus, line 452: do we need this?

LianaHus:
I wonder if click should also work on the menu. hover is fine, but i always try to click, and it doesn't work. Do you think we can have both?

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
joeizang: Done. Used pl-2 rule for separation

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
joeizang: No we don't. Left in error.

yann300:
- the icon doesn't seem well placed and it isn't big enough. can we make all the icons a bit bigger?
 - the "create" should also be in the dropdown imo.
 - the dropdown view itself, I think it needs some more design. Something seems off but I don't know how to point out it...

joeizang:
> I wonder if click should also work on the menu. hover is fine, but i always try to click, and it doesn't work. Do you think we can have both?

I will try to have both and see how it feels

yann300:
- I would be even fine to remove the display on hover and keep the onClick @LianaHus

joeizang:
> * I would be even fine to remove the display on hover and keep the onClick @LianaHus

I have moved this away from on hover to click now.

yann300:
> the dropdown view itself, I think it needs some more design. Something seems off but I don't know how to point out it...

could you try:
 - add separation between "delete" and "download" and "restore" and "close"
 - make **all** the icons a bit bigger.
 - the onhover and blue background doesn't look.
 - the "plus" and "menu" are apart which is weird..

ioedeveloper:
I agree with @yann300's comments. I also think tooltips background needs to change.

yann300:
for the "clone" it should be the github icon

yann300:
`clone` should be in a specific question

yann300:
- The background when hovering a file is not right, it should be the same as for the file explorer.
 - open the dropdown, click on a file in the file explorer, this should hide the dropdown.

yann300:
- can you increase the size of the Menu icon (settings `large` for font-size looks ok), and adapt with the `+` icon and the label "workspace"
 cc @ioedeveloper

joeizang:
> `clone` should be in a specific question

I am sorry @yann300 I don't understand this comment, can you elaborate?

yann300:
also "Workspace" should be all uppercase

joeizang:
> also "Workspace" should be all uppercase

Done. Thanks @yann300

Aniket-Engg:
Screenshot 2022-10-14 at 7 02 40 PM

here too

Issue #2616, opened 2022-07-12
Upgrade with Proxy error reporting


ryestew:
When using
<img width="297" alt="image" src="https://user-images.githubusercontent.com/6050001/178507295-d431282c-483e-413d-b3d1-fe86cb4fcbe5.png">

or
<img width="293" alt="image" src="https://user-images.githubusercontent.com/6050001/178507342-63389179-e456-4546-aaef-d85accad1b9e.png">

and the address is not that of the proxy 1967 contract, there is no error reported in Remix.

Aniket-Engg:
@ioedeveloper do you think this can be done?

PR #2984, opened 2022-10-04
Prevent proxy upgrade for invalid proxy address

ioedeveloper:
Closes https://github.com/ethereum/remix-project/issues/2616


Aniket-Engg (commented):
It doesn't seem working to me. If I put any address, upgrade proxy works.

ioedeveloper:
> It doesn't seem working to me. If I put any address, upgrade proxy works.

Any valid address will work, but an invalid input like name or number will not work.

Aniket-Engg (approved):

Issue #2979, opened 2022-10-04
UI of Open Zeppelin templates


ryestew:
When creating a new open zeppelin template - the radio buttons for Upgradeability, do not include an option for None.
<img width="528" alt="image" src="https://user-images.githubusercontent.com/6050001/193713807-2771d936-f426-4de1-afb9-e49731649ee7.png">

As a result, if you happen to check one of the upgradeable options, then you can't uncheck one.  There should be an option for **None** and this should be the default.

Aniket-Engg:
this UI has been refactored. Closing it.

PR #2986, opened 2022-10-05
Remove oz transparent from template modal

ryestew:
Also makes the uups proxy a checkbox and not a radio button.
fix https://github.com/ethereum/remix-project/issues/2979


ryestew:
Closed -this will be addressed in another PR that keeps the transparent pattern visible on the modal.

PR #3005, opened 2022-10-07
Custom tooltips for Static Analysis

### LINKED ISSUE #2432: Make all tooltips custom
For improved experience & consistency, make all Custom tool tips - instead of browser

joeizang:
fixes #2432


ioedeveloper (commented):
Instead of repeating imports for `OverlayTrigger` and `Tooltip` and introducing new blocks of code. Wouldn't it be better to have a custom component that wraps `children` around overlay and tooltip? That way, it can be easily reused easily.

Aniket-Engg (commented):
For `Run` button, tooltip is only visible if you hover on the text but not on the blue area around the text .

joeizang:
> For `Run` button, tooltip is only visible if you hover on the text but not on the blue area around the text .

Limitation of elements that are disabled to begin with. Tooltip workaround causes e2es to fail. @LianaHus recommended adding tooltip to text only for now.

Aniket-Engg (approved):

PR #3031, opened 2022-10-19
Update Injected Provider option to detect more browser wallets

taycaldwell:
### Background

The `Injected Provider` option displays `Injected Provider - MetaMask` if the `isMetaMask` field on the top level injected provider is `true`.

If `isMetaMask` is `false`, the option will just be listed as `Injected Provider`.

Screen Shot 2022-10-19 at 2 38 20 PM

#### Non-metamask wallets

Other wallets also follow this convention. For example:

- The Brave Wallet provider has `isBraveWallet`
- The Coinbase Wallet provider has `isCoinbaseWallet`

We can use these fields to detect other browser extensions as well.

### Problem

Currently, it isn't immediately obvious a user can use their browser wallet in their environment, unless they are using MetaMask or are familiar with the concept of an Injected Provider.

More-so, some wallets like Brave Wallet still set the `isMetaMask` field, in addition to their own field (i.e. `isBraveWallet`), leading to developer confusion, as Remix will display `Injected Provider - MetaMask` for Brave Wallet, even when MetaMask is not installed.

#### Multi-extension users

Some users have multiple browser extensions installed.

In this case, a specific wallet (such as MetaMask) should not be displayed over the other options. 

For example, if a user has Coinbase Wallet and MetaMask installed, Remix IDE should not display `Injected Provider - MetaMask`. Instead, generic language should be used (i.e.`Injected Provider`, `Browser wallet`, etc.).

### Solution

Remix IDE will detect which wallet browser extension the user is using. If detected, Remix will display the name of the wallet.

#### Detected wallets

- MetaMask
- Brave Wallet
- Coinbase Wallet

If no wallet is detected, Remix will fallback to simply displaying generic, wallet-agnostic language: `Injected Provider`.

*Note: The list of detected wallets can be extended to support others.*

#### MetaMask

Screen Shot 2022-10-19 at 2 51 04 PM

#### Brave Wallet

Screen Shot 2022-10-19 at 2 50 13 PM

#### Coinbase Wallet

Screen Shot 2022-10-19 at 2 52 48 PM

#### Multi-extension installed or wallet extension not detected

Screen Shot 2022-10-19 at 2 55 15 PM

### DIFF
--- a/libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
+++ b/libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
@@ -138,7 +138,11 @@ export const runTabInitialState: RunTabState = {
       dataId: 'settingsInjectedMode',
       title: 'Execution environment has been provided by Metamask or similar provider.',
       value: 'injected',
-      content: `Injected Provider${(window && window.ethereum && window.ethereum.isMetaMask) ? ' - Metamask' : ''}`
+      content: `${(window && window.ethereum && !(window.ethereum.providers && !window.ethereum.selectedProvider)) ?
+        window.ethereum.isCoinbaseWallet || window.ethereum.selectedProvider?.isCoinbaseWallet ? 'Coinbase Wallet' :
+        window.ethereum.isBraveWallet || window.ethereum.selectedProvider?.isBraveWallet ? 'Brave Wallet' :
+        window.ethereum.isMetaMask || window.ethereum.selectedProvider?.isMetaMask ? 'MetaMask' :
+        'Injected Provider' : 'Injected Provider'}`
     }],
     isRequesting: false,
     isSuccessful: false,
--- a/libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
+++ b/libs/remix-ui/run-tab/src/lib/reducers/runTab.ts
@@ -138,11 +138,10 @@ export const runTabInitialState: RunTabState = {
       dataId: 'settingsInjectedMode',
       title: 'Execution environment has been provided by Metamask or similar provider.',
       value: 'injected',
-      content: `${(window && window.ethereum && !(window.ethereum.providers && !window.ethereum.selectedProvider)) ?
-        window.ethereum.isCoinbaseWallet || window.ethereum.selectedProvider?.isCoinbaseWallet ? 'Coinbase Wallet' :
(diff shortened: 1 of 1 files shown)


yann300:
Thanks for the PR, is this still possible to display it like `injected Provider - Brave` or `injected - Brave`

taycaldwell:
Updated to keep the same naming convention as before: `Injected Provider - <wallet>`

taycaldwell:
Confirmed working on Coinbase Wallet using the Vercel link. Tested deploying a contract to Goerli.

Screenshot 2022-10-26 at 6 01 08 PM

Screenshot 2022-10-26 at 6 01 29 PM

Note: If multiple extensions are installed at once, Remix will fallback to just displaying `Injected Provider`.

taycaldwell:
@yann300 Please check out the latest and let me know if you are still experiencing any issues.

yann300 (approved):

vermouth22:
On testing this, I found that if Brave Ethereum Default Wallet Settings are set to "Brave Wallet" instead of "Brave Wallet - (Prefer extensions), Remix will not be able to switch to Metamask if user decides to use Metamask, even if Brave Wallet is not connected and Metamask is connected.
image

vermouth22:
Really the only condition under which "Injected Provider - Brave" will show in pulldown and Brave Wallet adress will actually display in FE is if "Brave Wallet" is selected in "Default Ethereum wallet" in Brave Wallet Settings. All other Settings of Default Ethereum Wallet seem to only allow display of Metamask wallet address in FE.

yann300:
yes it's seem the Brave setting doesn't handle at all the fact that metamask is installed which is unfortunate... I think this is an issue that all the Dapp that use web3 extension will have.
@taycaldwell 

```
Multi-extension users
Some users have multiple browser extensions installed.

In this case, a specific wallet (such as MetaMask) should not be displayed over the other options.

For example, if a user has Coinbase Wallet and MetaMask installed, Remix IDE should not display Injected Provider - MetaMask. Instead, generic language should be used (i.e.Injected Provider, Browser wallet, etc.).
```
It doesn't seem the PR is handling, right?
I had metamask installed and Brave wallet selected in the settings, the dropdown is still showing `Metamask`.

yann300:
> Some users have multiple browser extensions installed.

seems at the end it's more a browser issue than something that should be impl in a Dapp.. 
Anyway @vermouth22 I think I see it more clearly now. The wording `Prefer extensions`  is a bit weird but it just seem that metamask will be used if present. Without `Prefer extensions` means that Brave will be used in any case.

Issue #3044, opened 2022-10-26
L2 providers selection on a browser without Metamask


Aniket-Engg:
It doesn't select that L2 provider, neither show any message on UI. Shows error in browser console:

```
Uncaught (in promise) ReferenceError: ethereum is not defined
    at ExecutionContext.askPermission (app.0.27.0.1665065785599.js:2105:332)
    at ExecutionContext._callee$ (app.0.27.0.1665065785599.js:2115:153)
    at tryCatch (main.0.27.0.1665065785599.js:4696:81)
    at Generator.invoke [as _invoke] (main.0.27.0.1665065785599.js:4714:36)
    at Generator.next (main.0.27.0.1665065785599.js:4701:471)
    at asyncGeneratorStep (main.0.27.0.1665065785599.js:2476:260)
    at _next (main.0.27.0.1665065785599.js:2478:173)
    at main.0.27.0.1665065785599.js:2480:1
    at new Promise (<anonymous>)
    at ExecutionContext.<anonymous> (main.0.27.0.1665065785599.js:2478:86)
```

PR #3046, opened 2022-10-26
Ensure injected provider present

yann300:
fix https://github.com/ethereum/remix-project/issues/3044

### DIFF
--- a/apps/remix-ide/src/blockchain/execution-context.js
+++ b/apps/remix-ide/src/blockchain/execution-context.js
@@ -14,6 +14,8 @@ if (typeof window !== 'undefined' && typeof window.ethereum !== 'undefined') {
   web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'))
 }
 
+const noInjectedProviderMsg = 'No injected provider found. Make sure your provider (e.g. MetaMask) is active and running (when recently activated you may have to reload the page).'
+
 /*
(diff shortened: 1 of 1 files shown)


Aniket-Engg (commented):
Screenshot 2022-10-26 at 7 55 36 PM

It shows a toaster and keep the L2 provider selected. It can be better to do it same as we do for `Injected Provider`

Aniket-Engg (commented):
Not sure if it is only for me but the warning toaster is only shown once on selecting any L2 or injected provider

bunsenstraat:
> Not sure if it is only for me but the warning toaster is only shown once on selecting any L2 or injected provider

The issue is the toaster, not this PR. the toaster system gets the message as a state string. When the message is identical to the previous it won't update state so won't show the toaster. Best thing is to make another PR to change toaster behaviour. needs discussion @Aniket-Engg

bunsenstraat (approved):

Aniket-Engg:
> > Not sure if it is only for me but the warning toaster is only shown once on selecting any L2 or injected provider
> 
> The issue is the toaster, not this PR. the toaster system gets the message as a state string. When the message is identical to the previous it won't update state so won't show the toaster. Best thing is to make another PR to change toaster behaviour. needs discussion @Aniket-Engg

will be tracked in https://github.com/ethereum/remix-project/issues/3060

Aniket-Engg (approved):

PR #3048, opened 2022-10-26
add "create sol github action" to the FE menu

yann300:


libs/remix-ui/workspace/src/lib/actions/workspace.ts
```diff
@@ -533,6 +533,33 @@ export const createNewBranch = async (branch: string) => {
...
+  on: [push]
+
+  jobs:
+    run_sol_contracts_job:
+      runs-on: ubuntu-latest
+      name: A job to run solidity unit tests on github actions CI
+      steps:
+        - name: Checkout
+          uses: actions/checkout@v2
+        - name: Environment Setup
```
ioedeveloper, line 548: This step is not so necessary for a basic setup.

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -531,6 +535,40 @@ export function Workspace () {
...
+        onClick={(e) => {
+          e.stopPropagation()
+          addGithubAction()
+          _paq.push(['trackEvent', 'fileExplorer', 'workspaceMenu', 'addSolidityTesting'])
+          hideIconsMenu(!showIconsMenu)
+        }}
+        className='fab fa-github pl-2'
+      >
+      </span>
+      <span className="pl-3">{'Add Soldity Unit test github action'}</span>
```
ioedeveloper, line 569: Can we find a shorter title? It takes up so much space on the UI.

ioedeveloper: Also, i'm not sure if this option should be only visible for git workspaces. What do you think?

ioedeveloper:
We can also add a context menu option for this for files ending with `_test.sol`

libs/remix-ui/workspace/src/lib/actions/workspace.ts
yann300: ok I removed it

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
yann300: I would make it visible for all the workspace I think.. it gives a bit more visibilty to this feature an encourage to commit workspace to github..

joeizang (approved):
Maybe we can add syntax highlighting for yaml files to the monaco editor later on?

Issue #3050, opened 2022-10-26
Matomo - 4th param needed


ryestew:
The following lines need the Event Name param - 
Try to make the Event Name param be clearly associated with the Event Action - see https://github.com/ethereum/remix-project/issues/3049

_paq.push(['trackEvent', 'solidityStaticAnalyzer', 'analyzeWithRemixAnalyzer'])
_paq.push(['trackEvent', 'udapp', 'AtAddressLoadWithArtifacts'])
_paq.push(['trackEvent', 'solidityUnitTesting', 'runTests'])
_paq.push(['trackEvent', 'compiler', 'compiled_with_config_file'])
_paq.push(['trackEvent', 'fileExplorer', 'publishFileToGist'])
_paq.push(['trackEvent', 'udapp', 'AtAddressLoadWithABI'])
_paq.push(['trackEvent', 'fileExplorer', 'runScript'])
_paq.push(['trackEvent', 'compiler', 'compileWithHardhat'])
_paq.push(['trackEvent', 'udapp', 'syncContracts', compilationSource])
_paq.push(['trackEvent', 'fileExplorer', 'publishFolderToGist'])

PR #3053, opened 2022-10-27
Fix matomo 4th param

yann300:
fix https://github.com/ethereum/remix-project/issues/3050

### DIFF
--- a/apps/remix-ide/src/app/udapp/run-tab.js
+++ b/apps/remix-ide/src/app/udapp/run-tab.js
@@ -78,7 +78,7 @@ export class RunTab extends ViewPlugin {
   }
 
   sendTransaction (tx) {
-    _paq.push(['trackEvent', 'udapp', 'send', 'transaction'])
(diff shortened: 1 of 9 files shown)


apps/remix-ide/src/app/udapp/run-tab.js
```diff
@@ -78,7 +78,7 @@ export class RunTab extends ViewPlugin {
   }
 
   sendTransaction (tx) {
-    _paq.push(['trackEvent', 'udapp', 'sendTx'])
+    _paq.push(['trackEvent', 'udapp', 'send', 'transaction'])
```
ryestew, line 81: I think these names "send" and "transaction" are too generic - consider something like this:_paq.push(['trackEvent', 'udapp', 'sendTx', 'udappTransaction'])

ryestew (changes requested):

ryestew (approved):

PR #3088, opened 2022-11-07
Add a first commit when creating a git workspace

yann300:
Creating a workspace with git enabled should automatically add a first commit which should be the template contents.
As a result it is possible to directly start creating branches.


Aniket-Engg (approved):
LGTM!

PR #3094, opened 2022-11-08
Previously opened tabs will now be restored on new sessions

iamsethsamuel:
This PR adds a feature whereby if the user refreshes the browser for some reason or loads the editor after closing the tab, the previously opened files will be opened


iamsethsamuel:
> could it be possible to still focus on the home page after reloading?

While leaving the prev tabs openned? Yes

yann300:
yes so upon reload the home page should still be the one that user will see first.

iamsethsamuel:
> yes so upon reload the home page should still be the one that user will see first.

Good morning sir, I tried but I couldn't figure out how to open the Homepage. Can I get guidance or a walkthrough on how it should be implemented? I will be glad to join a call

yann300:
I think you should be able to do something like:
`await plugin.call('mainPanel', 'showContent', 'home')`

iamsethsamuel:
> I think you should be able to do something like: `await plugin.call('mainPanel', 'showContent', 'home')`

The problem is that the **TabsUI** component in **remix-ui/tabs/src/lib/remix-ui-tabs.tsx** is does have access to the **FileProviderContext**. And the restoreTabs() method is called from the file manager. There is no way for the 2 to communicate. So I thinking of a way of opening the **home** from the **fileManager**

apps/remix-ide/src/app/panels/tab-proxy.js
```diff
@@ -148,22 +151,56 @@ export class TabProxy extends Plugin {
...
+        const prevSession = window.localStorage.getItem("openedTabs")
+          
+        if(prevSession){
+            const tabs = JSON.parse(prevSession)
+      
+            for (const tab of tabs){
+              if(tab.name != null){
+                try{
+                  this.addTab(tab.name, tab.title, 
+                  () => {
```
iamsethsamuel, line 163: I need to find a way to switch the Tabs, this keeps erroring out. Any suggestions?

apps/remix-ide/src/app/panels/tab-proxy.js
iamsethsamuel, line 1: I noticed that <TabsUI /> has nothing to do with <FileManager /> so I refactored the logic to be handled in the TabProxy instead of the **FileManager**. The problem is that the tabs are not switching when clicked. 

Help me

Aniket-Engg:
As per discussion, we do not look for this feature as of now. So closing this.

PR #3095, opened 2022-11-08
If a tab is opened and no change is made on the file, the editor will…

iamsethsamuel:
… close open a new tab over it

This PR adds a feature whereby if a user opens a file, makes no changes to it, and clicks to open another file. The newly opened file will replace the unchanged file tab.

This prevents cases whereby a user is trying to find a file and ends up with too many unused tabs. I got the feature from VS Coded


bunsenstraat:
It's a personal preference to have this. In vscode this happens but can be very annoying IMO. In this implementation when the api opens a file or many previous files the user is looking at will be closed. The filemanager does not know what process opens the file.

iamsethsamuel:
> It's a personal preference to have this. In vscode this happens but can be very annoying IMO. In this implementation when the api opens a file or many previous files the user is looking at will be closed. The filemanager does not know what process opens the file.

Yeah, the first time they launched it, it was getting on my nerves. But when I got used to it, I can't live without it. I get really frustrated when I have like 10 tabs open that I have to close just to open 1 file

yann300:
- could we try to not have the homepage blinking
 - I think the behavior in vscode is also that if you double click on the file name it will not close the file anymore also if you click a single time, the filename is in italic.

iamsethsamuel:
> * could we try to not have the homepage blinking
> * I think the behavior in vscode is also that if you double click on the file name it will not close the file anymore also if you click a single time, the filename is in italic.

I am just noticing the double click and the italics now. It can be done, but the problem is I don't want to spend so much time working and the PR will be ignored for months

yann300:
ok I understand, so I think it's better to focus on https://github.com/ethereum/remix-project/pull/3094 then

iamsethsamuel:
> ok I understand, so I think it's better to focus on #3094 then

Okay sir, I will let you know when I am done

yann300:
closing this one as we are focusing on https://github.com/ethereum/remix-project/pull/3094 (i'll check that one)

PR #3136, opened 2022-11-17
Fix solidity config file

yann300:


Aniket-Engg:
What is it fixing? Can you add description?

yann300:
if you switch to a version which doesn't support `byzantium`, compilation will fail. removing that allow using the default

Aniket-Engg (approved):

Issue #3082, opened 2022-11-03
Tooltips to review


ryestew:
**Below are tooltips that need updating**

## Icon Panel
- [x] the icon panel has the native tooltips not the the custom ones.

## File Explorer
- [x] **GitHub GIST** 1. Needs to be better positioned
<img width="208" alt="image" src="https://user-images.githubusercontent.com/6050001/199806539-82cb852f-08c2-4c96-b28f-6a44a84bddbf.png">

## Deploy & Run

- [x] **tooltip on disabled Contract selectbox** 
       Sometimes the tooltip doesn't go away.  
<img width="929" alt="image" src="https://user-images.githubusercontent.com/6050001/199803725-229f19ef-04b7-466d-8df2-a548ecc44a1d.png">

- [x] **tooltip on Contract selectbox**  edit:
Maybe it is too long and needs a br tag. Edit: 2. Maybe it is too long and needs a br tag. 
When there is a compiled .sol file, choose the contract to deploy or to use with AtAddress.

- [x] **atAddress** 1. needs a br tag - its too long.
<img width="911" alt="image" src="https://user-images.githubusercontent.com/6050001/199804188-a95d3a36-0d8b-44c0-b2ab-b29414d1b947.png">

- [x] **Environment i** 1. slight adjustment in placement of the tooltip arrow
<img width="320" alt="image" src="https://user-images.githubusercontent.com/6050001/199811095-a1d61612-8659-4448-96af-427880fbf51c.png">

- [x] **Environment Plug** 1. text update: "you want to interact with."   And consider replacing Provider: here is the entire edit: Open chainlist.org and get the connection specs of the chain you want to interact with.
<img width="688" alt="image" src="https://user-images.githubusercontent.com/6050001/199807073-622969b4-e8a3-4329-919c-3ef55b44ffbd.png">

- [x] **Gas Limit** 1. badly positioned text.
<img width="441" alt="image" src="https://user-images.githubusercontent.com/6050001/199808276-0999d7a5-8696-4d2a-86e1-b96e9ea6c452.png">

- [x] **Proxy Contracts** 1. Needs the custom tooltip.
<img width="390" alt="image" src="https://user-images.githubusercontent.com/6050001/199804679-f7ea8788-fe07-4197-9340-b62e334b12e6.png">

- [x] **Function title** 1. Somehow I got a native tooltip on a function's name - but I didn't try to get it to appear.
[...]

PR #3137, opened 2022-11-17
Tooltip Review Story Fixes

joeizang:
addresses story and issue #3082


ryestew (changes requested):
## Deploy & Run

**Contract dropdown** - The tooltip when there is not a compiled contract, is not always coming up and when it does sometime is doesn't go away.  I am not so crazy about the center alignment - but it does seem to be consistant.  So maybe we should leave it as you have it - this is what I'm talking about:
image

**The atAddress tooltip** needs a slight edit: (just adding the word **either**)
To interact with a deployed contract either, enter its address and compile its source *.sol file (with the same compiler settings) or select its .abi file in the editor. </span>

**Tooltip for Load Contract from Address** I think it should have a capital A in address.  

**Proxy tooltips seem to have already been merged**

**Fallback / Receive**
the text update was not done - what was updated?

## Solidity Compiler
On the compile & run "i" monster tooltip: The small text change: "Click the i icon to learn more" not done yet.


## Static Analysis

**Blank tooltip (only an arrow) over titles of subjects**  - look like they have an extra line in them:
image

Run button similar problem with above:
image

ryestew:
## Deploy & Run
**Contract Dropdown** - When there is no contract compiled, I believe this still needs a tooltip.  If not let's discuss.
image

**AtAddress** To interact with a deployed contract, either  - the comma should go before the either (and there should be plenty of commas in ether - but that is another matter)(joke - please ignore the previous material in the parenthesis - but don't ignore the comma stuff)

## Static Analysis
The tooltip in the list are not needed because they are visible anyway.  So please remove the tooltips.
image

ryestew (approved):

Issue #3111, opened 2022-11-11
Run free function (without the need to wrapping it in a contract)


yann300:
feature request from [Let us run free functions directly, without having to add a dummy contract wrapper.](https://twitter.com/PaulRBerg/status/1590827240174284801)

PR #3147, opened 2022-11-21
Add solidity script

yann300:
fix https://github.com/ethereum/remix-project/issues/3111

to test: 
 - create a file with the following content:
 ```
 function run () public {
      console.log("test running free function");
    }
 ```
 -  right click in the function and click "Execute free function"


yann300:
- make sure compiling free function works with the solidity compiler.
 - run with right click / run free function.
 - next PR: use `.script.sol` , but rename `_test.sol` to `.test.sol`

ryestew (changes requested):
I don't see Execute Script in the rightclick menu.

ryestew (approved):

Issue #3154, opened 2022-11-22
Is debugger ready for proxy contract (UUPS) ?


lkebin:
When I debugging proxy contract transaction, the state panel always display `No data available` and have a alert message on the top

```
Source location not available, neither in Sourcify nor in Etherscan. Please make sure the Etherscan api key is provided in the settings.
```

![image](https://user-images.githubusercontent.com/1335935/203236172-b4cf8041-6693-404f-8231-00197929283b.png)

My test contract deployed on Remix VM (London), then debug the mint function call.

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.4;

import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";

contract MyToken is Initializable, ERC20Upgradeable, OwnableUpgradeable, UUPSUpgradeable {
    /// @custom:oz-upgrades-unsafe-allow constructor
    constructor() {
        _disableInitializers();
    }

    function initialize() initializer public {
        __ERC20_init("MyToken", "MTK");
        __Ownable_init();
        __UUPSUpgradeable_init();
    }

    function mint(address to, uint256 amount) public onlyOwner {
        _mint(to, amount);
    }

    function _authorizeUpgrade(address newImplementation)
        internal
        onlyOwner
        override
    {}
}
```

Aniket-Engg:
cc: @yann300

PR #3168, opened 2022-11-28
fix debugging proxy contract

yann300:
fix https://github.com/ethereum/remix-project/issues/3154


ioedeveloper:
@yann300 What is the difference between UUPSBytecode and UUPSDeployedBytecode?

yann300:
`UUPSBytecode` is the bytecode that you get after compiling, `UUPSDeployedBytecode` is the bytecode that is actually deployed. For instance `UUPSDeployedBytecode` doesn't contain the constructor code

ioedeveloper (approved):

PR #3173, opened 2022-11-29
show local variable value in the editor context

yann300:
- while in debug: show local variables values when hovering it

### DIFF
--- a/apps/remix-ide/src/app/plugins/parser/code-parser.tsx
+++ b/apps/remix-ide/src/app/plugins/parser/code-parser.tsx
@@ -78,6 +78,8 @@ export class CodeParser extends Plugin {
     setCurrentFileAST: (text?: string) => Promise<ParseResult>
     getImports: () => Promise<CodeParserImportsData[]>
 
+    debuggerIsOn: boolean = false
+
 
     constructor(astWalker: any) {
         super(profile)
@@ -98,7 +100,7 @@ export class CodeParser extends Plugin {
         }
         const showGasSettings = await this.call('config', 'getAppParameter', 'show-gas')
         const showErrorSettings = await this.call('config', 'getAppParameter', 'display-errors')
-        if (showGasSettings || showErrorSettings || completionSettings) {
+        if (showGasSettings || showErrorSettings || completionSettings || this.debuggerIsOn) {
             await this.compilerService.compile()
         }
     }
@@ -157,6 +159,16 @@ export class CodeParser extends Plugin {
 
         await this.compilerService.init()
         this.on('solidity', 'compilerLoaded', async () => { 
+            await this.reload()
+        })
+
+        this.on('debugger', 'startDebugging', async () => { 
+            this.debuggerIsOn = true
+            await this.reload()
+        })
+
+        this.on('debugger', 'stopDebugging', async () => { 
+            this.debuggerIsOn = false
             await this.reload()
         })
     }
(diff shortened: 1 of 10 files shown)


apps/remix-ide/src/app/plugins/parser/code-parser.tsx
```diff
@@ -173,7 +171,7 @@ export class CodeParser extends Plugin {
      * @returns 
      */
     async getLastCompilationResult() {
-        return this.compilerAbstract
+        return await this.call('compilerArtefacts', 'get', '__last')
```
bunsenstraat, line 174: that will not work. 'compilerArtefacts' doesn't have data from the parser compiler as 'compilerArtefacts' only listens to 'solidity' plugin which this isn't

libs/remix-ui/editor/src/lib/providers/hoverProvider.ts
```diff
@@ -145,6 +145,48 @@ export class RemixHoverProvider implements languages.HoverProvider {
...
+                    if (global !== null && global[nodeAtPosition?.expression?.name][nodeAtPosition.memberName] && global[nodeAtPosition?.expression?.name][nodeAtPosition.memberName]) {
+                        contents.push({
+                            value: `${nodeAtPosition.memberName} = ${global[nodeAtPosition?.expression?.name][nodeAtPosition.memberName]}`
+                        })
+                    }
+                }
+            } catch (e) {}
+
+            try {
+                const decodedVar = await this.props.plugin.call('debugger', 'decodeLocalVariable', nodeAtPosition.id)
```
bunsenstraat, line 172: the nodes parser knows about might have different ids as the parser compiler runs continuously when typing

libs/remix-ui/editor/src/lib/providers/hoverProvider.ts
```diff
@@ -145,6 +145,48 @@ export class RemixHoverProvider implements languages.HoverProvider {
...
+                if (decodedVar !== null) {
+                    const value = `${nodeAtPosition.name} = ${typeof(decodedVar.value) === 'string' ? decodedVar.value : JSON.stringify(decodedVar.value, null, '\t')}`
+                    contents.push({
+                        value
+                    })
+                }
+            } catch (e) {}
+
+            try {
+                const decodedVar = await this.props.plugin.call('debugger', 'decodeStateVariable', nodeAtPosition.id)
```
bunsenstraat, line 182: same here

libs/remix-ui/editor/src/lib/providers/hoverProvider.ts
bunsenstraat: so this would cause a bug

bunsenstraat (changes requested):
It won't work in this form. The two compilers should remain independent and the tree generated by continuously typing has different nodes from the one that the  debugger uses.

bunsenstraat:
@yann300 https://github.com/ethereum/remix-project/pull/3173/commits/cbf4d0c068131a063c7f81660f3a3e05c3cf1501 this will use the debugger's artefacts when it's debugging. It will also provide other necessary data for the hover to function.

bunsenstraat:
there is still a bug, the tracemanager trace is empty in some conditions, needs discussion & analysis.
condition is: start debug with terminal, stop it in the debugger, start it again in debugger. the trace will load but it's empty in the tracemanager instance.

libs/remix-debug/src/Ethdebugger.ts
```diff
@@ -100,6 +101,27 @@ export class Ethdebugger {
     return this.callTree.findScope(step)
   }
 
+  async decodeLocalVariableByIdAtCurrentStep (step: number, id: number) {
+    const variable = this.callTree.getLocalVariableById(id)
+    if (!variable) return null
+    const stack = this.traceManager.getStackAt(step)
```
bunsenstraat, line 107: this fails when restarting the debugger via the debugger itself. the trace is empty, if you start it with the terminal it's fine.

bunsenstraat (approved):

PR #3174, opened 2022-11-30
fixing readonly access with remixd

LianaHus:
authors @yann300 @LianaHus


libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -350,6 +350,10 @@ export function Workspace () {
     )
   }
 
+  const formatNameForReadonly = (name: string) => {
+    return global.fs.readonly ? name + " (read-only)" : name
```
yann300, line 354: from where does `global.fs` come from?

LianaHus: it's a file system we use. you mean where it is defined/?

yann300 (approved):

PR #3181, opened 2022-12-01
Add solhint to the solidity compiler ui

yann300:


Aniket-Engg:
Solhint is part of SSA plugin now.

Issue #3052, opened 2022-10-27
Security checks for proxy upgrade.


ioedeveloper:
TODO:
- [x] #3189
- [x] #3190
- [x] #3388
- [x] #3389

PR #3187, opened 2022-12-05
Save proxy addresses and storage layout for upgrade. (Secure Upgrade Part1)

ioedeveloper:
This PR creates a file based on network name for upgradeable contracts (`.deploys/upgradeable-contracts/goerli.json`).
- The proxy address is stored in other to provide a list of previously deployed proxy addresses a user can make upgrades to. This will be implemented in a follow-up PR.
- The storage layout is stored to enable verifying that any storage layout changes between subsequent versions are compatible. (see https://docs.openzeppelin.com/upgrades-plugins/1.x/network-files). This will be implemented in a follow-up PR.

Linked to https://github.com/ethereum/remix-project/issues/3052
Closes https://github.com/ethereum/remix-project/issues/3189

### DIFF
--- a/apps/remix-ide/src/blockchain/blockchain.js
+++ b/apps/remix-ide/src/blockchain/blockchain.js
@@ -177,11 +177,11 @@ export class Blockchain extends Plugin {
         _paq.push(['trackEvent', 'blockchain', 'Deploy With Proxy', 'Proxy deployment failed: ' + error])
         return this.call('terminal', 'logHtml', log)
       }
-      const hasPreviousDeploys = await this.call('fileManager', 'exists', `.deploys/upgradeable-contracts/${networkInfo.name}.json`)
+      const hasPreviousDeploys = await this.call('fileManager', 'exists', `.deploys/upgradeable-contracts/${networkInfo.name}/UUPS.json`)
       
       // make deploys folder read only.
       if (hasPreviousDeploys) {
-        const deployments = await this.call('fileManager', 'readFile', `.deploys/upgradeable-contracts/${networkInfo.name}.json`)
+        const deployments = await this.call('fileManager', 'readFile', `.deploys/upgradeable-contracts/${networkInfo.name}/UUPS.json`)
         const parsedDeployments = JSON.parse(deployments)
 
         parsedDeployments.deployments.push({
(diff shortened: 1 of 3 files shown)


apps/remix-ide/src/blockchain/blockchain.js
```diff
@@ -170,15 +170,39 @@ export class Blockchain extends Plugin {
...
+
+        parsedDeployments.deployments.push({
+          fork: networkInfo.currentFork,
+          proxyAddress: address,
+          layout: implementationContractObject.contract.object.storageLayout,
+          date: new Date().toISOString()
+        })
+        await this.call('fileManager', 'writeFile', `.deploys/upgradeable-contracts/${networkInfo.name}.json`, JSON.stringify(parsedDeployments, null, 2))
+      } else {
+        await this.call('fileManager', 'writeFile', `.deploys/upgradeable-contracts/${networkInfo.name}.json`, JSON.stringify({
```
yann300, line 195: I think we should find a way to not mix the different network.
maybe the solution would be to create a subfolder for each network.

ioedeveloper: Ok, i will make the change.

ioedeveloper: Upgrades don't have anything to do with previous implementation address, so i assumed it will not be needed. I can add it if you like.

ioedeveloper: Current implementation address becomes previous implementation address when upgrade is initiated. Anyway, i will add it.

apps/remix-ide/src/blockchain/blockchain.js
```diff
@@ -170,15 +170,41 @@ export class Blockchain extends Plugin {
...
   
         _paq.push(['trackEvent', 'blockchain', 'Deploy With Proxy', 'Proxy deployment failed: ' + error])
         return this.call('terminal', 'logHtml', log)
       }
-      if (networkInfo.name === 'VM') this.config.set('vm/proxy', address)
-      else this.config.set(`${networkInfo.name}/${networkInfo.currentFork}/${networkInfo.id}/proxy`, address)
+      const hasPreviousDeploys = await this.call('fileManager', 'exists', `.deploys/upgradeable-contracts/${networkInfo.name}/UUPS.json`)
+      
+      // make deploys folder read only.
+      if (hasPreviousDeploys) {
```
yann300, line 183: is this also possible to put the contract name in the JSON?

yann300:
I tried to update a proxy but the new deployment isn't added to the JSON..

apps/remix-ide/src/blockchain/blockchain.js
ioedeveloper: Yes, it is.

yann300:
if you have `lastUpdated` maybe you don't need `date`

yann300:
> if you have lastUpdated maybe you don't need date

or the other way around

yann300:
I think we have to keep track of the previous deployments. what do you think?
I mean `0xd8b934580fcE35a11B58C6D73aDeE468a2833fa8": {}` should be an array.
Cause if we don't keep track of the previous layouts in storage, we won't be able to do the sanity check?

yann300:
also if the key is the proxy address, maybe we don't need the `proxyAddress` attribute.

yann300 (approved):

PR #3192, opened 2022-12-06
space at the bottom of FE

LianaHus:


joeizang (changes requested):
There seems to be extra padding between items in file explorer. See the screenshots below
fepaddingsnapshot
The extra padding and margins is shown here:
fepadding issue

LianaHus:
@joeizang

joeizang (approved):

PR #3199, opened 2022-12-07
Use the cache for the calltree

yann300:
This update the Breakpoint Manager to use the cache populated by InternalCallTree instead of recomputing the line/column position.

### DIFF
--- a/libs/remix-debug/src/code/breakpointManager.ts
+++ b/libs/remix-debug/src/code/breakpointManager.ts
@@ -14,7 +14,6 @@ export class BreakpointManager {
   callTree
   solidityProxy
   breakpoints
-  locationToRowConverter
 
   /**
     * constructor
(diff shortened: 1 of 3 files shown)


libs/remix-debug/src/code/breakpointManager.ts
```diff
@@ -22,13 +22,12 @@ export class BreakpointManager {
...
     * @return {Function} _locationToRowConverter - function implemented by editor which return a column/line position for a char source location
     */
-  constructor ({ traceManager, callTree, solidityProxy, locationToRowConverter }) {
+  constructor ({ traceManager, callTree, solidityProxy }) {
     this.event = new EventManager()
     this.traceManager = traceManager
     this.callTree = callTree
     this.solidityProxy = solidityProxy
     this.breakpoints = {}
-    this.locationToRowConverter = locationToRowConverter
```
bunsenstraat, line 31: probably we could remove the declaration too at line 17

bunsenstraat (approved):

Issue #3216, opened 2022-12-13
Load language setting via a URL


ryestew:
We need to set the language choice of the **labels** via a URL param

PR #2581, opened 2022-07-04
UI Translation

drafish:
1.  add remix-ui-locale-module
2. add locale file, include en-US and zh-CN
3.  translate setting, pluginManager, filePanel, home, terminal, search, solidity, udapp, debugger
image

PR #3219, opened 2022-12-14
Intl patch2

drafish:
more content translated

request a review @yann300 @ryestew @Aniket-Engg


drafish:
I'v ran failed tests locally, they all passed.

ryestew:
@drafish - Please rebase this PR. Thanks

drafish:
> @drafish - Please rebase this PR. Thanks

Done

LianaHus:
@drafish  looks like we have conflicts here. Could you please fix this?

drafish:
@LianaHus I just fixed that.

ryestew (changes requested):
This PR has 2 parts to it.  The first is fine.  The second - add custom language with url, and e2e test for this feature (fix https://github.com/ethereum/remix-project/issues/3216 ) - seems to misunderstand the issue https://github.com/ethereum/remix-project/issues/3216.

I created that issue so that if a language is already setup in Remix, then there could be a URL param that loads Remix and would display the chosen language - so a url like remix.ethereum.org/?#lang=zh_CN  
or something like that (probably just lang=zh would be enough).  A URL like that would allow us, for example, to send users in China a specific url and they wouldn't need to go to the settings panel to choose the language.

I didn't quite understand what your language loading would do.  We can discuss that further.

But for this PR, please remove the commits with the custom language by url and the associated e2e tests for this.  The first part of the PR is needed.

It would be great if you can make another PR that would address selecting a language via a URL parameter.

drafish:
Oh I see. I misunderstand. I thought you want to add a custom language with url, just like add a custom compiler with url. I removed those commits. If you want this feature, I will make another PR.

About this issue #3216 , actually I'v implemented this feature in PR #2581 . Check the link below, you will see chinese as default language.
https://remix.ethereum.org/#locale=zh

But, it seems like there is a little issue in this feature. It should change the url parameter when I switch language. I will check what's wrong, and make another PR. And I will also change the parameter name, locale -> lang.  @ryestew

yann300 (approved):

ryestew (approved):

PR #3224, opened 2022-12-15
Lengthy Compile Button Text Fix

joeizang:
Fixes an issue where the filename being compiled is too long and so it skews the arrangement of text in the compile button.

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
+++ b/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
@@ -941,13 +941,15 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
               placement="auto"
               tooltipId="overlay-tooltip-compile"
               tooltipText={<div className="text-left">
-                  {!(configFilePath === '' && state.useFileConfiguration) && <div><b>Ctrl+S</b> for compiling</div>}
+                  {!(configFilePath === '' && state.useFileConfiguration) && <div><b>Ctrl+S</b> to compile {state.compiledFileName} </div>}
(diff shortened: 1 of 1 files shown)


Aniket-Engg (approved):

libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -941,13 +941,15 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
...
-                  {!(configFilePath === '' && state.useFileConfiguration) && <div><b>Ctrl+S</b> for compiling</div>}
+                  {!(configFilePath === '' && state.useFileConfiguration) && <div><b>Ctrl+S</b> to compile {state.compiledFileName.endsWith('.sol') ? state.compiledFileName : null} </div>}
                   {(configFilePath === '' && state.useFileConfiguration) && <div> No config file selected</div>}
                 </div>}
             >
               <div className="d-flex align-items-center justify-content-center">
                 { <i ref={compileIcon} className="fas fa-sync remixui_iconbtn ml-2" aria-hidden="true"></i> }
-                <div className="d-flex justify-content-between align-items-center">
+                <div className=""
+                  style={{ whiteSpace: 'nowrap', fontSize: 'calc(1em + (0.8vw - 0.5em) * 0.2)', overflow: 'hidden', textOverflow: 'ellipsis' }}
```
LianaHus, line 951: we shouldn't use css when we can use bootstrap class
and why do you need 'calc(1em + (0.8vw - 0.5em) * 0.2)'?

LianaHus: @joeizang

PR #3251, opened 2022-12-22
Fix deps ethereujs/vm

yann300:
partially_fix https://github.com/ethereum/remix-project/issues/3249

 - this fixes the deps issues between `@ethereumjs/utils` and `ethereumjs-utils`
 - moving to ethereumjs/vm 6.3.0.


Aniket-Engg (approved):
change set looks fine

PR #3252, opened 2022-12-23
set default solidity version to 0.8.17

### LINKED ISSUE #3248: set solidity compiler version to latest

Aniket-Engg:
fixes #3248 

This will set 0.8.17 as default Solidity version in `Solidity Compiler` plugin and will also set latest local version to 0.8.17


bunsenstraat:
I think we can automate this. We already have some scripts that run to fetch the remote lists, ie on e2e, probably can use that to get the default latest instead of this static.

bunsenstraat (approved):

PR #3254, opened 2022-12-27
show compiler input in details

Aniket-Engg:


yann300:
do you think there could be a more obvious place where to retrieve this? this is now buried at the end of the modal, maybe we could at least put it at the beginning.

Aniket-Engg:
> do you think there could be a more obvious place where to retrieve this? this is now buried at the end of the modal, maybe we could at least put it at the beginning.

Made `COMPILERINPUT` first item in modal

yann300:
ok please @ryestew @vermouth22 see this new feature, this needs some more visibility (video, tutorial, ...)

yann300 (approved):

PR #3267, opened 2022-12-30
first stepps

bunsenstraat:


bunsenstraat:
@yann300 @joeizang it works, needs some discussion though about some config params and outputs and what to do with MD files, syntax etc. but it works.

PR #3271, opened 2023-01-02
more tests for ERC20 template

Aniket-Engg:
related to #3114

### DIFF
--- a/libs/remix-ws-templates/src/templates/ozerc20/tests/MyToken_mintable_test.sol
+++ b/libs/remix-ws-templates/src/templates/ozerc20/tests/MyToken_mintable_test.sol
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: GPL-3.0
+
+pragma solidity >=0.7.0 <0.9.0;
+import "remix_tests.sol";
+import "remix_accounts.sol";
+import "../contracts/MyToken.sol";
+
+contract MyTokenTest is MyToken {
+
+    address acc0;
+    address acc1;
+    address acc2;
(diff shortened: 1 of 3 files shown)


libs/remix-ws-templates/src/templates/ozerc20/tests/MyToken_mintable_test.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    function testTokenApprove() public {
+        Assert.equal(allowance(acc0, acc3), 0, "token allowance should be zero initially");
+        approve(acc3, 500);
+        Assert.equal(allowance(acc0, acc3), 500, "token allowance did not match");
+    }
+
+    function failTestTokenApproveForZeroSpenderAddress() public {
+        approve(address(0), 500);
```
yann300, line 83: this test should fail? how is this tested in the test suite?

Aniket-Engg: It just shows error with revert reason on plugin as of now

yann300: maybe there's a way to get that revert reason in solidity with a `try catch`

Aniket-Engg: We can use `try catch` for external method, then it will be difficult to use custom sender in the required way

yann300 (approved):

PR #3281, opened 2023-01-11
fix hide warnings

### LINKED ISSUE #3278: The Option of "Hide warning" doesn't work
Here is the snapshot.
image

Aniket-Engg:
fixes #3278 

This also removes the warnings count badge from the Solidity Compiler plugin icon. Warnings will still be visible in editor.

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
+++ b/libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
@@ -43,7 +43,7 @@ export const SolidityCompiler = (props: SolidityCompilerProps) => {
       const hide = await api.getAppParameter('hideWarnings') as boolean || false
       setHideWarnings(hide)
     })()
-  }, [])
+  })
 
   useEffect(() => {
     if (badgeStatus[currentFile]) {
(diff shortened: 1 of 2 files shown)


libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
```diff
@@ -43,7 +43,7 @@ export const SolidityCompiler = (props: SolidityCompilerProps) => {
       const hide = await api.getAppParameter('hideWarnings') as boolean || false
       setHideWarnings(hide)
     })()
-  }, [])
+  })
```
ioedeveloper, line 46: @Aniket-Engg By removing the dependency array, this useEffect will be fired every single time this component is re-rendered. I don't think it is a good solution in my opinion, we can tie the change to a variable and pass it to the dependency array. `, [hideWarnings])`

ioedeveloper (changes requested):

libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
Aniket-Engg: I tried that. It is not solving the issue

ioedeveloper (approved):

PR #3306, opened 2023-01-25
Home Tab Quick Updates Fix

joeizang:
Fixes an issue where Intro to Solidity points to a different set of tutorials in LearnEth.

### DIFF
--- a/libs/remix-ui/home-tab/src/lib/components/homeTabLearn.tsx
+++ b/libs/remix-ui/home-tab/src/lib/components/homeTabLearn.tsx
@@ -31,11 +31,10 @@ function HomeTabLearn ({plugin}: HomeTabLearnProps) {
     window.open("https://remix-ide.readthedocs.io/en/latest/remix_tutorials_learneth.html?highlight=learneth#learneth-tutorial-repos", '_blank')
   }
 
-  const startLearnEthTutorial = async (tutorial: 'home' | 'basics' | 'soliditybeginner' | 'deploylibraries') => {
+  const startLearnEthTutorial = async (tutorial: 'basics' | 'soliditybeginner' | 'deploylibraries') => {
     await plugin.appManager.activatePlugin(['solidity', 'LearnEth', 'solidityUnitTesting'])
     plugin.verticalIcons.select('LearnEth')
-    tutorial === 'home' ? plugin.call('LearnEth', 'home') :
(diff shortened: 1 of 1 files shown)


ryestew (dismissed):
One more quick fix - when you click the LearnEth icon on the home tab - it should open the home of LearnEth and not the Remix intro.
image

joeizang:
> One more quick fix - when you click the LearnEth icon on the home tab - it should open the home of LearnEth and not the Remix intro. image

Default seems to be Remix Basics. Tried changing it but it hasn't made a difference.

LianaHus:
Because the method is to open a tutorial. you should probably just start the plugin instead

LianaHus:
it will not work if you already clicked on tutorial before, you need to expose a method

joeizang:
> it will not work if you already clicked on tutorial before, you need to expose a method

This has been addressed.

libs/remix-ui/home-tab/src/lib/components/homeTabLearn.tsx
```diff
@@ -31,13 +31,25 @@ function HomeTabLearn ({plugin}: HomeTabLearnProps) {
...
-  const startLearnEthTutorial = async (tutorial: 'basics' | 'useofweb3js' | 'deploylibraries') => {
+  const startLearnEthTutorial = async (tutorial: 'basics' | 'soliditybeginner' | 'deploylibraries') => {
     await plugin.appManager.activatePlugin(['solidity', 'LearnEth', 'solidityUnitTesting'])
     plugin.verticalIcons.select('LearnEth')
     plugin.call('LearnEth', 'startTutorial', 'ethereum/remix-workshops', 'master', tutorial)
     _paq.push(['trackEvent', 'hometab', 'startLearnEthTutorial', tutorial])
   }
 
+  const goToLearnEthHome = async () => {
+    console.log({ plugin })
```
Aniket-Engg, line 42: remove log

Aniket-Engg (commented):
Other things seem fixed. But it still opens the tutorial on clicking on complete bar instead of clicking on `Get Started` button

libs/remix-ui/home-tab/src/lib/components/homeTabLearn.tsx
joeizang: done

Aniket-Engg (approved):

PR #3311, opened 2023-01-26
resolver use package-lock.json or yarn lock

yann300:
- url resolver now resolves dependencies using the version number provided by yarn.lock or package-lock.json.
 - remove uneeded cache.
 - fix resolving the package.json dep.


bunsenstraat (approved):

libs/remix-url-resolver/src/resolve.ts
```diff
@@ -127,11 +135,51 @@ export class RemixURLResolver {
...
+              if (version) url = url.replace(pkg, `${pkg}@${version}`)
+            }
+          }
+        } catch (e) {
+          console.log(e)
+        }
+      }
       const req = 'https://unpkg.com/' + url
       const response: AxiosResponse = await axios.get(req, { transformResponse: [] })
+      console.log('handleNpmImport', req)
```
ioedeveloper, line 182: Do you need this console.log?

libs/remix-core-plugin/src/lib/compiler-content-imports.ts
```diff
@@ -16,12 +16,39 @@ export type ResolvedImport = {
...
+      try {
+        let yarnLock
+        if (await this.call('fileManager', 'exists', './yarn.lock')) {
+          yarnLock = await this.call('fileManager', 'readFile', './yarn.lock')
+        }
+
+        let packageLock
+        if (await this.call('fileManager', 'exists', './package-lock.json')) {
+          packageLock = await this.call('fileManager', 'readFile', './package-lock.json')
+          packageLock = JSON.parse(packageLock)
```
ioedeveloper, line 32: Needs to be tested with a large `package-lock.json` file

bunsenstraat: Very true. We need an example.

bunsenstraat:
I tested it with a package file but the behavior is inconsistent. Needs some debugging.

PR #3344, opened 2023-02-04
Nx

bunsenstraat:
do the minifying in our production build with Terser.
But turned off mangling, because that causes a bug in the build exposed by CI.  
Mangling does work with with the option: keep_fnames: true but it seems unstable in CI. so just disabling mangling should do fine. 
There are other minifier options available, swc, uglify & esbuild. But terser seems the only one that works for now, but it's very slow to build.
updated nx again. changed webpack config to reflect how it is done by the nx project generator. 
because we have minified we probably should turn it off on the cloudflare.. ? gzip should still be done by cloudflare, we can also do the zipping itself in webpack... our own gzip on production build. but that's not necessary.
most of the bandwidth is gained by gzip.
In this setup source maps are throwing warnings during the build so we have to ignore the warnings in webpack config.
The assets directory is still too big, filled with unused monaco stuff. will do that in another PR.
ps. netlify is not using our production builds at this moment. maybe it should, more realistic.


package.json
```diff
@@ -221,17 +221,18 @@
...
+    "@nrwl/jest": "15.6.3",
+    "@nrwl/js": "15.6.3",
+    "@nrwl/linter": "15.6.3",
+    "@nrwl/node": "15.6.3",
+    "@nrwl/react": "15.6.3",
+    "@nrwl/tao": "^15.6.3",
+    "@nrwl/web": "15.6.3",
+    "@nrwl/webpack": "15.6.3",
+    "@nrwl/workspace": "^15.6.3",
+    "@svgr/webpack": "^6.5.1",
```
Aniket-Engg, line 235: is this dep required?

bunsenstraat: you mean @svgr/webpack?
yes...

bunsenstraat: at least I got errors without it. can check again.

yann300 (approved):

PR #3350, opened 2023-02-06
exclude initial msgs to matomo

LianaHus:


apps/remix-ide/src/assets/js/loader.js
```diff
@@ -4,13 +4,19 @@ const domains = {
...
+
 if (domains[window.location.hostname]) {
   var _paq = window._paq = window._paq || []
   /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
   _paq.push(['disableCookies']);
   _paq.push(['enableJSErrorTracking']);
   _paq.push(['trackPageView']);
   _paq.push(['enableLinkTracking']);
+
+  if (!(window.localStorage.getItem('config-v0.8:.remix.config')) ||
```
yann300, line 16: there are some parenthesis that you don't need here.

apps/remix-ide/src/assets/js/loader.js
```diff
@@ -13,8 +13,8 @@ if (domains[window.location.hostname]) {
   _paq.push(['trackPageView']);
   _paq.push(['enableLinkTracking']);
   _paq.push(['enableHeartBeatTimer']);
-  if (!(window.localStorage.getItem('config-v0.8:.remix.config')) ||
-    ((window.localStorage.getItem('config-v0.8:.remix.config')) && !(window.localStorage.getItem('config-v0.8:.remix.config')).includes('settings/matomo-analytics'))) {
+  if (!window.localStorage.getItem('config-v0.8:.remix.config') ||
+    (window.localStorage.getItem('config-v0.8:.remix.config') && !window.localStorage.getItem('config-v0.8:.remix.config').includes('settings/matomo-analytics'))) {
```
yann300, line 17: what is supposed to return  `window.localStorage.getItem('config-v0.8:.remix.config').includes('settings/matomo-analytics')` ? if it's a bool, you can't check it like this you have to use `undefined` or `null`

yann300 (approved):

Issue #413, opened 2020-09-09
Export from Remix IDE


Aniket-Engg:
Investigate and implement feasible ones from these:

- [x] Export/Download a single file
- [x] Export/Import a folder
- [x] Export compilation details in a file for a compiled contract
- [ ] Export AST for a contract
- [x] Export workspace once #406 is implemented

Aniket-Engg:
AST export as a file is not required as per the team discussion. AST is there in compiler output in build-info folder files

PR #3358, opened 2023-02-10
download folder from FE

Aniket-Engg:
part of #413 

It allows to download a folder in zip format, by right clicking on folder name

### DIFF
--- a/apps/remix-ide/src/app/files/fileManager.ts
+++ b/apps/remix-ide/src/app/files/fileManager.ts
@@ -1,5 +1,6 @@
 'use strict'
 import { saveAs } from 'file-saver'
+import JSZip from 'jszip'
 import { Plugin } from '@remixproject/engine'
 import * as packageJson from '../../../../../package.json'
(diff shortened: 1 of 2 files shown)


yann300:
could be nice to have a way to download the root folder as zip.

ryestew:
Currently the Download link in the hamburger is for downloading all workspaces.  The label there should probably be updated to "Download All Workspaces" (I'll make an issue)  But then, should the **download workspace** link go in this FE context menu or in the Hamburger menu?  

If we can organize these context menu links so there are not too many - e.g. having sub menus - then I think it should go here.  Otherwise, the link to **download current workspace** could go in the hamburger.

yann300:
LGTM!

vermouth22:
On testing, this is behaving strangely. On right-click of a folder, it will display context menu and download successfully as a straight .zip of the folder. However, attempting to download other folders in the same Workspace or other Workspaces afterwards does not work. I have to reload the entire site to have the context menu display again for download.

Issue #3361, opened 2023-02-12
Flattened files not compiling


n3me5is-git:
The old flattener plugin was working good, the new embedded flattener is not working, it produces a solidity file that does not compile.
In my case I have a project that uses openzeppelin dependencies (NFT contract), che original file compiles, but the flattened version not. 

It seems that the functions are in a wrong order, so the interfaces are not before the implementations and so on. So the compiler throws an error. I'm sure it worked in the past with the old flattener plugin because I deployed the same code in the past, flattened it and successfully compiled it inside Remix and in bscscan too.

How to reproduce: create a NFT contract using openzeppelin dependencies, for example "contract MyContract is ERC721, ERC721Enumerable, ERC2981, Ownable" compile it, flatten it and then try to compile the flattened version.

Alucial:
i have the same issue after going live on eth mainnet.
Can we get a quick patch to the old flattener?

Aniket-Engg:
@Alucial can you share exact contract to quickly reproduce this issue? And what error compiler shows?

n3me5is-git:
Here is a demo code to reproduce it. Try to compile this code, it will compile without issues.
Then right click and Flatten it.
Then try co compile the flattened version.

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.15;

import "@openzeppelin/contracts@4.7.1/access/Ownable.sol";
import "@openzeppelin/contracts@4.7.1/token/ERC721/ERC721.sol";
import "@openzeppelin/contracts@4.7.1/token/ERC721/extensions/ERC721Enumerable.sol";
import "@openzeppelin/contracts@4.7.1/token/common/ERC2981.sol";
import "@openzeppelin/contracts@4.7.1/utils/Strings.sol";


contract MyNFT is ERC721, ERC721Enumerable, ERC2981, Ownable {

    using Strings for uint256;

    constructor() ERC721("MyNFT", "NFT") { }
    
    function _beforeTokenTransfer(address from, address to, uint256 tokenId) internal override(ERC721, ERC721Enumerable)
    {
        // DO SOMETHING
        super._beforeTokenTransfer(from, to, tokenId);
    }

    function _burn(uint256 tokenId) internal override(ERC721) {
        super._burn(tokenId);
    }

    function supportsInterface(bytes4 interfaceId) public view override(ERC721, ERC721Enumerable, ERC2981)
        returns (bool)
    {
        return super.supportsInterface(interfaceId);
    }

[...]

PR #3366, opened 2023-02-13
Flattener patch

joeizang:
Addresses bug #3361


Aniket-Engg (commented):
Changes in yarn.lock can be reverted

Aniket-Engg (approved):

Issue #3379, opened 2023-02-15
Delete all workspaces


LianaHus:
Add an item in hamburger to delete all workspaces at once

PR #3380, opened 2023-02-15
Delete all workspaces added in the workspaces hamburger

LianaHus:
fix https://github.com/ethereum/remix-project/issues/3379


ioedeveloper:
I reviewed the code and it looks good to me.

apps/remix-ide/src/app/tabs/locales/zh/filePanel.json
```diff
@@ -9,6 +9,8 @@
   "filePanel.workspace.rename": "重命名工作空间",
   "filePanel.workspace.delete": "删除工作空间",
   "filePanel.workspace.deleteConfirm": "确定要删除当前工作空间?",
+  "filePanel.workspace.deleteAll": "Delete All Workspaces",
+  "filePanel.workspace.deleteAllConfirm": "Are you absolutely sure you want to delete all your workspace?",
```
Aniket-Engg, line 13: I think it should be `all your workspaces`

Aniket-Engg: Not sure if we should also add something like `Deleted workspaces can not be restored in any manner`

LianaHus: thanks. will update

apps/remix-ide/src/app/tabs/locales/en/filePanel.json
```diff
@@ -9,6 +9,8 @@
   "filePanel.workspace.rename": "Rename Workspace",
   "filePanel.workspace.delete": "Delete Workspace",
   "filePanel.workspace.deleteConfirm": "Are you sure to delete the current workspace?",
+  "filePanel.workspace.deleteAll": "Delete All Workspaces",
+  "filePanel.workspace.deleteAllConfirm": "Are you absolutely sure you want to delete all your workspace? Deleted workspaces can not be restored in any manner.",
```
Aniket-Engg, line 13: Can you make `all your workspace` to `all your workspaces`

Aniket-Engg (approved):

bunsenstraat (changes requested):
The two next buttons after delete all have lost their hover and don't seem to be working in CI either.

bunsenstraat (approved):

PR #3390, opened 2023-02-16
double load fix

bunsenstraat:
Now webpack injects it because the new nx has it ON by default
You can see it by checking the doc in the network tab. it injects the scripts.


Aniket-Engg (commented):
Please confirm that desktop app will run fine with this change

Aniket-Engg (approved):

PR #3415, opened 2023-02-22
Misc: fixes providers list in udapp

yann300:


Aniket-Engg (commented):
I think `Injected Provider` at first place looks a bit odd as `Remix VM (merge)` is selected by default. I think we can put it after `Remix VM (Berlin)`

LianaHus (approved):

LianaHus:
I agree with Yann here. Till we don't have a second level Metamask can go to the first place.

PR #3429, opened 2023-02-23
download contract compilation details

Aniket-Engg:
related to #413 

This shows option to download compilation details on modal showing details.

On clicking on `Download` , modal will not close.


yann300:
there could be 3 options:
 - download compilation details as it is show in the modal
 - download the compilation result (this is the standard output)
 - downloac the compilation input (this is the standard input)

Aniket-Engg:
> there could be 3 options:
> 
> * download compilation details as it is show in the modal
> * download the compilation result (this is the standard output)
> * downloac the compilation input (this is the standard input)

will do it in a separate PR

yann300 (approved):

LianaHus (changes requested):
i see a small problem here. when we click download modal shouldn't disappear.

Aniket-Engg:
> i see a small problem here. when we click download modal shouldn't disappear.

ok, lets discuss this

Issue #3320, opened 2023-01-31
Add support for Trust Wallet provider


george-gkas:
:wave: Hello, `Remix` team. My name is George, and I'm currently a Frontend Engineer on the Web Extension team of [Trust Wallet](https://trustwallet.com). In the last couple of months, we introduced significant updates to our [web extension](https://trustwallet.com/browser-extension) in line with our values to support impactful Web3 technologies; we would like to contribute by adding an option for users to connect through Trust Wallet provider. 

Our web extension has reached hundreds of thousands of downloads, and we keep expanding. Many users request to connect with the Trust Wallet extension by default and face issues when multiple wallets are installed in their browsers. Currently, there is only the "Injected Provider" option which utilizes `window.ethereum`. When multiple injected providers exist, each can replace the other by writing to `window.ethereum` resulting in non-deterministic behaviors. Eventually, [EIP-5749](https://eips.ethereum.org/EIPS/eip-5749) will be adopted, but until then, we need a workaround. Supporting multiple injected providers will require re-architecting some core parts of the app. This scope seems to be quite extensive and might require significant effort. However, we would like to propose a workaround by adding a new provider to the [`providerList`](https://github.com/ethereum/remix-project/blob/master/libs/remix-ui/run-tab/src/lib/reducers/runTab.ts#L122). This solution seems to work quite well, without require many additions to the codebase. 

The result will be adding a new "Trust Wallet Provider" option as environment for users.  

<img width="293" alt="Screenshot 2023-01-31 at 5 38 42 PM" src="https://user-images.githubusercontent.com/118829911/215806254-7a249a7d-230c-441a-8d3b-1590d13c10f3.png">

If TW is not installed or active, users will be notified accordingly.

<img width="1040" alt="Screenshot 2023-01-31 at 5 41 29 PM" src="https://user-images.githubusercontent.com/118829911/215806725-1a8928a7-257f-4150-8b51-48b6bb58a78d.png">

What are the next steps? We have completed the required development work, and our internal QA team is doing an assessment. We open this issue to let you know about the upcoming changes. We are going to create a PR soon with more information. 

We are open to talks and any questions you might have. 🙏

PR #3454, opened 2023-02-28
add trustwallet provider

yann300:
fix https://github.com/ethereum/remix-project/issues/3320


OfficialSupportt:
@yann300 We'd be happy to look into this for you 
kindly file a support ticket below 
https://forms.gle/CDeM1AXFjSBt7qVK8

ryestew:
Testing when I have both Metamask and Trust on and off.  I'm refreshing Remix between changing what extension is on or off.

- When I have both MetaMask and Trust install and activated I see in the Environment list the Trust Wallet but not Metamask.
image

- When I turn off metamask Is still see the same Environment list as above.

- When I turn off Trust Wallet, I see Metamask -like this - so this is good:
image

- When Metamask is off and I see this list and I choose Injected provider, the Trust Wallet comes up.  Which is good.
But the Account is not getting filled:
image
And when I go back and select Trust Wallet - injected provider - the account still does not get populated.



-

ryestew:
When only Metamask is active, you just see Injected Provider - Metamask (good)
When both Metamask & Trust are active, you only see Trust.  (no option for the other wallet )
When only Trust Wallet is active, I see only the Trust (good)


When only Argent X is active, you don't see any injected provider (I made a mistake and activated that by mistake)(does it connect to Remix?)(It should)
image

yann300:
from the issue: `When multiple injected providers exist, each can replace the other by writing to window.ethereum resulting in non-deterministic behaviors. Eventually, [EIP-5749](https://eips.ethereum.org/EIPS/eip-5749) will be adopted, but until then, we need a workaround`
So having a perfect integration might be complicated for now

> When both Metamask & Trust are active, you only see Trust. (So no option for the other wallet )(limited functionality)

In that case `window.ethereum` is the trust wallet and metamask is completely shadowed..

Let check the rest of your question in a call

ryestew (approved):

ioedeveloper (approved):

PR #3455, opened 2023-02-28
Migrate wallet connect

ioedeveloper:


apps/remix-ide/src/blockchain/blockchain.tsx
```diff
@@ -111,12 +111,20 @@ export class Blockchain extends Plugin {
         this._triggerEvent('networkStatus', [this.networkStatus])
       })
     })
+
+    this.on('walletconnect', 'chainChanged', () => {
```
bunsenstraat, line 115: your plugin is called wallet-connect

apps/remix-ide/src/blockchain/blockchain.tsx
```diff
@@ -111,12 +111,20 @@ export class Blockchain extends Plugin {
...
+        this._triggerEvent('networkStatus', [this.networkStatus])
+      })
+    })
   }
 
   onDeactivation () {
     this.active = false
     this.off('injected', 'chainChanged')
     this.off('injected-trustwallet', 'chainChanged')
+    this.off('walletconnect', 'chainChanged')
```
bunsenstraat, line 127: same here

libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -135,6 +135,12 @@ export const setupEvents = (plugin: RunTab, dispatch: React.Dispatch<any>) => {
     dispatch(fetchAccountsListSuccess(accountsMap))
   })
 
+  plugin.on('walletconnect', 'accountsChanged', (accounts: Array<string>) => {
```
bunsenstraat, line 138: wallet-connect

.gitignore
```diff
@@ -57,3 +57,7 @@ testem.log
 .DS_Store
 .vscode/settings.json
 .vscode/launch.json
+libs/remix-node/
```
bunsenstraat, line 60: this is some left over that seems to be passed around, these 3 can be removed

apps/remix-ide/src/remixAppManager.js
```diff
@@ -16,7 +16,7 @@ const requiredModules = [ // services + layout views + system views
 // dependentModules shouldn't be manually activated (e.g hardhat is activated by remixd)
 const dependentModules = ['foundry', 'hardhat', 'truffle', 'slither']
 
-const loadLocalPlugins = ["doc-gen", "doc-viewer", "etherscan", "vyper"]
+const loadLocalPlugins = ["doc-gen", "doc-viewer", "etherscan", "vyper", "walletconnect"]
```
bunsenstraat, line 19: These are actually directory names, not plugin names. And as the plugin name is not the directory name it won't load it. So we either change everything to walletconnect instead of mixing with wallet-connect or change this to the dir or rename this var so it's clear it's a directory.

.gitignore
```diff
@@ -57,3 +57,5 @@ testem.log
 .DS_Store
 .vscode/settings.json
 .vscode/launch.json
+
+apps/remix-react
```
yann300, line 61: should this be removed?

apps/remix-ide/src/app/udapp/run-tab.js
```diff
@@ -169,7 +169,7 @@ export class RunTab extends ViewPlugin {
     await addProvider('injected-optimism-provider', 'L2 - Optimism Provider', true, false)
     await addProvider('injected-arbitrum-one-provider', 'L2 - Arbitrum One Provider', true, false)
   
-    await addProvider('walletconnect', 'Wallet Connect', false, false)
+    await addProvider('walletconnect', 'WalletConnect', false, false)
```
yann300, line 172: this second parameter is a plugin name (see above) I think this should be `Wallet Connect`

.gitignore
ioedeveloper: Done.

apps/remix-ide/src/app/udapp/run-tab.js
ioedeveloper: See Andy's comment.

yann300:
could the plugin switch to a "light" theme when remix does the same?

yann300:
when sending a tx from metamask mobile, I've got:
`TypeError: undefined is not an object, evaluating 'E[W].balance'`

yann300:
I couldn't make it work with Trust mobile

PR #3477, opened 2023-03-01
fix using custom provider

yann300:
some provider returns the metadatas together with the result.

### DIFF
--- a/apps/remix-ide/src/app/providers/injected-provider.tsx
+++ b/apps/remix-ide/src/app/providers/injected-provider.tsx
@@ -61,8 +61,11 @@ export class InjectedProvider extends Plugin implements IProvider {
     }
     try {
       if ((window as any) && typeof (window as any).ethereum.request === "function") (window as any).ethereum.request({ method: "eth_requestAccounts" });
-      const resultData = await this.provider.currentProvider.send(data.method, data.params)
-      resolve({ jsonrpc: '2.0', result: resultData.result, id: data.id })
+      let resultData = await this.provider.currentProvider.send(data.method, data.params)
+      if (resultData && resultData.jsonrpc === '2.0') {
+        resultData = resultData.result
+      }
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/providers/injected-provider.tsx
```diff
@@ -61,8 +61,11 @@ export class InjectedProvider extends Plugin implements IProvider {
     }
     try {
       if ((window as any) && typeof (window as any).ethereum.request === "function") (window as any).ethereum.request({ method: "eth_requestAccounts" });
-      const resultData = await this.provider.currentProvider.send(data.method, data.params)
-      resolve({ jsonrpc: '2.0', result: resultData.result, id: data.id })
+      let resultData = await this.provider.currentProvider.send(data.method, data.params)
+      if (resultData && resultData.jsonrpc === '2.0') {
```
bunsenstraat, line 65: shouldn't you also check resultData.jsonrpc before you check the value?
resultData && resultData.jsonrpc && resultData.jsonrpc === '2.0'

yann300: ok sure, I don't think that change anything cause `jsonrpc` will be undefined then.

apps/remix-ide/src/app/providers/injected-provider.tsx
```diff
@@ -61,8 +61,11 @@ export class InjectedProvider extends Plugin implements IProvider {
...
+        resultData = resultData.result
+      }
+      resolve({ jsonrpc: '2.0', result: resultData, id: data.id })
```
bunsenstraat, line 68: also it's undefined resultData you still want to resolve?

bunsenstraat: or maybe it would never be undefined? then you would not need if (resultData

bunsenstraat (approved):

PR #3481, opened 2023-03-02
Decrease the setInterval timeout and use injected api to update accounts list and chain context

yann300:
In order to test it, run `ganache-cli` , the api calls are going to b printed.

The 3 type of configurations must be tested:

 - using VM (merge, or mainnet fork)
 - using an external provider (ganache-cli, external node, hardhat)
 - using injected provider (metamask, ..)


apps/remix-ide/src/blockchain/providers/injected.js
```diff
@@ -16,17 +16,13 @@ class InjectedProvider {
     })
   }
 
-  resetEnvironment () {
+  async resetEnvironment () {
```
ioedeveloper, line 19: This function does nothing. Should it still be kept?

apps/remix-ide/src/blockchain/providers/node.js
```diff
@@ -24,17 +24,13 @@ class NodeProvider {
     })
   }
 
-  resetEnvironment () {
+  async resetEnvironment () {
```
ioedeveloper, line 27: Same here also.

libs/remix-core-plugin/src/lib/compiler-fetch-and-compile.ts
```diff
@@ -134,7 +134,18 @@ export class FetchAndCompile extends Plugin {
...
       this.unresolvedAddresses.push(contractAddress)
-      return localCompilation()
+      const compilation = await localCompilation()
+      if (compilation) {
+        let found = false
+        compilation.visitContracts((contract) => {
+          found = util.compareByteCode(codeAtAddress, '0x' + contract.object.evm.deployedBytecode.object)
+          return found
+        })
+        if (found) {
```
ioedeveloper, line 144: This will always be `false` if the execution of `compilation.visitContracts` takes too long.

libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -159,7 +175,7 @@ const getCompiledContracts = (compiler) => {
   return contracts
 }
 
-export const resetAndInit = (plugin: RunTab) => {
+export const resetAndInit = async (plugin: RunTab) => {
```
ioedeveloper, line 178: async not needed here

apps/remix-ide/src/blockchain/providers/injected.js
yann300: yeah so it complies with the interface

libs/remix-core-plugin/src/lib/compiler-fetch-and-compile.ts
yann300: the execution is sync here.

ioedeveloper (approved):

PR #3504, opened 2023-03-07
Download Workspace

Aniket-Engg:
Related to #413 

It adds an option in workspace hamburger to download currently selected workspace

### DIFF
--- a/apps/remix-ide-e2e/src/tests/workspace.test.ts
+++ b/apps/remix-ide-e2e/src/tests/workspace.test.ts
@@ -415,7 +415,7 @@ module.exports = {
       })
       .click('//*[@id="workspacesMenuDropdown"]/span/i')
       .waitForElementVisible('//*[@id="workspacesMenuDropdown"]/div/ul')
-      .click('//*[@id="workspacesMenuDropdown"]/div/ul/a[4]') // rename workspace_name
+      .click('//*[@id="workspacesMenuDropdown"]/div/ul/a[2]') // rename workspace_name
       .useCss()
       .waitForElementVisible('*[data-id="treeViewLitreeViewItemtests"]')
       .waitForElementVisible('*[data-id="modalDialogCustomPromptTextRename"]')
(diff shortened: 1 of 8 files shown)


apps/remix-ide-e2e/src/tests/workspace.test.ts
```diff
@@ -415,7 +415,7 @@ module.exports = {
       })
       .click('//*[@id="workspacesMenuDropdown"]/span/i')
       .waitForElementVisible('//*[@id="workspacesMenuDropdown"]/div/ul')
-      .click('//*[@id="workspacesMenuDropdown"]/div/ul/a[4]') // rename workspace_name
+      .click('//*[@id="workspacesMenuDropdown"]/div/ul/a[3]') // rename workspace_name
```
LianaHus, line 418: Would be better to use data-id select instead.

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -156,6 +159,14 @@ export function Workspace () {
     }
   }
 
+  const onFinishDownloadWorkspace = async () => {
+    try {
+      await global.dispatchHandleDownloadWorkspace()
+    } catch (e) {
+      console.error(e)
```
LianaHus, line 166: do we need a proper error report in UI here?

apps/remix-ide-e2e/src/tests/workspace.test.ts
Aniket-Engg: This PR is just fixing the already present e2e. Similar e2e are there in complete file

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
Aniket-Engg: ok, will show some error on UI too

Aniket-Engg: done

apps/remix-ide-e2e/src/tests/workspace.test.ts
LianaHus: I know. I fixed them in another file. I think we should refactor this as well, otherwise, every time we add an item, this will fail.

bunsenstraat (approved):

PR #3519, opened 2023-03-13
UI fixes in UDAPP

LianaHus:
some refactoring to use bootstrap instead of extra elements
hover color for some icons shouldn't be warning. this is fixed
textarea of signed msg should have readable text-color


joeizang:
In the image below, text on the buttons for deployed contracts is truncated. Is there a fix for this?
image

joeizang (approved):

PR #3524, opened 2023-03-14
fix normalizePath

yann300:


joeizang (approved):

libs/remixd/src/services/hardhatClient.ts
```diff
@@ -54,7 +54,7 @@ export class HardhatClient extends PluginClient {
         const errMsg = '[Hardhat Compilation]: Cannot compile in read-only mode'
         return reject(new Error(errMsg))
       }
-      const cmd = `npx hardhat compile --config ${configPath}`
+      const cmd = utils.normalizePath(`./node_modules/.bin/hardhat compile --config ${utils.normalizePath(configPath)}`)
```
bunsenstraat, line 57: this will only work if node is in the path that electron has when running standalone. 
[Hardhat Compilation]: env: node: No such file or directory

bunsenstraat (changes requested):
this will only work if node is in the path that electron has when running standalone.
which it isn't by default, could be in brew, /usr/local/bin, anything, without the path of the user we won't know. On my system it says when I run this in a standalone remix desktop build
[Hardhat Compilation]: env: node: No such file or directory

bunsenstraat (approved):

PR #3532, opened 2023-03-16
Update github action yml templates

ioedeveloper:


libs/remix-ui/workspace/src/lib/utils/constants.ts
```diff
@@ -11,10 +11,17 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Run SUT Action
-      uses: EthereumRemix/sol-test@v1
+      uses: EthereumRemix/sol-test@v1.1
       with:
         test-path: 'tests'
         compiler-version: '0.8.15'
+//      evm-version: 'berlin'
```
yann300, line 18: can you set it to the latest one? I think it will be `berlin` or `merge`

libs/remix-ui/workspace/src/lib/utils/constants.ts
```diff
@@ -11,10 +11,17 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Run SUT Action
-      uses: EthereumRemix/sol-test@v1
+      uses: EthereumRemix/sol-test@v1.1
       with:
         test-path: 'tests'
         compiler-version: '0.8.15'
+//      evm-version: 'merge'
```
yann300, line 18: need to update the evm version

yann300 (changes requested):

Issue #166, opened 2020-07-28
Cannot select a blob of text to copy-paste-google-fu, as the click event is moving editor to the place where error occurred.


stefek99:

yann300:
we should have a copy button

PR #3533, opened 2023-03-16
Add copy to clipboard to warning message

yann300:
fix https://github.com/ethereum/remix-project/issues/166


Aniket-Engg (commented):
Screenshot 2023-03-20 at 8 15 43 PM

It works but I think position of copy icon should be on right-top and icon should be visible on hover

ryestew (approved):

PR #3577, opened 2023-03-29
Workspacefix

bunsenstraat:
bug fixes for the situation when:
- localstorage holds a currentworkspace key but indexeddb is empty OR the currentworkspace is not in the DB
  getAvailableWorkspaceName prevent .find in null, would cause errors in this case preventing creating a workspace
- localstorage holds a currentworkspace key but indexeddb is empty OR the currentworkspace is not in the DB
  it should also create default workspaces in this scenario, which it didn't causing a blank list of workspaces
- it makes no sense to call global.dispatchFetchWorkspaceDirectory(ROOT_PATH) when the workspace is NO_WORKSPACE, that would cause errors ( workspace.ts:273 Error: ENOENT: /.workspaces/null )

some e2e to test this scenario
added tracking because this scenario probably means indexeddb was cleared

this is a blocker because it prevents you from creating workspaces when indexeddb is cleared!

### DIFF
--- a/apps/remix-ide/src/app/panels/file-panel.js
+++ b/apps/remix-ide/src/app/panels/file-panel.js
@@ -95,6 +95,7 @@ module.exports = class Filepanel extends ViewPlugin {
   }
 
   getAvailableWorkspaceName (name) {
+    if(!this.workspaces) return null
     let index = 1
     let workspace = this.workspaces.find(workspace => workspace.name === name + ' - ' + index)
     while (workspace) {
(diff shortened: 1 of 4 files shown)


apps/remix-ide/src/app/panels/file-panel.js
```diff
@@ -95,6 +95,7 @@ module.exports = class Filepanel extends ViewPlugin {
   }
 
   getAvailableWorkspaceName (name) {
+    if(!this.workspaces) return null
```
yann300, line 98: pehaps this should just return `name + ' - 1 '`. because i m not sure how the caller of `getAvailableWorkspaceName` will react if it returns null.
or handle the case where it returns null.

PR #3585, opened 2023-03-30
Fix generating inherited functions

yann300:
- compile an openzeppelin ERC20 token
 - generate the documentation
 - the inherited function, their interface and documentation should be included in the md file.

remaining issues:
 - there is no visible separator between function and events (but if one copy the md to hack md, the separator is visible...)
 - params description aren't generated

### DIFF
--- a/apps/doc-gen/src/app/docgen/themes/markdown/contract.hbs
+++ b/apps/doc-gen/src/app/docgen/themes/markdown/contract.hbs
@@ -1,6 +1,6 @@
 {{>common}}
 
-{{h 2}} .Contract
+{{h 2}} Contract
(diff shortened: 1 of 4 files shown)


apps/doc-gen/src/app/docgen/themes/markdown/contract.hbs
```diff
@@ -1,8 +1,46 @@
 {{>common}}
 
-{{#each items}}
+{{h 2}} .Contract
```
bunsenstraat, line 3: Why do you have a . before Contract

bunsenstraat (approved):

bunsenstraat:
The issues mentioned are not issues. Params are visible and the delimiter also.

PR #3589, opened 2023-04-02
Docgen patch

joeizang:
Fix bug where contracts in any folder other than contracts won't have documentation generated for it.


apps/doc-gen/src/app/docgen-client.ts
```diff
@@ -45,19 +48,22 @@ export class DocGenClient extends PluginClient {
         input: input,
         output: output
       }
-      this.fileName = fileName
-      this.eventEmitter.emit('compilationFinished', this.build, fileName)
+      const test = normalizeContractPath(fileName, true)
+      this.fileName = typeof test === 'string' ? test : test[test.length - 1]
```
yann300, line 52: `this.fileName` in that case doesn't contain `.sol`. This shouldn't eb the case.

apps/doc-gen/src/app/App.tsx
```diff
@@ -30,7 +30,7 @@ const App = () => {
   return (
     <div className="p-3">
       <h3>Compile a solidity contract in order to build documentation as markdown.</h3>
-      {fileName && <h6>File: {fileName.split('/')[1].split('.')[0].concat('.sol')}</h6>}
+      {fileName && <h6>File: {fileName.concat('.sol')}</h6>}
```
yann300, line 33: why concat `.sol`. this doesn't look right

apps/doc-gen/src/app/docgen/utils/normalizeContractPath.ts
```diff
@@ -0,0 +1,17 @@
+
+export function normalizeContractPath(contractPath: string, returnPath?: boolean): string | string[]{
```
yann300, line 2: what do you want to do here?

joeizang: was catering for a situation where full file path is not desired.

apps/doc-gen/src/app/App.tsx
joeizang: I have since removed the concat. Thanks for pointing that out.

apps/doc-gen/src/app/docgen-client.ts
joeizang: I have since fixed this also.

apps/doc-gen/src/app/docgen-client.ts
```diff
@@ -45,19 +48,22 @@ export class DocGenClient extends PluginClient {
...
+      const segmentedPathList = normalizeContractPath(fileName)
+      this.fileName = segmentedPathList[segmentedPathList.length - 1]
+      this.contractPath =  segmentedPathList[0]
+      this.eventEmitter.emit('compilationFinished', this.build, this.fileName)
     })
   }
 
   async docgen(builds: Build[], userConfig?: Config): Promise<void> {
     const config = { ...defaults, ...userConfig }
+    config.sourcesDir = this.contractPath.length > 0 ? this.contractPath : config.sourcesDir
```
yann300, line 60: it seems `this.contractPath.length` will always be `> 0`, in which case do you think it isn't?

joeizang: Ah..you are right.

yann300 (approved):

PR #3590, opened 2023-04-02
Check for github action updates monthly

jlosito:
This will use dependabot to check any of the github actions that this project uses for updates and submit pull requests with version bumps.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


jlosito:
@ioedeveloper Out of curiosity, was there any reason why this was closed? Should something be changed?

ioedeveloper:
@jlosito I closed the PR because i couldn't see the importance of the PR to the project and also i couldn't find the testing instructions for the PR.

jlosito:
How is this project keeping its dependencies up-to-date then if not by dependabot or by some other means like renovate?

jlosito:
How is this project keeping its dependencies up-to-date then if not by dependabot or by some other means like renovate?

Issue #3588, opened 2023-03-31
Error "transact to Contract.function errored: NotFound" when using blockhash


niccolopetti:
Here is a minimal code to reproduce the error:

```
pragma solidity ^0.8.17;

contract A{

        function foo() public view returns(bytes32){

        return blockhash(block.number - block.basefee);
    }

}
```

calling function foo from remix gives
```
call to A.foo
call to A.foo errored: NotFound
```

Aniket-Engg:
![Screenshot 2023-04-03 at 7 37 29 PM](https://user-images.githubusercontent.com/30843294/229534147-936e8e07-5ee4-4a2b-b700-4f04c0ff5e1b.png)

This works fine for me.

niccolopetti:
Can you try deploying multiple instances and trying interacting again?

![image](https://user-images.githubusercontent.com/22686876/229578558-3c7370fc-aa22-4517-9830-d993cfaaca28.png)

I just tried deploying again and at first it worked, then deploying another instance and calling the function on the new instance reproduced the issue for me, 
Ps: I'm deploying to Remix VM (Merge) chain in case it matters for reproducing the error

Aniket-Engg:
Oh I see. I can reproduce it with second instance.

PR #3598, opened 2023-04-05
Fix save blocks ref

yann300:
fix https://github.com/ethereum/remix-project/issues/3588


libs/remix-simulator/src/vm-context.ts
```diff
@@ -181,14 +201,28 @@ export class VMContext {
     } else
       stateManager = new StateManagerCommonStorageDump()
 
-    const common = new Common({ chain: 'mainnet', hardfork })
-    const blockchain = new (Blockchain as any)({ common })
+    const consensusType = hardfork === 'berlin' || hardfork === 'london' ? ConsensusType.ProofOfWork : ConsensusType.ProofOfStake
```
ioedeveloper, line 204: Is this check sufficient? How about other hardforks like istanbul?

yann300: currently we instanciate only "berlin" and "london", so this is the last two on proof of stake fork we are still using.

ioedeveloper (approved):

PR #3601, opened 2023-04-06
Unwanted Event Listener removal for Docgen

joeizang:
Fixes the blanket removal of the 'compilationFinished' event listener. Not removing this causes the documentation generation to never actually happen.

### DIFF
--- a/apps/doc-gen/src/app/App.tsx
+++ b/apps/doc-gen/src/app/App.tsx
@@ -29,7 +29,7 @@ const App = () => {
 
   return (
     <div className="p-3">
-      <h5 className="h-5 mb-3">Compile a solidity contract in order to build documentation as markdown or right click on a contract and click on "Generate Documentation".</h5>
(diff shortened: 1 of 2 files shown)


apps/doc-gen/src/app/App.tsx
```diff
@@ -29,7 +29,7 @@ const App = () => {
 
   return (
     <div className="p-3">
-      <h5 className="h-5 mb-3">Compile a solidity contract in order to build documentation as markdown or right click on a contract and click on "Generate Documentation".</h5>
+      <h5 className="h-5 mb-3">Compile a Solidity contract and generate its documentation as Markdown. (Shortcut: right-click on a contract in the File Explorer and select "Generate Docs" from the context menu.).</h5>
```
Aniket-Engg, line 32: details you have added in not a `shortcut`

joeizang: If you check the dev-ux channel in discord, you will understand why the wording has changed.

Aniket-Engg (approved):

PR #3604, opened 2023-04-07
Setup Zoom On Mouse Hover

### LINKED ISSUE #3310: Improvements to Sol2UML integration
Some features that are nice to have but will also enhance usability are listed below.

- [ ] Add buttons that offer ability to download svg as a png image or a pdf
- [ ] Add image zoom on hover where any part of the svg hovered over magnifies it
- [ ] Add buttons to handle zooming and zoom reset
- [ ] Look into the possibility of interacting with the different parts of the uml diagram when clicked upon.

joeizang:
Fixes #3310


Aniket-Engg (commented):
I cannot see any magnification

joeizang:
> https://deploy-preview-3604--remixproject.netlify.app/

image
Not sure but it works fine from the netlify link: Here is a screenshot of the zoom

Aniket-Engg (commented):
Ok I see.

1.It is visible for an UML appearing full screen in editor. But doesn't work for Ballot contract in default workspace
2. If UML is zoomed in, magnification is highly displaced from cursor postion

joeizang:
> Ok I see.
> 
> 1.It is visible for an UML appearing full screen in editor. But doesn't work for Ballot contract in default workspace 2. If UML is zoomed in, magnification is highly displaced from cursor postion

@Aniket-Engg I have been digging into the documentation of the library we used and it zooms on when "On all components, the zoom functionality will be disabled if the large image size is <= to the small image's current rendered size..."
So I think those don't need magnification that's why it doesn't zoom the smaller contract uml diagrams. I will still keep digging into the docs of the library but I don't think its a problem per se.

ryestew:
It works fine for me and I think we should merge it. 
Nevertheless, I do have a UX point to make:

Typically when I want to scroll around a page I use the mouse scroll wheel.  But on the UML diagram, the mouse wheel zooms in or out.  Complicating this is the mouse pointer magnifier - which also zooms in.  So there are 2 ways of zooming - the Mouse pointer and the scroll wheel.  So its a little hard to scroll around the page - because you need to use the browser's scroll bar and you need to remember to do that too ( I instinctively forgot).  
All that said - think we should merge this.

ryestew (approved):

joeizang:
> It works fine for me and I think we should merge it. Nevertheless, I do have a UX point to make:
> 
> Typically when I want to scroll around a page I use the mouse scroll wheel. But on the UML diagram, the mouse wheel zooms in or out. Complicating this is the mouse pointer magnifier - which also zooms in. So there are 2 ways of zooming - the Mouse pointer and the scroll wheel. So its a little hard to scroll around the page - because you need to use the browser's scroll bar and you need to remember to do that too ( I instinctively forgot). All that said - think we should merge this.

I will make an issue of this and create a pr to address this.

Issue #3596, opened 2023-04-04
UI issues in Flatly theme


Aniket-Engg:
![Screenshot 2023-04-04 at 2 41 00 PM](https://user-images.githubusercontent.com/30843294/229820882-809ebdb2-05ef-42fd-9055-632542e65a52.png)

1. Text is not aligned
2. Icons are not visible

PR #3605, opened 2023-04-11
flatly theme color scheme improvements

LianaHus:
fix https://github.com/ethereum/remix-project/issues/3596


Aniket-Engg (commented):
Icons are visible so 2 is fixed. But text in 1 is still not properly aligned

Aniket-Engg (approved):

PR #3619, opened 2023-04-13
Support for scientific notation for function arguments

shafu0x:
Converts scientific notation in function arguments to big numbers
- uses https://github.com/PaulRBerg/evm-bn for the conversion


Aniket-Engg:
This is implemented by https://github.com/ethereum/remix-project/pull/3613

Issue #3628, opened 2023-04-18
Display a warning in udapp that the evmversion should match the current network fork.


yann300:
The warning should be displayed:
 - when the current network isn't "1" (mainnet)

The warning should state something like:
"Please make sure that the current network you deploy to is compatible with the evm version {evm version} . Otherwise any deployment will fail."

LianaHus:
I'm opening the issue to improve the UI

Working version:
![Screenshot from 2023-06-20 13-14-08](https://github.com/ethereum/remix-project/assets/9685356/80745423-7db6-46e3-b3ac-942373e28603)

@aniket suggested trying shorter names. To be discussed in UX meeting

Aniket-Engg:
closed by #3895

PR #3629, opened 2023-04-18
warn the user about evm version

yann300:
fix https://github.com/ethereum/remix-project/issues/3628

### DIFF
--- a/apps/remix-ide/src/app/tabs/locales/en/udapp.json
+++ b/apps/remix-ide/src/app/tabs/locales/en/udapp.json
@@ -5,6 +5,7 @@
   "udapp.value": "Value",
   "udapp.contract": "Contract",
   "udapp.compiledBy": "Compiled by {compilerName}",
+  "udapp.warningEvmVersion": "Please make sure that the current network is following the evm version: {evmVersion}. Otherwise any deployment will fail.",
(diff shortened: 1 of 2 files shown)


apps/remix-ide/src/app/tabs/locales/en/udapp.json
```diff
@@ -5,6 +5,7 @@
   "udapp.value": "Value",
   "udapp.contract": "Contract",
   "udapp.compiledBy": "Compiled by {compilerName}",
+  "udapp.warningEvmVersion": "Please make sure that the current network is following the evm version: {evmVersion}. Otherwise any deployment will fail.",
```
LianaHus, line 8: i think you don't need "that" here

LianaHus:
https://github.com/ethereum/remix-project/blob/0698464ec73b9df7430ac6d19a77ee2ee0816231/libs/remix-ui/run-tab/src/lib/css/run-tab.css#L184 could you please remove this line? to align those badges

LianaHus:
and I don't know why the mouse pointer is not working on hover for only this badge.

LianaHus (changes requested):
see comments.

yann300:
> Can selecting evm in solidity compiler help guarantee compatibility in run and deploy?

yes, but one need to make sure the right evm version is selected in "run and deploy"

> If it can then maybe we might not need the one of the badges (the one that says compiled with Remix)

Compilation can also come from other sources (hardhat, ..)

joeizang (approved):
LGTM

PR #3642, opened 2023-04-21
Solhint linter plugin

joeizang:
Sets up the initial grunt work for solhint plugin. Other UI based prs will follow


apps/solhint/src/profile.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "0.1.0",
+  "events": ["lintOnCompilationFinished", "lintingFinished"],
+  "methods": [
+    "lintWithoutCompilationCustomAction",
+    "lintOnCompilation",
+    "lintContractCustomAction"
+  ],
+  "kind": "none",
+  "icon": "https://raw.githubusercontent.com/protofire/solhint/master/solhint-icon.png",  
+  "location": "sidePanel",
```
bunsenstraat, line 14: if you don't want UI you can set this to hiddenPanel

apps/solhint/src/app/SolhintPluginClient.ts
```diff
@@ -0,0 +1,145 @@
...
+        formattedMessage: `${report.message}\n${report.fix ? report.fix : ''}`,
+        type: this.severity[report.severity] || 'error',
+        column: report.column,
+        line: report.line - 1
+      }
+    })
+    return hints
+  }
+
+  severity = {
```
bunsenstraat, line 98: const?

apps/solhint/src/app/SolhintPluginClient.ts
```diff
@@ -0,0 +1,145 @@
...
+    2: 'error',
+    3: 'warning'
+  }
+  
+  rules = {
```
bunsenstraat, line 103: const

apps/solhint/src/app/SolhintPluginClient.ts
```diff
@@ -0,0 +1,145 @@
+import { PluginClient } from '@remixproject/plugin'
+import { createClient } from '@remixproject/plugin-webview'
+import EventEmitter from 'events'
+import { processStr } from 'solhint'
+import { applyExtends } from 'solhint/lib/config/config-file'
```
bunsenstraat, line 5: isn't everything properly exported?

apps/solhint/src/app/App.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { SolHint } from "./SolhintPluginClient"
+
+
+const client = new SolHint()
+
+export default function App() {
+
+  return (
+    <>
+      <div className="">
```
bunsenstraat, line 11: no div needed right

apps/solhint/src/app/SolhintPluginClient.ts
joeizang: they are. Had a resolution issue but it appears to be non existent.

apps/solhint/src/app/SolhintPluginClient.ts
```diff
@@ -0,0 +1,145 @@
...
+  "extends": "solhint:recommended",
+  "plugins": [],
+  "rules": {
+    "avoid-suicide": "error",
+    "avoid-sha3": "warn"
+  }
+}`
+
+export class SolHint extends PluginClient {
+  mdFile: string
```
bunsenstraat, line 33: unused

apps/solhint/src/app/App.tsx
joeizang: yeah

apps/solhint/src/app/SolhintPluginClient.ts
```diff
@@ -0,0 +1,145 @@
...
+  eventEmitter: EventEmitter
```
bunsenstraat, line 34: unused

apps/solhint/src/app/SolhintPluginClient.ts
joeizang: yeah

apps/solhint/src/app/SolhintPluginClient.ts
joeizang: true

apps/solhint/src/app/SolhintPluginClient.ts
joeizang: they are class properties so const wouldn't work here

apps/solhint/src/app/SolhintPluginClient.ts
joeizang: class properties

apps/solhint/src/profile.json
joeizang: true. Mistaken here.

yann300:
- unchecking `Basic` is disabling the  `Analyze` button.
 - `Basic` don't show any result for the Storage contract. should perhaps show "no result" if this is the case.
 - slither can be checked even if not connected to remixd and cannot be unchecked.
 - slither actually starts running even if remixd isn't connected.
 - when hovering a warning it should probably display a link cursor rather than the normal one.

ryestew (changes requested):
1. I just checked to make sure that slither is working in your PR.  I don't see the warnings in the Slither tab - although I see in the terminal that there are 385 or so warnings from Slither.  I used the remixd command to connect to the Remix instance - remixd -s ./ --remix-ide https://deploy-preview-3642--remixproject.netlify.app

2. The dot (.) should be taken out of the icon next to static analysis.
3. I'm not sure why the Linter has a different color badge than Basic in the tabs.

image

4. The tooltips seem to have an extra leading space.  They are also complete sentences - so they need a capital letter to start and a period at the end.

image


Also in the next PR for Slither update, it probably needs a spinner - to show that Slither is working - because slither is slower to produce the results.

ryestew (changes requested):
I connect with Remixd and analyse with Slither.  I get this but in the Remixd terminal there are no errors.  So I"m not sure there is an error - but it is telling me there is one.

image

And the tooltip in the Analysis button should be:
Run analysis (lowercase a in analysis)

yann300:
the autorun seems on, is that wanted?

yann300:
when clicking on slither and if remixd isn't activated, we shouldn't show the modal to connect to remix.
this is too much of a disruption cause the use doesn't know that this will switch the workspace.

yann300:
clicking on slither result doesn't highlight the code editor.

yann300:
`Show warnings for external libraries` doesn't seems to work (see slither result)

yann300:
loading slither results 2 times in a row is slowing down the UI.

yann300:
after runner slither compilation, I've got these error in the remixd terminal:
`rytic_compile.platform.exceptions.InvalidCompilation: Invalid solc compilation Invalid option to --combined-json: compact-format
`
perhaps @Aniket-Engg has some ideas.

yann300:
somehow I can't uncheck slither when it's checked.

ryestew (changes requested):
The tool tips seem like they have a blank space before the first letter.
Like this one: 
image

Also this text should be updated to:
To run Slither analysis, Remix IDE must be connected to your local filesystem with remixd.

Before a file is compiled the disabled button currently says:
image

It should be updated for the case when there is no file compiled:
To run analysis tools, first compile a file.

The placement of this tooltip is far from the text.
image
But I'm not sure its needed - except that I'm not sure what general warnings means - so if that is something special - we should keep this tooltip.  Otherwise, it not adding info so it could be removed.

Similarly - does the tooltip for show warnings for external libraries add anything?  I don't think so - so I think it can be removed. 

The slither tab will get removed if slither is not activated in another PR right?

ryestew (changes requested):
When connected with Remixd , slither can take a few minutes to run.  This is clear if one looks in the terminal, but if not you just end up hitting the analyse button a lot.  I'd recommend putting a spinner in the tab before the slither results return.

Actually, I am not getting any slither results returned...that's odd.

ryestew (approved):

PR #3648, opened 2023-04-25
CI: Upgrade from node.js 14 to node.js 16.

paulmillr:
Node.js 14 will be unsupported and cease to receive security updates in 6 days. It is necessary to upgrade. https://github.com/nodejs/Release, https://endoflife.date/nodejs


bunsenstraat:
https://github.com/ethereum/remix-project/pull/3649

PR #3671, opened 2023-05-01
Fix layout shift in explorer

himanshupal:
This commit tries to fix the slightly annoying layout shift in explorer tree, applied fix can be seen in video attached, & can be compared visually to current production webapp

https://user-images.githubusercontent.com/40417771/235526209-4d55bd59-3987-429a-868c-4cb0c9554eb8.mp4


bunsenstraat:
seems good to me. @LianaHus any thoughts. @ryestew ?

LianaHus (approved):

Issue #3651, opened 2023-04-26
Editor: breakpoints appearing when pasting


ryestew:
Often (but not always) when I paste code into an editor tab I get a bunch of breakpoints added.  When this happens the top one usually doesn't turn off when clicked.
![image](https://user-images.githubusercontent.com/6050001/234445124-23fa893b-ddce-4526-aa50-69d5df4b917a.png)

PR #3675, opened 2023-05-03
fix adding breakpoint

yann300:
fix https://github.com/ethereum/remix-project/issues/3651


ryestew (commented):
it would be better if the breakpoints are added by a click event and not by a mouseup event.
Still this version is an improvement.

ryestew (approved):

PR #3706, opened 2023-05-16
check if there are constructor params

Aniket-Engg:
Show `Constructor Arguments` textbox only if there are params in constructor method


yann300:
we should probably add a way for the user to provided decoded value

PR #3707, opened 2023-05-16
Add 'web3 scripts' (to FE context menu)

yann300:
add scripts for:
 - etherscan
 - contract deployer (create2)


yann300:
some more insight about this:
Another solution would be to add these helper scripts directly to the templates.

apps/remix-ide/src/app/tabs/locales/en/filePanel.json
```diff
@@ -29,6 +29,8 @@
   "filePanel.tssoltestghaction": "Mocha Chai Test Workflow",
   "filePanel.workspace.slitherghaction": "Adds a preset yml file to run slither analysis on github actions CI",
   "filePanel.slitherghaction": "Slither Workflow",
+  "filePanel.workspace.helperscripts": "Adds convenient scripts to the 'scripts' directory",
```
ioedeveloper, line 32: Can you also include the translation for it?

yann300: yes I've added the translation label.
cc @ryestew

yann300:
after discussion, we should have sub menus for this.

joeizang (approved):
LGTM

PR #3715, opened 2023-05-18
chain id in etherscan script

### LINKED ISSUE #3712: etherscan: improve the verify script
This script currently accept an api key but it doesn't accept the url endpoint, this should probably be the case.

Aniket-Engg:
Fixes #3712


yann300:
If we don't want to complexify the UI, we should probably add a note somewhere that it is preferable to use the script for  for instance verifying in multiple network at the same time.

Issue #3705, opened 2023-05-15
WalletConnect is not updating Deploy & Run


ryestew:
Even though this was tested - I am now seeing a problem - walletConnect on my phone and in the plugin is not that same as what is showing in Deploy&Run.

<img width="1136" alt="image" src="https://github.com/ethereum/remix-project/assets/6050001/6843db72-3c9e-4f21-bf73-6f0fe9a4cdc2">

but after I change the network on the phone a few times, it syncs.

yann300:
you mean the balance isn't the same? I think the balance get updated every 20s , so this might take time.

ryestew:
not the balance - but that WalletConnect is on goerli and Deploy & Run Environment shows Mainnet

PR #3716, opened 2023-05-18
Fix wallet connect

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/3705


yann300 (approved):

ryestew (commented):
Still problems.  To see the problem:
1. connect Remix with injected provider - metamask and then choose goerli.
2. open up the mobile wallet and set it to polygon
3. switch the environment to WalletConnect and connect the mobile wallet.
4. The badge in Deploy and Run should switch to polygon but it doesn't.
5. but connect with the mobile wallet and then switch back to injected provider and the inject provider is not updating the site correctly.

PR #3718, opened 2023-05-18
Fix Solidity Compiler Version Issue in Webpack

dreamqip:
This pull request addresses a compatibility issue between the Solidity compiler and Webpack. Currently, the Solidity compiler version used in the project package remix-solidity is incompatible, resulting in compilation errors during the build process.

To resolve this issue, the pull request includes the following changes:

Update the Solidity compiler version to the compatible version.


Aniket-Engg:
It will be further tracked under https://github.com/ethereum/remix-project/issues/5080 . So closing it

Issue #3597, opened 2023-04-04
Etherscan Plugin Improvement and fixes


Aniket-Engg:
As per current status:
- [x] Nothing happens on clicking icons if no API key provided. Actually that is a feature but user has no idea what to do next.
- [x] After providing API key, clicking on Save, no message appears.
- [x] No check on API key length

**Home Tab:**
- [x] Form validation improvement, unnecessary cross icon
- [x] Show constructor arguments field conditionally if required
- [x] Show error in red. 
- [x] Show selected network in the plugin itself in read-only text

**Receipts:**
- [x] Requests seems to pending in queue forever

**Settings:**
- [x] Show API key in dots

yann300:
- Unable to verify contracts which has openzeppelin imports.
 - Unable to verify contracts that uses proxy.

LianaHus:
- [ ] titles
- [ ] font awsome icons instead of svg's

Aniket-Engg:
- [ ] Add a user-friendly way for encoded constructor arguments

PR #3750, opened 2023-05-31
Enable proxy contract verification in Etherscan plugin

Aniket-Engg:
Related to https://github.com/ethereum/remix-project/issues/3597#issuecomment-1497446036

This PR integrates the proxy contract verification and proxy contract verification receipt status API from Etherscan. Here is the related documentation: https://docs.etherscan.io/api-endpoints/contracts#verify-proxy-contract

As verification on Etherscan integrate proxy contract with its implementation, the response we get is like this:
```
// OK
{"status":"1","message":"OK","result":"The proxy's (0xbc46363a7669f6e12353fa95bb067aead3675c29) implementation contract is found at 0xe45a5176bc0f2c1198e2451c4e4501d4ed9b65a6 and is successfully updated."}
```

So on receipts status page, we show status as `Successfully updated` along with an info icon, hovering on it shows complete message on tooltip.

Screenshot 2023-06-07 at 12 13 47 PM

### DIFF
--- a/apps/etherscan/src/app/utils/verify.ts
+++ b/apps/etherscan/src/app/utils/verify.ts
@@ -121,6 +121,10 @@ export const verify = async (
             apiKeyParam,
             etherscanApi
           )
+          if (receiptStatus.status === '1') {
+            receiptStatus.message = receiptStatus.result
+            receiptStatus.result = 'Successfully Updated'
+          }
         } else receiptStatus = await getReceiptStatus(
           result,
           apiKeyParam,
--- a/apps/etherscan/src/app/utils/verify.ts
+++ b/apps/etherscan/src/app/utils/verify.ts
@@ -21,7 +21,9 @@ export const verify = async (
     contractArgumentsParam: string,
     contractName: string,
     compilationResultParam: CompilerAbstract,
-    chainRef: number | string, 
+    chainRef: number | string,
+    isProxyContract: boolean,
+    expectedImplAddress: string, 
     client: PluginClient,
     onVerifiedContract: (value: EtherScanReturn) => void,
     setResults: (value: string) => void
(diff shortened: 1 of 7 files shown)


apps/etherscan/src/app/utils/verify.ts
```diff
@@ -85,13 +87,20 @@ export const verify = async (
...
-        contractaddress: contractAddress, // Contract Address starts with 0x...
         sourceCode: JSON.stringify(jsonInput),
         contractname: fileName + ':' + contractName,
         compilerversion: `v${contractMetadataParsed.compiler.version}`, // see http://etherscan.io/solcversions for list of support versions
         constructorArguements: contractArgumentsParam ? contractArgumentsParam.replace('0x', '') : '', // if applicable
       }
 
+      if (isProxyContract) {
+        data.action = "verifyproxycontract"
+        data.expectedimplementation = expectedImplAddress
```
yann300, line 98: you don't seem to use that `expectedimplementation` anywhere, what is that for?

Aniket-Engg: It is an optional parameter and used in the API. See: https://docs.etherscan.io/api-endpoints/contracts#verify-proxy-contract

yann300: ok where is it used for the api call?

yann300:
- [ ] I think the `it's a proxy contract` should also contain the information that the impl contract should be verified first.
 - [ ] Could we also add a script which verify proxy contract?

Aniket-Engg:
> * [ ]  I think the `it's a proxy contract` should also contain the information that the impl contract should be verified first.
> * [ ]  Could we also add a script which verify proxy contract?

--- I have added an info icon with tooltip to explain first point
For second point, current scripts can be extended to support proxy contract verification

yann300 (approved):

yann300:
still missing extending the script to allow proxy verification, this will be done in a next pr.

Issue #3755, opened 2023-06-01
Badges in the icon panel should not have a tooltip


ryestew:
Having 2 tooltips on the icon panel makes it too noisy.  The plugin icon tooltip is the essential one and the badge tooltip is self evident what it means, so it should be removed.

yann300:
Removing the tooltip isn't something everyone agree and we can find a better solution, this can be done in 0.34.1

PR #3779, opened 2023-06-09
Static Analysis Checkbox Fix

### LINKED ISSUE #3781: Linting problem
When using the right click menu to lint the contract, the contract compiles but the static analysis plugin is not visible.
(Even if it was - a user would still need to know to go there to see the results - but that's OK for now)(address this somehow later).

And when I go to the Static Analysis plugin - it says I need to compile the file - so its not getting the compilation result from the right click lint.

joeizang:
fixes #3781 #3755

### DIFF
--- a/libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
+++ b/libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
@@ -116,13 +116,6 @@ export const RemixUiStaticAnalyser = (props: RemixUiStaticAnalyserProps) => {
       await run(state.data, state.source, state.file, state, props, isSupportedVersion, showSlither, categoryIndex, groupedModules, runner,_paq, message, showWarnings, allWarnings, warningContainer,calculateWarningStateEntries, warningState, setHints, hints, setSlitherWarnings, setSsaWarnings, slitherEnabled, setStartAnalysis)
     }
     props.event.trigger('staticAnaysisWarning', [])
-    // if (basicEnabled) {
-    //   if (state.data !== null) {
-    //     runAnalysis().catch(console.error);
-    //   }
-    // } else {
-    //   props.event.trigger('staticAnaysisWarning', [])
-    // }
     return () => { }
   }, [state])
 
@@ -387,13 +380,6 @@ export const RemixUiStaticAnalyser = (props: RemixUiStaticAnalyserProps) => {
   const hintErrors = hints.filter(hint => hint.type === 'error')
   const noLibSlitherWarnings = slitherWarnings.filter(w => !w.options.isLibrary)
   const slitherErrors = noLibSlitherWarnings.filter(slitherError => slitherError.options.type === 'error')
-  // const noLibsRemixWarnings = ssaWarnings.filter(w => !w.options.isLibrary)
-  // const groupedWarnings = ssaWarnings.reduce((resultingObject, warning) => {
-  //   if (!resultingObject[warning.warningModuleName]) resultingObject[warning.warningModuleName] = []
-  //   resultingObject[warning.warningModuleName].push(warning)
-  //   return resultingObject
-  // }, {})
-  // console.log({ slitherWarnings, noLibsRemixWarnings, ssaWarnings, groupedWarnings })
 
   const tabKeys = [
     {
--- a/apps/remix-ide/src/app/tabs/analysis-tab.js
+++ b/apps/remix-ide/src/app/tabs/analysis-tab.js
@@ -30,7 +30,6 @@ class AnalysisTab extends ViewPlugin {
     this.registry = Registry.getInstance()
     this.element = document.createElement('div')
     this.element.setAttribute('id', 'staticAnalyserView')
-    this.errorCount = 0
     this._components = {}
     this._components.registry = this.registry
     this._deps = {


ryestew (changes requested):
This PR should be fixing the problem where the static analysis checkbox for Slither:
1. should be deactivate when Slither is not connected through Remixd 
2. the checkbox's label too should be disabled (not as bright as the other labels)
3. the tooltip should say why its deactivated

But as I see it now when Slther is not connected
1. The checkbox is not deactivated
2. The label is the same intensity as the others
3. The tooltip is the one for when the Slither is connected

Note: In the current alpha - the label is disabled when slither is not connected and the tooltip is correct but the checkbox itself is active.

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
```diff
@@ -629,29 +635,29 @@ export const RemixUiStaticAnalyser = (props: RemixUiStaticAnalyserProps) => {
...
             onClick={handleSlitherEnabled}
-            checked={showSlither}
-            disabled={true}
+            checked={showSlither && slitherEnabled}
+            disabled={slitherEnabled}
             label="Slither"
             onChange={() => {}}
             optionalClassName="mr-3"
-            title="To run Slither analysis, Remix IDE must be connected to your local filesystem with remixd."
+            title={slitherEnabled ? "Run Slither static analysis." : "To have slither active, you need to have remixd connected to your local filesystem"}
```
ryestew, line 660: Update to:
To run Slither analysis, Remix IDE must be connected to your local filesystem with Remixd.

ryestew (changes requested):
- [ ] **Disabled Slither is not correct yet**
In the current Remix-alpha, the disabled state looks like this:
image
The label looks like it is in the disabled state. Some adjustments are needed on your PR to give it the same look.

In the disabled state, the Slither tab should not show - even when the Slither checkbox is attempted to be clicked on - in its disabled state. So this image is not correct:
image

- [ ] **Turn off tooltip on the analyse button when the file is compiled:**
image

- [x] **Last Results for - doesn't make sense before you click the Analyze button.**
Not sure what to do about this one - maybe this can be a simple text fix:  maybe we don't need Last results for at all - just printing out the name of the file will better.

- [x] **Remix shows badge with 0 before the Analysis has been run**
Before the Analyze button is clicked, the Remix tab shows the results (zero) and the Linter is blank
image
Badge should not display until the analysis has been run. So the linter badge is correct and the Remix badge needs an update.

- [ ] **Slither tries to run even when disabled**
When I click on the slither checkbox it does not check because it is disabled.  Then I run analysis in in the terminal, it looks like it is trying to run Slither:
image

- [ ] **When Remixd is connected**
When the Slither checkbox can be checked, the text is in the disabled state (so currently the checked state has the text as disabled and the unchecked state has the text has enabled)
image

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
joeizang: done

ryestew:
A very small update for the Slither tooltip when remixd is connected:
Slither runs Slither static analysis.

and to be consistant, please update Linter to:
Linter runs SolHint static analysis.

ryestew (changes requested):
- [ ] Before the analyse button has been hit, there is the text that says **No Results.**  There are no results because the tool hasnt' run yet - so I think the text is more confusing than helpful. (see below)
- [x] There is too much space below the Analyze button.

image


- [x] there is too much padding below the contract name.
image


- [ ] A very small update for the Slither tooltip when remixd is connected:
Slither runs Slither static analysis.

- [ ] and to be consistant, please update Linter to:
Linter runs SolHint static analysis.

- [ ] When I ran the static anaylsis tool, Linter finished first, then Slither and last came Remix.  Before the results came in, Remix showed **no results** (see the image below).  As soon as the results came in the badge populated ans so did the cards.  I think **No Results** is a legacy that can be removed.  I'm not sure why Remix analysis come in slower than Slither - or if it makes any difference which order things come in - but I was surprised that the Remix one took a while.  So this item may have nothing to change.  But please check if Remix analysis seems to take longer than you'd expect on your machine.

image

ryestew:
I'm still getting the Slither error - but maybe this is not the PR to address that.
image

ryestew (changes requested):
- [ ] The tooltip on the Analyse button should be on the button and not just on the text -which before anything is compiled is just the word Analyse.

ryestew (approved):

libs/remix-ui/static-analyser/src/lib/remix-ui-static-analyser.tsx
```diff
@@ -379,7 +385,15 @@ export const RemixUiStaticAnalyser = (props: RemixUiStaticAnalyserProps) => {
   }
 
   const hintErrors = hints.filter(hint => hint.type === 'error')
-  const slitherErrors = slitherWarnings.filter(slitherError => slitherError.options.type === 'error')
+  const noLibSlitherWarnings = slitherWarnings.filter(w => !w.options.isLibrary)
+  const slitherErrors = noLibSlitherWarnings.filter(slitherError => slitherError.options.type === 'error')
+  // const noLibsRemixWarnings = ssaWarnings.filter(w => !w.options.isLibrary)
```
yann300, line 390: could you remove it?

joeizang: done

PR #3796, opened 2023-06-13
fix vm fork

yann300:
rewrite of the CustomEthersStateManager.
the previous implementation didn't handle well the checkpoint/revert.

in order to test it:

 - compile
```
contract Campaign {
    uint public cake;
    function contribute() public payable {
       cake++;
    }
}
```

- deploy and call contribute
- access "cake", this should print "1"
- call contribute again
- access "cake", this should print "2"


ryestew (approved):

libs/remix-simulator/src/vm-context.ts
```diff
@@ -124,6 +88,158 @@ class StateManagerCommonStorageDump extends DefaultStateManager {
...
+  blockTag: bigint | 'earliest',
+  /**
+   * A {@link Trie} instance
+   */
+  trie?: Trie
+}
+
+class CustomEthersStateManager extends StateManagerCommonStorageDump {
+  private provider: ethers.providers.StaticJsonRpcProvider | ethers.providers.JsonRpcProvider
+  private blockTag: string
```
ioedeveloper, line 102: `blockTag` type definition (`string`) does not match it's type definition on line 93 (`bigint | 'earliest'`)

libs/remix-simulator/src/vm-context.ts
```diff
@@ -124,6 +88,158 @@ class StateManagerCommonStorageDump extends DefaultStateManager {
...
+
+  constructor(opts: CustomEthersStateManagerOpts) {
+    super(opts)
+    if (typeof opts.provider === 'string') {
+      this.provider = new ethers.providers.StaticJsonRpcProvider(opts.provider)
+    } else if (opts.provider instanceof ethers.providers.JsonRpcProvider) {
+      this.provider = opts.provider
+    } else {
+      throw new Error(`valid JsonRpcProvider or url required; got ${opts.provider}`)
```
ioedeveloper, line 111: `provider` can be of 3 types `string | ethers.providers.StaticJsonRpcProvider | ethers.providers.JsonRpcProvider`. Shouldn't there be a third check for `else if (opts.provider instanceof ethers.providers.StaticJsonRpcProvider)` before the else statement?

ioedeveloper (approved):

PR #3802, opened 2023-06-15
feat: support iframe, and keep the pace with parent if ide in a iframe

fwx5618177:
To: to support iframe.

Desc:
- If we want to embed ide in a iframe, and we want to keep the pace with parent window's ethereum provider.
- If in a cross-origin env, it couldn't support to share the ethereum.

Result:
- Could get the correct address and provider from the parent iframe in the same origin.

### DIFF
--- a/apps/remix-ide/src/blockchain/execution-context.js
+++ b/apps/remix-ide/src/blockchain/execution-context.js
@@ -8,7 +8,18 @@ const _paq = window._paq = window._paq || []
 let web3
 
 if (typeof window !== 'undefined' && typeof window.ethereum !== 'undefined') {
-  var injectedProvider = window.ethereum
+  // To support iframe issue.
+  // if we want to embed ide in a iframe, and we want to keep the pace with parent window's ethereum provider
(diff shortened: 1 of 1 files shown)


Aniket-Engg:
@fwx5618177 Thanks for the PR. Would you like to share the motivation behind the PR?

fwx5618177:
> @fwx5618177 Thanks for the PR. Would you like to share the motivation behind the PR?

Action: Managing Multiple Wallets in iFrame and External Environments

Description:
When incorporating an IDE within an iFrame while coexisting with an external environment, it's important to note that there can be two distinct wallets connected. This occurs due to the separation between the iFrame's wallet and the external wallet.

Issue:
In such scenarios, we may encounter the connection of two different wallets: one associated with the iFrame and another with the external environment. Each wallet operates independently in its respective context.

yann300:
Closing this as it's been quite some time, thanks for the patience and the PR.
We are going to beta test the desktop version soon, do you want to be part of it?
Thanks.

PR #3807, opened 2023-06-16
improvements in pride theme and fixing an issue in SUT

LianaHus:
The btns in SUT had a wrong html structure but the issue was visible on some themes only


Aniket-Engg (commented):
Screenshot 2023-06-19 at 3 57 08 PM

To me SUT buttons look like this, which seems bad.

joeizang (changes requested):
image

I think the choice of black text and icons for primary buttons is an issue because pride is a light theme. Primary colors should contrast with light text and icons. See the image below from cerulean.

image

joeizang:
Oh and the thick border around the vertical icons panel cuts into the space that badges should take. See image above

joeizang (approved):

Issue #3813, opened 2023-06-19
Etherscan issues part 2


LianaHus:
part 1 is here https://github.com/ethereum/remix-project/issues/3597

- [ ] To keep this consistency with other parts of the ide we need to use in-place value verification(for example 'required' error msg if the address is empty) instead of toasters ( we are minimizing the usage of toasters through the app.)
- [x] When VM is selected in D&R we need to disable the Verify button and the label should be 'VM - not supported" in the env field
- [x] The active icon from the header menu of the plugin should be brighter.

ryestew:
There is the api key input in the plugin and also in the settings panel.  When is the one in the settings panel used?

PR #3840, opened 2023-06-21
disable verify for not supported VM

Aniket-Engg:
Related to #3813


LianaHus:
I thought we should have a tooltip on verify btn when it is disabled, it is a good practice to give a hint on disabled btn.
Is it possible to add this?

LianaHus (changes requested):
missing a tooltip on verify btn when it is disabled

Aniket-Engg:
See, tooltip is there: https://github.com/ethereum/remix-project/blob/master/apps/etherscan/src/app/components/SubmitButton.tsx#L19 but it is not easily visible. May be needs some improvement

LianaHus (approved):

PR #3850, opened 2023-06-26
call gpt service

yann300:


apps/remix-ide/src/app/plugins/openaigpt.tsx
```diff
@@ -0,0 +1,37 @@
...
+        method: 'POST',
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({ prompt }),
+      })
+    ).json()
+
+    console.log(result)
```
bunsenstraat, line 33: remove

apps/remix-ide/src/app/plugins/openaigpt.tsx
```diff
@@ -0,0 +1,37 @@
...
+    this.call('terminal', 'log', { type: 'typewritersuccess', value: result.choices[0].message.content })
```
bunsenstraat, line 34: maybe check if the result has choices etc?

bunsenstraat:
when terminal is minimized it doesn't open it.

yann300:
> when terminal is minimized it doesn't open it.

not sure I get what you mean...
I pushed the other comments.

bunsenstraat:
> > when terminal is minimized it doesn't open it.
> 
> not sure I get what you mean... I pushed the other comments.

When you minimise the terminal and then call the gpt the results won't show unless you open the terminal again. would be confusing.

yann300:
> > > when terminal is minimized it doesn't open it.
> > 
> > 
> > not sure I get what you mean... I pushed the other comments.
> 
> When you minimise the terminal and then call the gpt the results won't show unless you open the terminal again. would be confusing.

I just pushed a fix.

Aniket-Engg (approved):
This works fine. 

But I think there should be improvement in UX to hightlight `ASK GPT` properly in compilation error card. That can come in a different PR.

bunsenstraat (approved):

ryestew:
Documentation is needed both in the Remix-ide docs and the docs for plugins:
it can be called from a plugin using plugin.call('openaigpt', 'message', message)

PR #3921, opened 2023-07-17
SSA-QA-Patch

joeizang:
Fixes issue raised by @LianaHus in qa doc where tabs don't show correctly based on checkbox states


Aniket-Engg (approved):
LGTM

PR #3939, opened 2023-07-23
Eslint 2 spaces fix

bunsenstraat:
We have incosistent indentation in the project, some files use 1 space, some 2, some 4.
Most commonly used throughout the project is 2.
sets all indentation on code files ( ts, tsx, js, jsx ) to 2 spaces
created a lint-fix-targets.js script to fix it
set eslint rule to enforce

if a change is needed, just change the eslint base file and run node lint-fix-targets.js


joeizang (approved):
We should probably add a prettierrc file to set and take advantage of prettier to auto format code files and husky to ensure it is linting. But great work @bunsenstraat

LianaHus (approved):

PR #3947, opened 2023-07-25
Sol2uml Theme Reaction Refactor

### LINKED ISSUE #3832: Sol2Uml styling improvements
The current setup for styling generated uml diagrams is dependent on a hand rolled mapping of values from the css file of every theme in the ide. To improve this so it picks values dynamically from set themes some changes are required.

- [x] ThemeModule plugin should push a payload along with the theme changed event.
- [x] Look into how best to have a global config of current themes and their values  so that plugins can access them
- [x] Use css variables dynamically as values for the uml diagram

joeizang:
fixes #3832

### DIFF
--- a/libs/remix-ui/solidity-uml-gen/src/types/index.ts
+++ b/libs/remix-ui/solidity-uml-gen/src/types/index.ts
@@ -16,7 +16,6 @@ export interface ISolidityUmlGen extends ViewPlugin {
     showUmlDiagram(path: string, svgPayload: string): void
     updateComponent(state: any): JSX.Element
     setDispatch(dispatch: React.Dispatch<any>): void
-    mangleSvgPayload(svgPayload: string) : Promise<string>
     generateCustomAction(action: customAction): Promise<void>
     flattenContract (source: any, filePath: string, data: any): Promise<string>
     hideSpinner(): void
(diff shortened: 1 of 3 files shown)


libs/remix-ui/solidity-uml-gen/src/types/index.ts
```diff
@@ -27,5 +26,5 @@ export interface ISolidityUmlGen extends ViewPlugin {
 
 export type ThemeQualityType = { name: string, quality: 'light' | 'dark', url: string }
 
-export type ThemeSummary = { themeName: string, backgroundColor: string, textColor?: string,
-shapeColor?: string, fillColor?: string }
\ No newline at end of file
+export type ThemeSummary = { name: string, quality: 'light' | 'dark', url: string, backgroundColor: string, textColor?: string,
```
fwx5618177, line 29: Maybe we could use intersection types?

```typescript
export type ThemeQualityType = { name: string, quality: 'light' | 'dark', url: string }
export type ThemeSummary = ThemeQualityType & {
  backgroundColor: string,
  textColor?: string,
  shapeColor?: string,
  fillColor?: string,
}
```

Aniket-Engg (approved):

Issue #3915, opened 2023-07-14
Optimizations to Etherscan react code


joeizang:
Even though this is merged, I think it's worth noting that the code marked might need to be refactored and optimized. When you have a function inside of a react component and you use it inside of a useEffect, that function refrence will be created everytime and closures will be created everytime which could bring performance penalties that are unwanted. Solution would be to wrap your function inside of a useCallback which basically memoizes the function and its references so that react doesn't need to recreate anything if the function itself doesn't actually change. Function name is updateConsFields and code in question is from line 55 - 70 in VerifyVIew.tsx

fwx5618177:
@joeizang @Aniket-Engg Have solved it.

Here is PR: https://github.com/ethereum/remix-project/pull/3945

PR #3950, opened 2023-07-26
Updates to VerifyView component

rich-choy:
## Related issues

- #3915

## Summary of changes

- Wrap `updateConsFields` callback with `useCallback`
- Move dependent `useEffect` below `updateConsFields` and add missing dependencies


Aniket-Engg:
A different plugin is getting developed for this. So closing it

PR #3952, opened 2023-07-26
Removed the caret down icon

AnuragNishad12:
I doest see a point in caret down icon so i just removed it 
image


Aniket-Engg:
fixed by #3971

Issue #3927, opened 2023-07-18
Quick fixes in editor


Aniket-Engg:
Add various possible quick fixes in the editor for compiler errors and warnings

Aniket-Engg:
Closing as of now. More specific issues can be further created if required

PR #3953, opened 2023-07-26
add quickfix for visibility

Aniket-Engg:
Related to #3927 

If there is no visibility given in a method, quick fix will add `public` visibility


joeizang (approved):
LGTM

PR #3954, opened 2023-07-27
Added Cursor Pointer On list item

AnuragNishad12:
The files which we create doesn't have cursor as pointer 
image
So , I just Added It


joeizang:
I am not sure about this pr. The feature he alleges he has fixed is a non issue for me.

LianaHus:
After this change, the cursor is a pointer also on the margin part of the FE (lower the last element) and part of the menu items at the top.
The best would be to add this for
1. 'remixui_items' class in file-explorer.css
2. on this span https://github.com/ethereum/remix-project/blob/startCoding/libs/remix-ui/workspace/src/lib/components/file-explorer-menu.tsx#L75

LianaHus:
Also those 2 labels should be spans

https://github.com/ethereum/remix-project/blob/startCoding/libs/remix-ui/workspace/src/lib/components/file-explorer-menu.tsx#L86 
https://github.com/ethereum/remix-project/blob/startCoding/libs/remix-ui/workspace/src/lib/components/file-explorer-menu.tsx#L111C25-L111C25

PR #3961, opened 2023-08-02
Add Prettier formatting and git  pre-commit hook

joeizang:
Adds prettier formatting to the workspace so before git commit the precommit  hook will ask prettier to format all js, tsx, ts and jsx files to be formatted based on the rules we set in the config. This doesn't replace eslint and the work @bunsenstraat did. It simply extends it.

### How to test:

- checkout this branch locally
- mess up the formatting of any tsx, ts, js or jsx file.
- try to do a git commit and the pre-commit hook will kick in  and trigger prettier to format the code file.


yann300:
I do  `git commit -m "test"` but the prettier doesn't execute..

PR #3971, opened 2023-08-07
Remove File Explorer Redundant Caret Icon

### LINKED ISSUE #3522: Caret in the File Explorer is not needed
This caret in the FE is not needed.  It has no functionality.  There may be an old issue about this.

image

joeizang:
fixes #3522


LianaHus:
github import doesn't work anymore
all cases when we add subrootfolder
also I connecte to remixd with long folder list I can't scroll to the end

LianaHus (changes requested):
github import doesn't work anymore
all cases when we add subrootfolder
also, I connected to remixd with a long folder list I can't scroll to the end

LianaHus:
could you please add pb-1 on a span for menu items?
my other comment about remixd is fixed in my pr.

LianaHus (approved):

PR #3976, opened 2023-08-09
Removed a element

AnuragNishad12:
image


LianaHus:
i don't see this element. What should I do to reproduce this?

Aniket-Engg:
> i don't see this element. What should I do to reproduce this?

you need to right click on search plugin icon

LianaHus:
ok the issue is search plugin has no documentation URL in it's profile @Aniket-Engg

Aniket-Engg:
Fixed by #3991

PR #3990, opened 2023-08-14
fixing fe layout for long lists

LianaHus:
When we connect to localhost with a large amount of files the scrollable area of FE is bigger than the screen. so we can't see the last files.


joeizang (approved):
LGTM

PR #3992, opened 2023-08-15
Intl patch3

drafish:
### changes
1. patch a lot of en locale messages
2. add `index.js` for Spanish and French
3. add some new locale json files

This time, I didn't submit zh locale messages. I think it will be better if I upload zh locale json files on crowdin. And @ryestew will submit a pr from crowdin to update the non-english locale json.

Since crowdin is already handling the non-english languages, there is no need to maintain them manually. Because everything you update on non-english languages will be overwritten by crowdin. 

For example, here is [a pr](https://github.com/ethereum/remix-project/pull/3970/files) that submitted by @LianaHus . I see you updated 4 languages. But actually you only need to update english. Other three will be updated by crowdin. And I think it will be better if non-english languages only be updated by crowdin, because it wouldn't cause any conflicts when merging crowdin pr.

And some times you will add some new strings, you probably will wonder if you need to add those new strings to non-english languages. The answer is still no. Because non-english languages will be filled in the gaps with english. Check the code below.
https://github.com/ethereum/remix-project/blob/master/apps/remix-ide/src/app/tabs/locales/zh/index.js#L15-L17

There is a situation that you need to notice. When you add a new json file, you need to add it in other languages and import it in `index.js`. Otherwise, the translations wouldn't work.

That's about everything. If you guys have any questions, feel free to ask me anytime. 

The intl feature is coming from [Remix CC](https://github.com/PlanckerLabs/remix-cc), and we will provide long term support for it, and any other features that we probably will contribute in the future.


ryestew (commented):
This PR looks OK.  It is such a large PR it needs another reviewer.

ryestew (approved):
I reviewed with @joeizang

joeizang (approved):

PR #4010, opened 2023-08-18
Monaco editor version update

Aniket-Engg:


LianaHus:
- there is a border on selected line in editor

Aniket-Engg:
> * there is a border on selected line in editor
> * greater/less symbols have red color
>   Frame 7 (2)

These are default changes done in the newer version

LianaHus:
> > * there is a border on selected line in editor
> > * greater/less symbols have red color
> >   Frame 7 (2)
> 
> These are default changes done in the newer version

ok but this should be fixed. we can discuss <> but selected line looks ugly

LianaHus (changes requested):
As discussed in the chat, please restore the current(live) style of the selected line. If you still suggest changing it, we can discuss it on the UX call.

LianaHus (approved):

PR #4016, opened 2023-08-21
fix simulator server

yann300:


libs/remix-simulator/bin/ethsim
```diff
@@ -26,14 +26,19 @@ program
   .option('-c, --coinbase [coinbase]', 'specify coinbase', '0x0000000000000000000000000000000000000000')
   .option('--rpc', 'run rpc server only', true)
   .option('--details', 'display payloads for every requests and their responses', false)
-  .action(() => {
+  .action((option) => {
+    console.log('coinbase: ', option.coinbase)
```
Aniket-Engg, line 30: do you want to keep these logs?

yann300: yes I'd prefer

Aniket-Engg: ok

PR #4018, opened 2023-08-21
Prettier format workspace.

joeizang:
This uses prettier to format and clean all *.ts, *.tsx and *.js files in our codebase.


yann300 (approved):

drafish:
Hey you formate the codebase. Great job @joeizang  

But there still have one more issue. It is a small conflict between prettier and eslint. I have `prettier` installed on my vscode. So everytime I save the files, it will formate my code automatically. It does not need to wait until pre-commit hook to do that. 

If you are using vscode, just install prettier extension, and try to save `libs/remix-ui/solidity-unit-testing/src/lib/solidity-unit-testing.tsx` , then you will see the problem.

image

L869 ~ L885 is formatted, and the red text-decoration shows up, because it's not fit the rules of eslint. It will be fixed eventually by lint-staged in pre-commit hook. But it's still annoying when I review on vscode.

I'v already figure out what cause this issue. Check #3939  @bunsenstraat added a rule(`"indent": ["error", 2]`) in eslint config file. This is the rule that cause this conflict. It's not a big problem. This conflict only exist when you are using ternary operations. But I still hope you can fix it. The simplest way is just delete this rule.

yann300:
see above , @bunsenstraat are you ok to remove that rule?

PR #4034, opened 2023-08-24
write multiple files from plugin

TylerSehr:


joeizang:
@TylerSehr Hi. Can you please describe the motivation behind the pr. It helps quite a bit.

apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -217,6 +218,38 @@ class FileManager extends Plugin {
...
+  * @param {string[]} filePaths paths of the files
+  * @param {string[]} data content to write to each file
+  * @param {string} folderPath base folder path
+  * @returns {void}
+  */
+  async writeMultipleFiles(filePaths, fileData, folderPath) {
+    try {
+      let alert = true
+      for (let i = 0; i < filePaths.length; i++) {
+        let installPath = folderPath + "/" + filePaths[i]
```
joeizang, line 232: Since installPath is not modified any where else. This can be a const rather than a let.

joeizang (changes requested):

TylerSehr:
writeMultipleFiles allows for a plugin to add entire protocols to the remix editor, preserve the file structure for the protocols, and only prompt once to write the files into a single folder. 

Currently plugins are able to do this using github gists, but that results in file collisions and certain files can be overwritten. 

We're updating the Cookbook plugin to support full protocols, so we want to be able to import an entire protocol into remix.

apps/remix-ide/src/app/files/fileManager.ts
TylerSehr: Updated installPath to const rather than let

joeizang (approved):
LGTM

apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -217,6 +218,38 @@ class FileManager extends Plugin {
...
+          this.emit('fileAdded', path)
+        }
+        alert = false
+      }
+    } catch (e) {
+      throw new Error(e)
+    }
+  }
+
+
```
LianaHus, line 251: Please remove extra lines here

TylerSehr:
Added rename Cookbook Plugin

PR #4035, opened 2023-08-28
cursor pointer of enabled AtAddress btn

### LINKED ISSUE #3889: AtAddress button is still clickable when it's disabled
AtAddress button is still clickable when it's disabled. And when it's enabled, cursor does not become small hand, it's still an arrow.

I know it's a small problem. But I still hope remix can fix it.

I'v already fixed it in brochain ide. I will submit a pr for remix.

LianaHus:
Fix #3889

@drafish The issue was the focus around the btn but I rmed it. thanks for the issue


drafish (approved):
nice job @LianaHus

PR #4037, opened 2023-08-28
add interaction with faucet for ephemery testnet

yann300:


apps/remix-ide/src/app/providers/injected-ephemery-testnet-provider.tsx
```diff
@@ -1,17 +1,41 @@
...
   constructor() {
-    super(profile, 'Ephemery Testnet', '0x259C709', ['https://otter.bordel.wtf/erigon', 'https://eth.ephemeral.zeus.fyi'])
+    super(profile, 'Ephemery Testnet', '', ['https://otter.bordel.wtf/erigon', 'https://eth.ephemeral.zeus.fyi'])
+  }
+
+  async init() {
+    const chainId = await new Web3(this.rpcUrls[0]).eth.getChainId()
+    this.chainId = `0x${chainId.toString(16)}`
+    this.chainName = `Ephemery Testnet ${chainId}`
+    console.log('ephemeral chainid', chainId, this.chainId)
```
Aniket-Engg, line 23: remove log

Aniket-Engg (commented):
When I tried to load balance from faucet by clicking the icon from Environment textbox:
1. It opens the dropdown which is bad. May be we can put this icon on the right size of ACCOUNT field
2. It shows error in toaster: `Unable to call the faucet api Failed to fetch`

yann300:
> It opens the dropdown which is bad. May be we can put this icon on the right side of ACCOUNT field

L2 provider are also using this mechanism to show more action, but I agree UX isn't ok here.

yann300:
> It shows error in toaster: Unable to call the faucet api Failed to fetch . I think right faucet API link is https://faucet.bordel.wtf/api/claim

This is a cors error that should be fixed by the remote service itself. I already reported that error.

joeizang (approved):

yann300:
please don't merge, we still to properly test it.

ioedeveloper (commented):
Code changes looks fine. Still need to test UI.

PR #4045, opened 2023-08-31
quickfix to mark a contract abstract

Aniket-Engg:
This adds a quickfix for compiler error: `TypeError: Contract "ContractName" should be marked as abstract.`

related to #3927 

Use this code to reproduce that compiler error:
```
interface Store {
    function store(uint256 num) external;
}

contract Storage is Store {

    uint256 number;
}
```


yann300:
I don't see the "abstract" quick fix witht the code sample that you provided..

joeizang (approved):
LGTM

PR #4050, opened 2023-09-04
Quickfix to add 'virtual' and 'override' to function

Aniket-Engg:
related to #3927 

It shows quickfix for compilation errors like:
`TypeError: Trying to override non-virtual function. Did you forget to add "virtual"` 
&
`TypeError: Overriding function is missing "override" specifier`

code to reproduce these errors:
```
pragma solidity 0.8.18;

contract Store {
    function store(uint256 num) public {
        num = num + 1;
    }
}

contract Storage is Store {
    function store(uint256 num) public {
        num = num + 2;
    }
}
```

This PR also involves some code refactoring


LianaHus (changes requested):
IMHO it should do the opposite thing. instead of updating the file to the current version we should update the current version to the acceptable one.

LianaHus (approved):

PR #4059, opened 2023-09-11
display custom error details for injected web3

### LINKED ISSUE #3765: Custom error not working correctly.
Hello Members!

Custom Error is not working same as in VM and Testnet(Sepolia, Goerli..etc).
The caller will not get the reason behind why the function call is failing.

Following snippet works as expected in VM but when deployed in Sepolia...Does not provide which custom error was triggered...
```


// SPDX-License-Identifier: GPL-3.0
pragma solidity ^0.8.19;

contract Test {

    error InvalidAddressError();
    error InvalidAddressWithParamError(address _addr);
    
    function checkCustom() public pure returns (bool){
        address _input = address(0); 
        if (_input == address(0)) revert InvalidAddressError();
        return true;
    }

    function checkRequire() public pure returns (bool){
        address _input = address(1);

        require(_input == address(0), "############ Address must be non zero");
        return true;
    }

    function checkCustomParametrised() public pure returns (bool){
        address _input = address(4);

        if (_input != address(0)) revert InvalidAddressWithParamError(_input);
        return true;
    }
   
}
```

Aniket-Engg:
fixes #3765

### DIFF
--- a/libs/remix-lib/src/execution/txExecution.ts
+++ b/libs/remix-lib/src/execution/txExecution.ts
@@ -244,6 +244,6 @@ export function checkInjectedError (errorObj, compiledContracts) {
         msg += `\n${JSON.stringify(decodedCustomErrorInputsClean, null, ' ')}`
       }
     }
-  return msg
+    return msg
   }
 }
--- a/libs/remix-lib/src/execution/txExecution.ts
+++ b/libs/remix-lib/src/execution/txExecution.ts
@@ -54,12 +54,11 @@ export function callFunction (from, to, data, value, gasLimit, funAbi, txRunner,
 /**
   * check if the vm has errored
   *
-  * @param {Object} execResult    - execution result given by the VM
+  * @param {Object} execResult    - execution result given by the VM, contain errorMessage and errorDate
   * @param {Object} compiledContracts - Object containing contract compilation details
-  * @param {boolean} isInjectedWeb3 - If true, provider is set to injected web3
   * @return {Object} -  { error: true/false, message: DOMNode }
   */
-export function checkError (execResult, compiledContracts, isInjectedWeb3) {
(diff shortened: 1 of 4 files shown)


libs/remix-lib/src/execution/txExecution.ts
```diff
@@ -173,3 +173,77 @@ export function checkVMError (execResult, compiledContracts) {
   ret.message = `${error}\n${exceptionError}\n${msg}\nDebug the transaction to get more information.`
   return ret
 }
+
+export function checkInjectedError (errorObj, compiledContracts) {
```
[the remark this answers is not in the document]
Aniket-Engg: done. Please check

libs/remix-lib/src/execution/txExecution.ts
```diff
@@ -75,20 +75,20 @@ export function checkVMError (execResult, compiledContracts) {
     error: false,
     message: ''
   }
-  if (!execResult.exceptionError) {
+  if (!execResult.exceptionError && !isInjectedWeb3) {
```
yann300, line 78: seems like `error` and `returnData` are both the same but present under different properties... could you add a type `execResult`, so you don't have to do stuff like `isInjectedWeb3 ? execResult.error : exceptionError`

yann300: that way you may not need to have the `isInjected`

Aniket-Engg: I think `execResult` is quite different for VM and injected Web3. For Injected web3, it is just an object having error details

Aniket-Engg: updated

yann300 (approved):

PR #4063, opened 2023-09-12
Fixcompiler2

bunsenstraat:
The timestamp is used when the compiler ( only the instance in the code parser ) is in Retrigger mode, so that when multiple compiles are triggered only the last one executes. The timestamp is used to abort a previous compile. But it was implemented in the wrong way because each new job from an import would get the latest timestamp and not the one from the origin job that triggered. Now the timestamp is being passed around correctly.
Basically this would become visible when switching between files, the compilation data coming in was not from the file you opened but from the previous file. This would become apparent when a file had a lot of imports because then import jobs would be associated with the wrong timestamp.
Added a test to switch quickly between files and verify the compile data is visible in the hover.


Aniket-Engg (approved):
LGTM

PR #4080, opened 2023-09-21
style of ask gpt and infinit loop and other issues fix

LianaHus:
- fixing the issue of the second click on "ask gpt"
- corrected the number of iterations in the timer to use the lengths of the message 
- removed the plugin from the public list
- style improvements
- scroll terminal to bottom after each line (improvement in typewriter)

### DIFF
--- a/libs/remix-ui/renderer/src/lib/renderer.tsx
+++ b/libs/remix-ui/renderer/src/lib/renderer.tsx
@@ -96,11 +96,11 @@ export const Renderer = ({message, opt = {}, plugin}: RendererProps) => {
           <div className="close" data-id="renderer" onClick={handleClose}>
             <i className="fas fa-times"></i>
           </div>
-          <div className="d-flex flex-row-reverse">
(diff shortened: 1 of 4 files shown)


LianaHus:
An open question: Do we need to track the usage in matomo. @yann300 @Aniket-Engg

Aniket-Engg (commented):
LGTM

libs/remix-ui/renderer/src/lib/renderer.tsx
```diff
@@ -76,6 +78,7 @@ export const Renderer = ({message, opt = {}, plugin}: RendererProps) => {
       explain why the error occurred and how to fix it.
       `
       await plugin.call('openaigpt', 'message', message)
+      _paq.push(['trackEvent', 'GPTSupport', 'askGPT'])
```
Aniket-Engg, line 81: Most of the places, we use 4 fields for this

Aniket-Engg (approved):

PR #4084, opened 2023-09-25
fixing drag drop issues

bunsenstraat:
BUG: if you dragged a file or folder over other folders they would open too quickly, instantly, making it completely impossible to drag anything around in a complex workspace. ie try dragging over the .deps folder, you'd never get past it. 
Solution: add a timer to the open folder event when dragging ( 500ms ) and resetting that timer.

BUG: you can't drop anything on root
Solution: add a dropzone underneath

Improvement: add modals to confirm moving. it misfires too quickly otherwise


joeizang (commented):
LGTM

libs/remix-ui/drag-n-drop/src/lib/drag-n-drop.tsx
```diff
@@ -75,7 +87,12 @@ export const Draggable = (props: DraggableType) => {
           onDrop={(event) => {
             handleDrop(event)
           }}
-          onDragStart={() => {
+          onDragStart={(event) => {
+            if(destination && destination.path === '/'){
```
LianaHus, line 91: space is missing after if

libs/remix-ui/drag-n-drop/src/lib/drag-n-drop.tsx
```diff
@@ -75,7 +87,12 @@ export const Draggable = (props: DraggableType) => {
...
+              event.preventDefault()
+              event.stopPropagation
+            }else
```
LianaHus, line 94: same here

LianaHus (approved):

PR #4088, opened 2023-09-25
fix filtering event

yann300:
this fixes 2 issues:

 - `changeEvent.tx.to || '')` instead of checking that the "to" is null, we do the check against the `receipt.contractAddress` which contains the address of the newly created contract.
 - use `log.address` instead of `tx.to`, that way it won't obfuscate event emiitted from a call.


yann300:
e2e will be added in https://github.com/ethereum/remix-project/issues/4089

bunsenstraat (approved):

PR #4092, opened 2023-09-26
Fix Copy to Clipboard

### LINKED ISSUE #3512: copy to clipboard copy wrong address
- switch to injected
 - "copy to clipboard" the account address in remix, this is the right one
 - select another account in metamask
 - the account is changed in the dropdown list
 - click on "copy to clipboard", the copied account is still the previous one

joeizang:
fixes #3512


ryestew:
I'm getting a spinner in the account pulldown that is not stopping.

image

PR #4098, opened 2023-09-27
File sorting

### LINKED ISSUE #123: File Explorer - make files sort alphabetically by default
The file explorer (browser and remixd) does not show the files in alphabetical order, I don't know what order it uses, but it is all jumbled.

This makes life hard when you're dealing with e.g. 20 contracts.

Can the default file order be alphabetical.

Perhaps in future a sort ascend/descend could be added too.

bunsenstraat:
fixes #123 
sorts by file & folders and alphabetically. folders first, then files.


libs/remix-ui/workspace/src/lib/components/file-explorer.tsx
```diff
@@ -315,6 +317,21 @@ export const FileExplorer = (props: FileExplorerProps) => {
     }
   }
 
+  useEffect(() => {
+    if(files[ROOT_PATH]){
```
LianaHus, line 321: space

PR #4107, opened 2023-10-02
show erc list in editor

Aniket-Engg:
Show a list of listed erc interfaces and standards when user types `erc` in the editor.

On clicking enter, relevant code should be pasted in the code according to selection.

Inspired from: https://github.com/juanfranblanco/vscode-solidity/tree/master#erc-erc-drafts-and-smart-contracts-snippets--reference

erc details are fetched from: https://github.com/juanfranblanco/vscode-solidity/blob/master/snippets/solidity.json


libs/remix-ui/editor/src/lib/providers/completion/completionGlobals.ts
```diff
@@ -188,7 +188,145 @@ export function getCompletionSnippets(range: monacoTypes.IRange, monaco): monaco
       insertText: '// SPDX-License-Identifier: MIT',
       insertTextRules: monaco.languages.CompletionItemInsertTextRule.InsertAsSnippet,
       range
+    },
+    /* eslint-disable */
+    // Related to https://github.com/juanfranblanco/vscode-solidity/blob/master/snippets/solidity.json
+    {
+      label: 'erc20i',
```
yann300, line 195: probably the labels should be more explicit, like `generate erc20i` for instance.

Aniket-Engg: > probably the labels should be more explicit, like `generate erc20i` for instance.

that is a keyword to show the list

Aniket-Engg: > probably the labels should be more explicit, like `generate erc20i` for instance.

that is a keyword to show the list

Aniket-Engg: > btw just wondering, do we already have snippet for generating loops?

yes, we have

yann300: > that is a keyword to show the list. We can update the detail field

yes, in  https://github.com/juanfranblanco/vscode-solidity/raw/master/screenshots/ercautocomplete1.png
there's a quick and a full description associated with the keyword.

Aniket-Engg: > > probably the labels should be more explicit, like `generate erc20i` for instance.
> 
> that is a keyword to show the list. We can update the detail field

updated @yann300

yann300:
ok I see thanks for the update, is this also possible to add the full description, as in the screenshot above?

Aniket-Engg:
> ok I see thanks for the update, is this also possible to add the full description, as in the screenshot above?

it is there, when you click on `>` of particular option

bunsenstraat (approved):

PR #4108, opened 2023-10-03
Update web3js 4

yann300:


package.json
```diff
@@ -154,6 +154,7 @@
     "color-support": "^1.1.3",
     "commander": "^9.4.1",
     "core-js": "^3.6.5",
+    "cors": "^2.8.5",
```
bunsenstraat, line 157: you use that somewhere?

yann300: it's used by remix-simulator....

bunsenstraat (approved):

PR #4110, opened 2023-10-04
Write multiple files hotfix

TylerSehr:
WriteMultipleFiles times out unless user selects remember when calling the function. 

to reproduce:
- use Cookbook plugin, and select protocols
- click on a protocol
- click accept on the popup, but not tick remember
- call will time out and protocol will not open.


apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -227,22 +227,21 @@ class FileManager extends Plugin {
   */
   async writeMultipleFiles(filePaths, fileData, folderPath) {
     try {
-      let alert = true
-      for (let i = 0; i < filePaths.length; i++) {
-        const installPath = folderPath + "/" + filePaths[i]
+      filePaths.forEach(async (filePath, index) => {
```
joeizang, line 230: forEach is not a good place to do anything async. It is generally not recommended to do anything async in a foreach because forEach loops doesn't wait for any async ops to finish and the result of which is race conditions and bugs.
Using a for...of or a Promise.all might be better here but your milage may vary.

apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -227,22 +227,21 @@ class FileManager extends Plugin {
...
+        const installPath = folderPath + "/" + filePath
+        const alert = !index
```
joeizang, line 232: You have two variables named the same thing only differentiated by scope which isn't ideal. Your code should communicate intent as much as possible and if your have an inner alert and an outer one then what is the purpose of the outer one.

joeizang (changes requested):

apps/remix-ide-e2e/src/tests/compiler_api.test.ts
```diff
@@ -34,7 +34,15 @@ module.exports = {
       .click('*[data-id="terminalClearConsole"]')
   },
 
-  'Should compile using "compileWithParamaters" API with optimization off check default runs #group3': function (browser: NightwatchBrowser) {
+  'Should compile using "compileWithParamaters" API with optimization off check default runs': function (browser: NightwatchBrowser) {
```
LianaHus, line 37: Why was this removed from the group?

bunsenstraat:
It looks to me like lots of changes from some other branch were merged in this one. Beter to start with a clean branch from master and apply this fix there.

PR #4115, opened 2023-10-06
chore: add notification workflow after publishing a new version, auto…

fwx5618177:
- Add a new workflow. It could notify us in Discord, Email, Slack, TG, and so on after publishing a new version on the main branch or a special branch.


fwx5618177:
@Aniket-Engg @step21 @evertonfraga  PTAL.

step21:
why am I tagged here?

yann300:
we already added some notifications recently, so we won't merge this PR as it is now. As far as I see the `Notify on Discord` would be interesting for us (we don't have public tg chat, slack, etc..).

@Aniket-Engg would this be something interesting to have?

yann300:
maybe I misunderstood it, are you looking to send the notification in the remix discord public channel or in your own channels infra?

PR #4128, opened 2023-10-11
Set default solidity compiler version to 0.8.22

Aniket-Engg:


LianaHus (approved):

LianaHus:
Shouldn't you update a script as well? downloadsolc_assets

Aniket-Engg:
> Shouldn't you update a script as well? downloadsolc_assets

That automatically fetch the latest version

yann300 (approved):

PR #4132, opened 2023-10-11
fix method write multiple

bunsenstraat:
- This calls the permission modal on the writeMultipleFiles only once per call
- removed some clutter we don't need
- added a test for the API method

TO TEST:
go to the cookbook plugin and run a protocol import it should write multiple files at once after permissions


Aniket-Engg (approved):
LGTM

Issue #4046, opened 2023-08-31
FE style fixes


LianaHus:
![Screenshot from 2023-08-31 15-21-47](https://github.com/ethereum/remix-project/assets/9685356/0738432d-a7a8-4b19-a3cd-c6279fdc1ef9)

- [x]  Better margins like on Screenshot
- [x]  Bigger menu items
- [x]  Cut the long file name
- [x]  Add tooltips with path, with delay
- [x]  The click on the file/folder name moves it to the left (**Issue**!)
- [x]  last icon should be far fa
- [x]  last two menu items label->span
- [x]  cursor pointer for files and menu items

LianaHus:
![Screenshot from 2023-09-05 14-15-25](https://github.com/ethereum/remix-project/assets/9685356/36af1126-bbef-4953-b8fc-9c76c18bf153)

PR #4141, opened 2023-10-16
FE tooltips are back.

LianaHus:
We have a path now instead of names, and we show the tooltip with a delay.
Besides it is on the label and not the line, so when the user is hovering they will not see tooltips on short names.

The issue is part of  https://github.com/ethereum/remix-project/issues/4046


ryestew:
while typing more than the limited number of characters, the text disappears - so you don't know what the name is.

PR #4143, opened 2023-10-16
Oz proxy backwards compatible

bunsenstraat:
- allows deploying and upgrading on v4 and v5 of the erc1967 proxy
- add tests for both versions plus tests for the debugger

how to test:
- add a token, v5 will require version > 8.20:
- v5 can be added using remix templates
- v4 example:

```
      // SPDX-License-Identifier: MIT
      pragma solidity ^0.8.4;
      
      import "@openzeppelin/contracts-upgradeable@4.8.3/token/ERC721/ERC721Upgradeable.sol";
      import "@openzeppelin/contracts-upgradeable@4.8.3/access/OwnableUpgradeable.sol";
      import "@openzeppelin/contracts-upgradeable@4.8.3/proxy/utils/Initializable.sol";
      import "@openzeppelin/contracts-upgradeable@4.8.3/proxy/utils/UUPSUpgradeable.sol";
      
      contract MyToken is Initializable, ERC721Upgradeable, OwnableUpgradeable, UUPSUpgradeable {
          /// @custom:oz-upgrades-unsafe-allow constructor
          constructor() {
              _disableInitializers();
          }
      
          function initialize() initializer public {
              __ERC721_init("MyToken", "MTK");
              __Ownable_init();
              __UUPSUpgradeable_init();
          }
      
          function _authorizeUpgrade(address newImplementation)
              internal
              onlyOwner
              override
          {}
      }
```

- deploy this with proxy, for v5 add an address at parameter
- upgrade this contract with:
```
      import "./myTokenV1.sol";

      contract MyTokenV2 is MyToken {
        function version () public view returns (string memory) {
            return "MyTokenV2!";
        }
    }
```
- run a call ie 'version' on the second deployed proxy
- it should return MyTokenV2!
- run the debugger on this call


libs/remix-core-plugin/src/lib/openzeppelin-proxy.ts
```diff
@@ -46,28 +46,36 @@ export class OpenZeppelinProxy extends Plugin {
...
-          }
-        }))
-      }
-    }))
+    await Promise.all(
+      Object.keys(contracts).map(async (name) => {
+        if (ast) {
+          const UUPSSymbol = ast.exportedSymbols[UUPS] ? ast.exportedSymbols[UUPS][0] : null
+
+          await Promise.all(
```
yann300, line 54: is there only changes in the indentation?

libs/remix-core-plugin/src/lib/openzeppelin-proxy.ts
```diff
@@ -118,51 +137,58 @@ export class OpenZeppelinProxy extends Plugin {
...
-          funAbi: UUPSupgradeAbi,
-          funArgs: [newImplAddress],
-          linkReferences: {},
-          dataHex: fnData.replace('0x', '')
+    await this.blockchain.runTx(
+      args,
+      () => {},
+      () => {},
+      () => {},
+      async (error, txResult, _address, returnValue) => {
```
yann300, line 154: this part didn't change too right? except of the new logs in the terminal.

bunsenstraat: no it's wrapped in a callback from the tx that gets the version of the erc

bunsenstraat: so it makes two calls depending on the version

libs/remix-core-plugin/src/lib/openzeppelin-proxy.ts
bunsenstraat: yeah

bunsenstraat: I can reset that formatting

yann300 (approved):

PR #4144, opened 2023-10-17
Circuit compiler UI

ioedeveloper:
User Interface for Circom Plugin


apps/circuit-compiler/src/app/app.tsx
```diff
@@ -1,17 +1,140 @@
...
+  const [plugin, setPlugin] = useState<CircomPluginClient>(null)
+  const [locale, setLocale] = useState<{code: string; messages: any}>({
+    code: 'en',
+    messages: null
+  })
+  const [isContentChanged, setIsContentChanged] = useState<boolean>(false)
 
   useEffect(() => {
-    new CircomPluginClient()
+    const plugin = new CircomPluginClient()
```
bunsenstraat, line 21: setting plugin clients in useeffect and in a state will cause problems. especially when using strict mode. I've seen situations with other plugins doing this where it would either try to connect to Remix twice or never connect at all. You can can listen on the onload event here just fine. but I'd set the plugin as a const and export it.

ioedeveloper: Can you make a commit with your recommended changes?

ryestew:
The name should updated to:
Circom ZKP compiler

ryestew:
Needs a button to make some example files - but if this is a workspace and a workspace is created - somehow the user needs to be informed that a workspace has been created.  So you could generate workspace and automatically get switched to the FE.

ryestew:
and we need documentation for this plugin

apps/circuit-compiler/src/app/app.tsx
bunsenstraat: sure

apps/circuit-compiler/src/profile.json
```diff
@@ -8,8 +8,8 @@
   "canActivate": [],
   "url": "",
   "description": "Enables circuit compilation and computing a witness for ZK proofs",
-  "icon": "https://docs.circom.io/assets/images/favicon.png",
-  "location": "hiddenPanel",
+  "icon": "assets/img/circom.webp",
```
yann300, line 11: the icon doesn't show up...

apps/circuit-compiler/src/app/services/circomPluginClient.ts
```diff
@@ -1,17 +1,26 @@
...
+  }
+  private lastCompiledCircuitPath: string = ''
+  private lastParsedFiles: Record<string, string> = {}
+  private lastCompiledFile: string = ''
 
   constructor() {
     super()
     createClient(this)
     this.internalEvents = new EventManager()
-    this.methods = ['init', 'parse']
```
bunsenstraat, line 14: set the methods before the createclient

bunsenstraat: If I remember correctly :D

bunsenstraat:
I think you can add the plugin tor required plugins in app manager so you don't have to give permissions.

apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -255,13 +255,13 @@ class FileManager extends Plugin {
    * @param {string} path path of the file
    * @returns {string} content of the file
    */
-  async readFile(path) {
+  async readFile(path, isBuffer?) {
```
bunsenstraat, line 258: if we start doing this we should maybe just add encoding as a param, which is the standard fs way. so we can support everything.

bunsenstraat (changes requested):

ryestew (commented):
For translations, you have updated the index.js  and the circuit.json file in remix-ide/src/app/tabs/locales/en.

You don't need to add the circuit.json file in the es, fr, and zh folders.  Please remove these .json files - only keep the one in the en  folder.

bunsenstraat (approved):

PR #4152, opened 2023-10-20
React18v4

bunsenstraat:


.circleci/config.yml
```diff
@@ -166,6 +166,7 @@ jobs:
           at: .
       - run: unzip ./persist/dist.zip
       - run: yarn install --cwd ./apps/remix-ide-e2e --modules-folder ../../node_modules
+      - run: mkdir node_modules/hardhat && wget https://unpkg.com/hardhat/console.sol -O node_modules/hardhat/console.sol
```
yann300, line 169: why do you need to do this?

apps/remix-ide-e2e/src/helpers/init.ts
```diff
@@ -27,30 +27,36 @@ export default function (browser: NightwatchBrowser, callback: VoidFunction, url
...
               display:none !important;
             }
           `);
         })
-      }
+    }})
+    .perform(() => {
+      browser.execute(function () { 
+        (window as any).logs = []
+        console.log = function () {
```
yann300, line 51: is this a new feature for the CI?

libs/remix-ui/run-tab/src/lib/actions/actions.ts
```diff
@@ -88,6 +88,7 @@ export const updateScenarioPath = (dispatch: React.Dispatch<any>, path: string)
 }
 
 export const setSendTransactionValue = (dispatch: React.Dispatch<any>, value: string) => {
+  console.log('setSendTransactionValue', value)
```
yann300, line 91: to remove?

.circleci/config.yml
bunsenstraat: the console.sol is slow to load from unpkg in e2e. it caused issues

bunsenstraat: this puts it where the localhost can find is using the nx_npm_url param

apps/remix-ide-e2e/src/helpers/init.ts
bunsenstraat: yeah you can fetch the logs then in any test. helpful but maybe not necessary for this PR.

yann300 (approved):

drafish:
https://github.com/ethereum/remix-project/blob/react18v4/package.json#L369-L371

```
  "resolutions": {
    "@types/react": "^17.0.24"
  }
```

This should be removed. I added this to fix typescript error in #2581 . Because it will default to be ^18 . Now @bunsenstraat update react version to ^18, this should be removed.

PR #4153, opened 2023-10-20
Move compilation details to Main panel plugin

joeizang:


ryestew (changes requested):
This is great!  It just needs a few fixes:
- [ ] The carets disappear in the down state - see image below
- [ ] In ? do the same thing as the carets - is that correct - if so they should be removed.
- [ ] Click on bytecode and there is a lot of stuff - but it is not changing the right side scroll bar (which should be much much longer)(horizontal scroll does not reflect the contents either):
image

joeizang:
> This is great! It just needs a few fixes:
> 
> * [ ]  The carets disappear in the down state - see image below
> * [ ]  In ? do the same thing as the carets - is that correct - if so they should be removed.
> * [ ]  Click on bytecode and there is a lot of stuff - but it is not changing the right side scroll bar (which should be much much longer)(horizontal scroll does not reflect the contents either):
> 
> image

- 1. The carets always disappeared and is the default behaviour.
- 2. This is also the default behaviour. Maybe we keep this default behaviour and propose new behaviour for another pr?
- 3. I am not sure but will try and wrap that and see how it looks.

ryestew:
If it is the default behavior, then lets merge this and fix in another PR (which could be in the next release).

joeizang:
> If it is the default behavior, then lets merge this and fix in another PR (which could be in the next release).

You will need to approve this. 😄

yann300:
just a thought, maybe the some of the important section could be unfolded by default...

ryestew (approved):

PR #4164, opened 2023-10-25
Semaphore template

ioedeveloper:


yann300:
When a semaphore template is being created, probably we should activate the circom plugin.

bunsenstraat:
> When a semaphore template is being created, probably we should activate the circom plugin.

In the PR the circuit plugin doesn't load right? is that correct?

yann300:
> > When a semaphore template is being created, probably we should activate the circom plugin.
> 
> In the PR the circuit plugin doesn't load right? is that correct?

In this PR I don't think so, it probably should

ioedeveloper:
> > When a semaphore template is being created, probably we should activate the circom plugin.
> 
> In the PR the circuit plugin doesn't load right? is that correct?

In this PR, the circuit-compiler UI is not present, it only contains semaphore template. This PR has the changes for circuit-compiler UI https://github.com/ethereum/remix-project/pull/4144

yann300:
> > > When a semaphore template is being created, probably we should activate the circom plugin.
> > 
> > 
> > In the PR the circuit plugin doesn't load right? is that correct?
> 
> In this PR, the circuit-compiler UI is not present, it only contains semaphore template. This PR has the changes for circuit-compiler UI #4144

ok so let's do it in an another PR..

yann300 (approved):

PR #4166, opened 2023-10-26
Update web3js 4 With Injected provider

yann300:
should fix sending tx with injected & walletconnect.

### DIFF
--- a/apps/remix-ide/src/blockchain/blockchain.tsx
+++ b/apps/remix-ide/src/blockchain/blockchain.tsx
@@ -962,7 +962,7 @@ export class Blockchain extends Plugin {
       cb(null, txResult, address, returnValue)
     } catch (error) {
       if (this.isInjectedWeb3()) {
-        let errorObj = error.replace('Returned error: ', '').replace('error: ', '')
+        const errorObj = error.replace('Returned error: ', '').replace('error: ', '')
         if (errorObj) {
           const compiledContracts = await this.call('compilerArtefacts', 'getAllContractDatas')
           const injectedError = txExecution.checkError({ errorMessage: errorObj }, compiledContracts)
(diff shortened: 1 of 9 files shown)


apps/remix-ide/src/blockchain/blockchain.tsx
```diff
@@ -962,11 +962,10 @@ export class Blockchain extends Plugin {
...
     } catch (error) {
       if (this.isInjectedWeb3()) {
-        let errorObj = error.replace('Returned error: ', '')
-        errorObj = JSON.parse(errorObj)
-        if (errorObj.errorData) {
+        const errorObj = error.replace('Returned error: ', '').replace('error: ', '')
+        if (errorObj) {
           const compiledContracts = await this.call('compilerArtefacts', 'getAllContractDatas')
-          const injectedError = txExecution.checkError({ errorMessage: errorObj.error, errorData: errorObj.errorData }, compiledContracts)
+          const injectedError = txExecution.checkError({ errorMessage: errorObj }, compiledContracts)
```
yann300, line 968: Check custom error.

apps/remix-ide/src/blockchain/providers/injected.ts
```diff
@@ -29,16 +33,16 @@ export class InjectedProvider {
...
-    this.executionContext.web3().eth.getGasPrice(cb)
+    this.executionContext.web3().eth.getGasPrice().then((result => cb(null, result)))
   }
 
   signMessage (message, account, _passphrase, cb) {
     const messageHash = hashPersonalMessage(Buffer.from(message))
     try {
       message = isHexString(message) ? message : Web3.utils.utf8ToHex(message)
-      this.executionContext.web3().eth.personal.sign(message, account, (error, signedData) => {
+      this.executionContext.web3().eth.personal.sign(message, account).then((error, signedData) => {
```
ioedeveloper, line 43: This change can be an async await call, since it is in a try catch statement. That way we eliminate duplicate catch functions. Just a suggestion, it doesn't need to be changed.

ioedeveloper (approved):

PR #4167, opened 2023-10-27
Compile details update

joeizang:


libs/remix-ui/solidity-compile-details/src/lib/solidity-compile-details.tsx
```diff
@@ -13,6 +14,10 @@ export interface RemixUiCompileDetailsProps {
   saveAs: any
 }
 
+export type ContractPropertyName = 'compilerInput' | 'name' | 'metadata' | 'bytecode' | 'abi' | 'storageLayout'
```
yann300, line 17: this type is already define below, we should have one declaration. I know this add a dependency, but that's probably  better than duplicating the types.

libs/remix-ui/solidity-compile-details/src/lib/solidity-compile-details.tsx
```diff
@@ -1,6 +1,8 @@
 import { CopyToClipboard } from '@remix-ui/clipboard'
 import { CustomTooltip } from '@remix-ui/helper'
 import { TreeView, TreeViewItem } from '@remix-ui/tree-view'
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
+import { ContractPropertyName } from 'libs/remix-ui/solidity-compiler/src/lib/types'
```
yann300, line 5: please check but I think this isn't the right to write this import. You hae to use the module name (`solidity-compiler`)

joeizang: You are right. Its done.

yann300 (approved):

PR #4178, opened 2023-10-31
Compilation details cleanup

joeizang:
Clean up wrong entries and remove wrongly placed files.


apps/remix-ide/src/remixAppManager.js
```diff
@@ -114,7 +114,8 @@ export function isNative(name) {
     'injected',
     'doc-gen',
     'doc-viewer',
-    'circuit-compiler'
+    'circuit-compiler',
+    'compliationDetails'
```
Aniket-Engg, line 118: I think there is a typo in this

joeizang: Ugh!!! I will fix it now.

Aniket-Engg (approved):

PR #4179, opened 2023-10-31
create lang selector component

joeizang:
Adds dropdown selector where IDE text can be switched based on language code selected.


ryestew (changes requested):
The page has to be very very wide to see the language selector.  At least 1805px - for it to be totally in view.  It needs to be visible when the page is less wide.  On Remix Live, the featured box

ryestew (approved):

Issue #4071, opened 2023-09-18
Why I'm Getting the Gas estimation failed error 


AnuragNishad12:
When I connect My ganache with remix IDE and deploy the contract Then I gets this error 
<img width="403" alt="image" src="https://github.com/ethereum/remix-project/assets/124987038/90c4531c-2981-4750-bf34-c94236e68ba4">

yann300:
Are you using ganache UI, please check they are supporting the latest fork.
I am sure Ganache cli is, but not sure about the UI version.
Also please post the soliditi code so we can try to reproduce.

PR #4183, opened 2023-11-01
add more info for invalid opcode

### LINKED ISSUE #4079: "Invalid Opcode" issue might reveal an issue with the current used evm version
When this message pop-up (either in a modal with https://github.com/ethereum/remix-project/issues/4071 or in the terminal), there's a chance that the issue is that the contract has been compiled with an evm setting that isn't compatible with the current provider, we should precise that in the error message.
Could be something like "Please make sure the evm version used by the current provider/chain is compatible with the evm version the contract has been compiled with".

Aniket-Engg:
fixes #4079 

and fixes https://github.com/ethereum/remix-project/issues/4071

### DIFF
--- a/libs/remix-lib/src/execution/txExecution.ts
+++ b/libs/remix-lib/src/execution/txExecution.ts
@@ -83,7 +83,7 @@ export function checkError (execResult, compiledContracts) {
   const error = `Error occured: ${execResult.errorMessage}.\n`
   let msg = ''
   if (exceptionError === errorCode.INVALID_OPCODE) {
-    msg = '\t\n\tThe execution might have thrown.\n'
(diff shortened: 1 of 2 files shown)


libs/remix-lib/src/execution/txExecution.ts
```diff
@@ -83,7 +83,7 @@ export function checkError (execResult, compiledContracts) {
   const error = `Error occured: ${execResult.errorMessage}.\n`
   let msg = ''
   if (exceptionError === errorCode.INVALID_OPCODE) {
```
yann300, line 85: in my case, `exceptionError` was `invalid opcode: PUSH0` so it didn't get in that "if". Is this possible to handle that case? just using `includes` might just do the trick.

Aniket-Engg: Done

yann300 (approved):

PR #4198, opened 2023-11-07
error handling improvements

LianaHus:


apps/remix-ide/src/app/files/fileManager.ts
```diff
@@ -704,7 +704,7 @@ class FileManager extends Plugin {
   }
 
   async openFile(file?: string) {
-    if (!file) {
+    if (!file && file != '') {
```
yann300, line 707: are you really sure it's needed? or why did you need to do that?

LianaHus: there was a case of empty line instead of the name while I was debugging

yann300: can you check if you still  really need this? afaik `if (!file)` should do the same.

PR #4228, opened 2023-11-09
support submodules

bunsenstraat:


apps/remix-ide/src/app/files/dgitProvider.js
```diff
@@ -105,15 +105,38 @@ class DGitProvider extends Plugin {
...
-    if (refresh) {
+    const newgitmodules = await this.parseGitmodules() || []
+    // find the difference between the two gitmodule versions
+    const toRemove = gitmodules.filter((module) => {
+      return !newgitmodules.find((newmodule) => {
+        return newmodule.name === module.name
+      })
+    })
+
+    for (const module of toRemove) {
```
yann300, line 121: You also need to clone the git module that are not already here right? or this done elsewhere.

libs/remix-ui/workspace/src/lib/actions/workspace.ts
```diff
@@ -423,7 +424,7 @@ export const switchToWorkspace = async (name: string) => {
     await plugin.fileProviders.workspace.setWorkspace(name)
     await plugin.setWorkspace({ name, isLocalhost: false })
     const isGitRepo = await plugin.fileManager.isGitRepo()
-
+    const hasGitSubmodule = await plugin.fileManager.hasGitSubmodules()
```
yann300, line 427: it doesn't seem to be used.

apps/remix-ide/src/app/files/dgitProvider.js
bunsenstraat: somewhere else

yann300 (approved):

PR #4229, opened 2023-11-09
consider remappings while compilation

Aniket-Engg:
mappings will be fetched from remapping.txt file

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
+++ b/libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
@@ -106,10 +106,15 @@ export class CompileTabLogic {
   compileFile (target) {
     if (!target) throw new Error('No target provided for compiliation')
     return new Promise((resolve, reject) => {
-      this.api.readFile(target).then((content) => {
+      this.api.readFile(target).then(async(content) => {
         const sources = { [target]: { content } }
         this.event.emit('removeAnnotations')
         this.event.emit('startingCompilation')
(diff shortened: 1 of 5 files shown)


libs/remix-ui/solidity-compiler/src/lib/logic/compileTabLogic.ts
```diff
@@ -106,10 +106,15 @@ export class CompileTabLogic {
   compileFile (target) {
     if (!target) throw new Error('No target provided for compiliation')
     return new Promise((resolve, reject) => {
-      this.api.readFile(target).then((content) => {
+      this.api.readFile(target).then(async(content) => {
         const sources = { [target]: { content } }
         this.event.emit('removeAnnotations')
         this.event.emit('startingCompilation')
+        if(!this.useFileConfiguration && await this.api.fileExists('remappings.txt')) {
```
yann300, line 113: even if the fileConfiguration is used, maybe still relevant to  use`remappings.txt`. what's the behavior in foundry?

yann300: I mean in foundry who has the priority, remapping.txt or the compileConfig, or is this a merge of both?

Aniket-Engg: Foundry picks the remappings from remappings.txt and foundry.toml both. CompilerConfig is specific to Remix

yann300 (approved):

PR #4245, opened 2023-11-15
support log transaction from iframe plugin

drafish:
I'm developing an iframe plugin for xinghuo bif, and I want to send a transaction log, and I got some issues.

### Here is the first one.
image
Check the code here - https://github.com/ethereum/remix-project/blob/master/libs/remix-ui/terminal/src/lib/components/Context.tsx#L19

`provider` will be missing if the transaction log is called from iframe plugin. 
### Here is another one.
The decoded input does not convert correctly when it's BigNumber. 
image
Here is the correct convert result. 
image
BigNumber will turn to a ordinary object if it's passed through postMessage. 

@Aniket-Engg @ryestew

### DIFF
--- a/libs/remix-ui/terminal/src/lib/remix-ui-terminal.tsx
+++ b/libs/remix-ui/terminal/src/lib/remix-ui-terminal.tsx
@@ -116,7 +116,7 @@ export const RemixUiTerminal = (props: RemixUiTerminalProps) => {
         }
         scriptRunnerDispatch({
           type: message.type ? message.type : 'log',
-          payload: { message: [message.value], provider: message.provider },
+          payload: { message: [message.value] },
         })
       },
     })
--- a/libs/remix-ui/terminal/src/lib/remix-ui-terminal.tsx
+++ b/libs/remix-ui/terminal/src/lib/remix-ui-terminal.tsx
@@ -116,7 +116,7 @@ export const RemixUiTerminal = (props: RemixUiTerminalProps) => {
         }
         scriptRunnerDispatch({
           type: message.type ? message.type : 'log',
-          payload: { message: [message.value] },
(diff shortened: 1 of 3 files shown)


libs/remix-ui/terminal/src/lib/remix-ui-terminal.tsx
```diff
@@ -116,7 +116,7 @@ export const RemixUiTerminal = (props: RemixUiTerminalProps) => {
         }
         scriptRunnerDispatch({
           type: message.type ? message.type : 'log',
-          payload: { message: [message.value] },
```
yann300, line 119: I don't think we should do that here.
What are the input parameters?
You could have a default behavior when provider is not present.

drafish: Here is an example of input parameters. @yann300 
```js
{
  type: 'knownTransaction',
  value: {
    tx: {
      hash: resp.detail.hash,
      isCall: false,
      blockNumber: resp.detail.ledger_seq,
      transactionCost: resp.detail.actual_fee,
      from: resp.detail.transaction.source_address,
      to: null,
      transactionIndex: resp.detail.transaction.nonce,
    },
    receipt: { status: '0x1', contractAddress: resp.detail.contractAddress },
    resolvedData: {
      contractName: selectedContract.name,
      to: null,
      fn: '(constructor)',
      params: eventsDecoder._decodeInputParams(getinputParameters(data.dataHex.replace(data.contractBytecode, '')), txHelper.getConstructorInterface(selectedContract.abi)),
      contractAddress: resp.detail.contractAddress,
    },
    logs: logs,
  },
  provider: 'bif',
}
```

yann300 (approved):

PR #4260, opened 2023-11-21
fix using checkRevertBeforeSending and ignoreGasPricing

yann300:
fix https://github.com/ethereum/remix-project/issues/4252


libs/remix-simulator/src/VmProxy.ts
```diff
@@ -123,6 +123,7 @@ export class VmProxy {
     this.stateCopy = await this.vm.stateManager.copy()
     this.incr++
     this.processingHash = bufferToHex(data.hash())
+    console.log('txWillProcess', this.processingHash )
```
Aniket-Engg, line 126: Do you want to keep logs?

Aniket-Engg (approved):

PR #4269, opened 2023-11-24
Add slide in Featured section to show latest release highlights

Aniket-Engg:


LianaHus:
1. we need a new image. should necessary be a new one per release but should be different from the second tab's image
2. please fill in the text before merging (you can use previous release highlights)
3. I believe the link should lead to the blog post not the github release

LianaHus:
you could use https://remix-project.org/static/media/remix-ide-mobile.1c7e5b45e55d17968d778cae5dd1949d.svg for now. but please create a webp first

vermouth22:
At this point, it is still good for "feature requests" and qualitative
feedback, so we should re-characterize that as a unique ask. But, the
timeframe for the capture of a "snapshot" of user level numbers can't
really last beyond a "season" or quarter. There are too many other
variables that occur over time that introduce noise to the result, and it
becomes too difficult to account for the noise.

LianaHus:
I'm approving the pr, but we will change the image later. Also, we were discussing to have 3 highlights, not 4 to not mess with height(out-of-element scope button). We will try to have smaller releases anyway that should be enough.

LianaHus (approved):

PR #4309, opened 2023-12-06
fix parser: remappings

yann300:
This allows our solidity parser to use the `remappings.txt` file.

To test it, create a workspace if type v4 periphery, update submodules, it will show an error in the editor at the top.

### DIFF
--- a/apps/remix-ide/src/app/plugins/parser/services/code-parser-compiler.ts
+++ b/apps/remix-ide/src/app/plugins/parser/services/code-parser-compiler.ts
@@ -142,6 +142,14 @@ export default class CodeParserCompiler {
         this.compiler.set('runs', state.runs)
         this.compiler.set('useFileConfiguration', true)
         this.compiler.set('compilerRetriggerMode', CompilerRetriggerMode.retrigger)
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/plugins/parser/services/code-parser-compiler.ts
```diff
@@ -142,6 +142,12 @@ export default class CodeParserCompiler {
         this.compiler.set('runs', state.runs)
         this.compiler.set('useFileConfiguration', true)
         this.compiler.set('compilerRetriggerMode', CompilerRetriggerMode.retrigger)
+
+        if (await this.plugin.call('fileManager', 'exists','remappings.txt')) {
+          const remappings = await this.plugin.call('fileManager', 'readFile','remappings.txt')
+          this.compiler.set('remappings', remappings.split('\n').filter(Boolean))
+        }
```
Aniket-Engg, line 149: Please add an else section to reset the state which helps when remapping file is deleted.

Aniket-Engg (approved):

PR #4319, opened 2023-12-08
add custom endpoint for AI suggestion

yann300:
create a script in Remix:
like
```
(async () => {
  try {
    remix.call('copilot-suggestion' as any, 'useRemoteService', 'http://replit.remixcompletion.pro/infer')
    console.log('done')
  } catch (e) {
    console.log(e.message)
  }
})()
```
and run it to use a remote endpoint instead of the in-browser model.

### DIFF
--- a/apps/remix-ide/src/app/plugins/copilot/suggestion-service/copilot-suggestion.ts
+++ b/apps/remix-ide/src/app/plugins/copilot/suggestion-service/copilot-suggestion.ts
@@ -1,18 +1,20 @@
 import {Plugin} from '@remixproject/engine'
 import {SuggestionService, SuggestOptions} from './suggestion-service'
+import axios, {AxiosResponse} from 'axios'
 const _paq = (window._paq = window._paq || []) //eslint-disable-line
 
 const profile = {
   name: 'copilot-suggestion',
   displayName: 'copilot-suggestion',
   description: 'Get Solidity suggestions in editor',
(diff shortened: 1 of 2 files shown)


STetsing:
@yann300  where am I supposed to execute that JS?

yann300:
@STetsing you have to run it from Remix.

STetsing (commented):
The results is not displayed in the editor, Increasing the request timeout (if existent) could help

STetsing:
> > * The replit endpoints works well
> > * the starcoder endpoint ends up with a `   Failed to load resource: net::ERR_FAILED starcoder.remixcompletion.pro/infer`
> 
> the second issue looks like a server issue, is this the case?

no, the server returns a payload indeed, what is the request max timeout? the min inference time for starcoder ti max_new_word=5 is 6-5 seconds (mean value).

yann300:
> what is the request max timeout

I think this can be configured by passing some options to https://github.com/ethereum/remix-project/pull/4319/files#diff-456892ab56ea0fe769bc1bdb2c0ece2855f1ce80583e2309939106e9cf727f0eR68 but I don't know the default value..

STetsing:
The endpoints works well after reducing the max number of tokens (especially for the starcoder model). The timeout value must be increased for large generation requests

STetsing (approved):
Working as Is

PR #4333, opened 2023-12-12
always delete the code-sample workspace but propose to turn it into a normal workspace

yann300:
- load remix using a url params.
 - this will open the file in the `code-sample` workspace.
 - a warning icon (top of the FE) signigy that the workspace won't be persisted. 
 - click here will rename the workspace to persist the file.
 - reload remix, the `code-sample` workspace should be removed.

### DIFF
--- a/apps/remix-ide/src/app/tabs/locales/en/filePanel.json
+++ b/apps/remix-ide/src/app/tabs/locales/en/filePanel.json
@@ -8,6 +8,7 @@
   "filePanel.restore": "Restore",
   "filePanel.workspace.create": "Create Workspace",
   "filePanel.workspace.rename": "Rename Workspace",
+  "filePanel.workspace.save_workspace": "Save Workspace",
   "filePanel.workspace.delete": "Delete Workspace",
   "filePanel.workspace.deleteConfirm": "Are you sure to delete the current workspace?",
   "filePanel.workspace.download": "Download Workspace",
@@ -129,5 +130,5 @@
   "filePanel.movingFolderFailed": "Moving Folder Failed",
   "filePanel.movingFolderFailedMsg": "Unexpected error while moving folder: {src}",
   "filePanel.workspaceActions": "Workspace actions",
-  "filePanel.saveCodeSample": "This code sample will not be persisted. Please click here to save this code as a persisted workspace."
+  "filePanel.saveCodeSample": "This code-sample workspace will not be persisted. Click here to save it."
 }
--- a/apps/remix-ide/src/app/tabs/locales/en/filePanel.json
+++ b/apps/remix-ide/src/app/tabs/locales/en/filePanel.json
@@ -128,5 +128,6 @@
   "filePanel.movingFileFailedMsg": "Unexpected error while moving file: {src}",
   "filePanel.movingFolderFailed": "Moving Folder Failed",
   "filePanel.movingFolderFailedMsg": "Unexpected error while moving folder: {src}",
-  "filePanel.workspaceActions": "Workspace actions"
+  "filePanel.workspaceActions": "Workspace actions",
+  "filePanel.saveCodeSample": "This code sample will not be persisted. Please click here to save this code as a persisted workspace."
 }
--- a/libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
+++ b/libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
@@ -1,7 +1,7 @@
 import React, {useState, useEffect, useRef, useContext, ChangeEvent} from 'react' // eslint-disable-line
 import {FormattedMessage, useIntl} from 'react-intl'
 import {Dropdown} from 'react-bootstrap'
-import {CustomIconsToggle, CustomMenu, CustomToggle, extractNameFromKey, extractParentFromKey} from '@remix-ui/helper'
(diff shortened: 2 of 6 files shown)


LianaHus:
there is an error in browser

yann300:
> there is an error in browser

You have to tell me a bit more. which error? when?

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -937,6 +937,14 @@ export function Workspace() {
                     <label className="pl-2 form-check-label" style={{wordBreak: 'keep-all'}}>
                       <FormattedMessage id='filePanel.workspace' />
                     </label>
+                    {selectedWorkspace && selectedWorkspace.name === 'code-sample' && <CustomTooltip
+                      placement="right"
+                      tooltipId="saveCodeSample"
+                      tooltipClasses="text-nowrap"
+                      tooltipText={<FormattedMessage id="filePanel.saveCodeSample" />}
+                    >
+                      <i onClick={() => renameCurrentWorkspace()} className="far fa-exclamation-triangle text-info ml-2 mt-1" aria-hidden="true"></i>
```
joeizang, line 946: I think the exclamation should be either be text-warning or danger. In dark mode, it's contrast isn't great. The size could be a bit bigger though.

joeizang (commented):
Reloading the page doesn't nuke the code-sample workspace. I had to remove the url param before it worked.

yann300:
> Reloading the page doesn't nuke the code-sample workspace. I had to remove the url param before it worked.

ah yes sry I didn't mention that, the URL param should be removed.

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
LianaHus: please use align-self-center instead of mt-1.
I agree with Joe, it should rather be text-warning.

LianaHus:
> > there is an error in browser
> 
> You have to tell me a bit more. which error? when?

CacheFS.js:115 Uncaught (in promise) Error: ENOENT: /.workspaces/code-sample
    at CacheFS._lookup (CacheFS.js:115:1)
    at CacheFS.lstat (CacheFS.js:216:1)
    at DefaultBackend.unlink (DefaultBackend.js:138:1)
    at PromisifiedFS.unlink (PromisifiedFS.js:161:1)
    at PromisifiedFS.unlink (PromisifiedFS.js:118:1)
﻿

libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -943,7 +943,7 @@ export function Workspace() {
                       tooltipClasses="text-nowrap"
                       tooltipText={<FormattedMessage id="filePanel.saveCodeSample" />}
                     >
-                      <i onClick={() => renameCurrentWorkspace()} className="far fa-exclamation-triangle text-info ml-2 mt-1" aria-hidden="true"></i>
+                      <i onClick={() => renameCurrentWorkspace()} className="far fa-exclamation-triangle text-warning ml-2 mt-1" aria-hidden="true"></i>
```
LianaHus, line 946: mt-1 is still there

yann300: it's done.

apps/remix-ide/src/app/tabs/locales/en/filePanel.json
```diff
@@ -127,5 +127,6 @@
   "filePanel.movingFileFailedMsg": "Unexpected error while moving file: {src}",
   "filePanel.movingFolderFailed": "Moving Folder Failed",
   "filePanel.movingFolderFailedMsg": "Unexpected error while moving folder: {src}",
-  "filePanel.workspaceActions": "Workspace actions"
+  "filePanel.workspaceActions": "Workspace actions",
+  "filePanel.saveCodeSample": "This code sample will not be persisted. Please click here to save this code as a persisted workspace."
```
ryestew, line 131: This code sample will not be persisted. Please click here to make it a saved workspace.

ryestew (changes requested):

apps/remix-ide/src/app/tabs/locales/en/filePanel.json
Aniket-Engg, line 131: This code-sample workspace will not be persisted. Click here & rename to save it.
or
This is a temporary workspace. Click here & rename to save it.

Aniket-Engg (commented):
When I open Remix with a `#code=base64Content...` , it opens FE, shows file in editor and then moves to Solidity Compiler.

At the end of loading user sees Solidity Compiler plugin which can make is difficult to inform the user about temporary workspace

yann300:
> When I open Remix with a `#code=base64Content...` , it opens FE, shows file in editor and then moves to Solidity Compiler.
> 
> At the end of loading user sees Solidity Compiler plugin which can make is difficult to inform the user about temporary workspace

ok I see, well if the user switch back to the FE  it will be displayed.
tbh I won't really bother for this. Also usually this loads some documentation which isn't meant to be persisted

LianaHus:
"This code-sample workspace will not be persisted. Click here to save it."

Screenshot from 2024-01-08 17-04-19

Aniket-Engg (approved):

LianaHus (approved):

PR #4367, opened 2023-12-21
add cookbook template

yann300:


libs/remix-ui/home-tab/src/lib/components/homeTabGetStarted.tsx
```diff
@@ -61,14 +61,23 @@ function HomeTabGetStarted({plugin}: HomeTabGetStartedProps) {
     let templateDisplayName = TEMPLATE_NAMES[templateName]
     const metadata = TEMPLATE_METADATA[templateName]
     if (metadata) {
-      await plugin.call('dGitProvider', 'clone', {url: metadata.url, branch: metadata.branch}, templateDisplayName)
-      return
+      if (metadata && metadata.type === 'git') {
```
Aniket-Engg, line 64: I think `metadata && ` can be removed from both the plugins

libs/remix-ui/workspace/src/lib/utils/constants.ts
```diff
@@ -87,15 +87,33 @@ export const TEMPLATE_NAMES = {
...
+  },
+  'uniswapV4HookBookMultiSigSwapHook': {
+    type: 'plugin',
+    name: 'cookbookdev',
+    endpoint: 'openPattern',
+    params: ['Uniswap-V4-HookBook-MultiSigSwapHook', true]
   }
 }
+
+export type TemplateType = {
```
Aniket-Engg, line 112: Not sure if this is a right file to add a type def

Aniket-Engg:
Screenshot 2024-01-08 at 4 56 49 PM

May be we need to do something with the template name

Aniket-Engg:
editor by default open `src/MULTI_SIG/MultiSigSwapHook.sol` which can not be successfully compiled on Remix

yann300:
> editor by default open `src/MULTI_SIG/MultiSigSwapHook.sol` which can not be successfully compiled on Remix

it should compile after the plugin is updated.

libs/remix-ui/home-tab/src/lib/components/homeTabGetStarted.tsx
```diff
@@ -61,14 +61,23 @@ function HomeTabGetStarted({plugin}: HomeTabGetStartedProps) {
     let templateDisplayName = TEMPLATE_NAMES[templateName]
     const metadata = TEMPLATE_METADATA[templateName]
     if (metadata) {
-      await plugin.call('dGitProvider', 'clone', {url: metadata.url, branch: metadata.branch}, templateDisplayName)
-      return
+      if (metadata.type === 'git') {
+        await plugin.call('dGitProvider', 'clone', {url: metadata.url, branch: metadata.branch}, templateDisplayName)
+      } else if (metadata && metadata.type === 'plugin') {
```
Aniket-Engg, line 66: Same for this condition, `metadata && ` can be removed

Aniket-Engg (approved):

PR #4372, opened 2023-12-22
Script to create a PR for beta using existing PR number

Aniket-Engg:


bunsenstraat (approved):
if we use ts-node or npx ts-node we can ts the file? not important but might be more consistent to have ts as much as we can.

yann300 (approved):

Aniket-Engg:
> if we use ts-node or npx ts-node we can ts the file? not important but might be more consistent to have ts as much as we can.

Done

PR #4383, opened 2023-12-29
Low Level Icon link Fix

### LINKED ISSUE #4359: Low level interactions i icon links to old solidity version
the icon here:
image

needs to link to the latest solidity version not to 0.6.2

joeizang:
fixes #4359


libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
```diff
@@ -284,11 +285,12 @@ export function UniversalDappUI(props: UdappProps) {
             <div className="py-2 border-top d-flex justify-content-start flex-grow-1">
               <FormattedMessage id="udapp.lowLevelInteractions" />
             </div>
-            <CustomTooltip placement={'bottom-end'} tooltipClasses="text-wrap" tooltipId="receiveEthDocstoolTip" tooltipText={<FormattedMessage id="udapp.tooltipText8" />}>
+            {/* <CustomTooltip placement={'bottom-end'} tooltipClasses="text-wrap" tooltipId="receiveEthDocstoolTip" tooltipText={<FormattedMessage id="udapp.tooltipText8" />}>
               <a href="https://solidity.readthedocs.io/en/v0.6.2/contracts.html#receive-ether-function" target="_blank" rel="noreferrer">
```
ryestew, line 289: should this part be erased?  I think you have it updated in lowLevelInteractionIcon.tsx

ryestew (commented):
I'm not sure which is better - specifying the version number or using this url with the "latest" in it:
https://docs.soliditylang.org/en/latest/contracts.html#receive-ether-function

joeizang:
> I'm not sure which is better - specifying the version number or using this url with the "latest" in it: https://docs.soliditylang.org/en/latest/contracts.html#receive-ether-function

It won't matter. This will get the version of the compiler that is loaded.

libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
joeizang: You are right. This should be nuked.

ryestew (approved):

LianaHus (approved):

PR #4404, opened 2024-01-08
Vyper Plugin UI Updates

### LINKED ISSUE #4395: Vyper plugin updates
- [x] The default view that comes up has the warning:  "we only support vyper < 2.6" -  this needs to be  removed. 
- [x] The default view ls the local compiler - but should be switched to the remote compiler to be the default

- [ ] The active state for the remote vs local compiler looks in the dark theme reversed - I mean the lighter background looks like its the one in the active state - but it is not the active one  - see the image below...

image

- [ ] In the light theme there is not much  differentiations between these buttons:

image

These could be switched to radio buttons for the moment.  Then we wouldn't need to monkey with getting the backgrounds to work in all the themes and it would be clear which one is in the active state.

- [x] As stated above , the warning text for the local compiler should be removed.
- [x] The warning text of the remote compiler should be removed.
- [x] In a near term update, a user will be able to choose the vyper version of the remote compiler - right?  But for now it is 0.3.10 - right?
- [x] Clone Vyper lang repo needs a tool tip to tell users to go to FE to see the files
- [ ] Other design features of the Sol compiler should be incorporated.
- [x] The profile name should be Vyper compiler and the Vyper Compiler line with the logo should be removed.
- [x] The documentation in the plugin needs to be in the remix-ide docs not the old Vyper plugin.

joeizang:
fixes #4395


apps/vyper/src/app/app.tsx
```diff
@@ -30,7 +31,7 @@ const App: React.FC = () => {
   const [output, setOutput] = useState<any>({})
   const [state, setState] = useState<AppState>({
     status: 'idle',
-    environment: 'local',
+    environment: 'remote',
```
yann300, line 34: why did you change the label to `remote`.

ryestew (approved):

PR #4446, opened 2024-01-11
FE hover icons

joeizang:
Adds icons to file explorer when hovered over. Offers basic shortcut icons to create, rename and delete files or folders

### DIFF
--- a/libs/remix-ui/workspace/src/lib/components/file-explorer-hovericons.tsx
+++ b/libs/remix-ui/workspace/src/lib/components/file-explorer-hovericons.tsx
@@ -1,0 +1,100 @@
+import React, { useState } from 'react'
+import { CustomTooltip } from '@remix-ui/helper'
+import { FormattedMessage } from 'react-intl'
+import { ROOT_PATH } from '../utils/constants'
+
+export type FileHoverIconsProps = {
+  hover: boolean
+  isEditable: boolean
+  file: any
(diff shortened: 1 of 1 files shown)


LianaHus (changes requested):
Looks very good!

Some small notes from my side:

when I try to rename and then click on something else (abort it), I see the modal "A file or folder 1_Storage.sol already exists at this location. Please choose a different name." If the name staid the same it shouldn't be shown IMHO

libs/remix-ui/workspace/src/lib/css/file-explorer.css
```diff
@@ -60,4 +60,16 @@ ul {
...
+
+.remixui_icons {
+
+}
+
+.remixui_icons:hover {
+    color: var(--text);
+}
+
+.remixui_icons_space {
```
LianaHus, line 73: can't you use bootstrap class here?

joeizang: you are right. mr-2 does the trick

yann300:
> when I try to rename and then click on something else (abort it), I see the modal "A file or folder 1_Storage.sol already exists at this location. Please choose a different name." If the name staid the same it shouldn't be shown IMHO

@LianaHus this is fixed now

yann300 (approved):

PR #4448, opened 2024-01-12
Copy share URL for file

Aniket-Engg:


libs/remix-ui/workspace/src/lib/actions/index.ts
```diff
@@ -373,6 +374,33 @@ export const copyFile = async (src: string, dest: string) => {
...
 }
 
+export const copyShareURL = async (path: string) => {
+  const fileManager = plugin.fileManager
+
+  try {
+    const host = '127.0.0.1'
+    const port = 5001
+    const protocol = 'http'
+    // const projectId = ''
```
yann300, line 384: should you remove this?

yann300 (approved):

libs/remix-ui/workspace/src/lib/actions/index.ts
Aniket-Engg: this will be required while connecting with actual node

Aniket-Engg:
As discussed, I have hidden the option from context menu as of now.

PR #4487, opened 2024-01-24
Update WalletConnect's Web3Modal SDK to the latest version

glitch-txs:
- Update Web3Modal SDK to v3.5.7
- It add support for EIP-6963
- Added support for more chains
- Removed wagmi & viem (they were only used by Web3Modal v2) in favor of ethers 5: Ethers was already installed and Wagmi v2 was having some conflicts in the bundle level.
Remix-ide-ethers


ioedeveloper:
Thanks for the PR @glitch-txs .
I just found one issue while testing it. The network and account balance doesn't get updated on the plugin after switching network on metamask (chrome browser extension) to another network (E.g switching from Sepolia to Mainnet).
Screenshot 2024-01-25 at 13 36 55

glitch-txs:
Hi @ioedeveloper thanks for pointing this out, it seems like there's an issue with the modal when the chain si not registered, could you confirm this? from Ethereum to Sepolia it works fine on my end but for Linea (or any not configured chain) it is not working.

Could you confirm is this is the case? I will bring this to our engineer team and upgrade the fix. Thanks! 🙏

ioedeveloper:
> Could you confirm if this is the case? I will bring this to our engineer team and upgrade the fix. Thanks! 🙏

@glitch-txs This is not the case for me. It also fails for registered chains. I get the error below on the browser console when i switch chains from the mobile wallet (ios) metamask.
Screenshot 2024-01-30 at 12 48 49

glitch-txs:
Thanks for sharing this @ioedeveloper, we found the issue in one of our low level API's, we'll fix it and update this repo with a new version with this fixed. 👍

glitch-txs:
hi @ioedeveloper thanks again for your feedback, we have fixed the bug and I've update it Web3Modal to the latest version :)
Could you take a look again please?

Thanks! 👍

ioedeveloper (approved):

PR #4512, opened 2024-02-02
Fix circom ref

yann300:
- fix a possible issue with absolute path in circom.
 - add sindri template.
 - nr extension in monaco editor.

### DIFF
--- a/apps/circuit-compiler/src/app/services/circomPluginClient.ts
+++ b/apps/circuit-compiler/src/app/services/circomPluginClient.ts
@@ -54,6 +54,7 @@ export class CircomPluginClient extends PluginClient {
       fileContent = await this.call('fileManager', 'readFile', path)
     }
     this.lastParsedFiles = await this.resolveDependencies(path, fileContent)
+    console.log(this.lastParsedFiles)
     const parsedOutput = this.compiler.parse(path, this.lastParsedFiles)
 
     try {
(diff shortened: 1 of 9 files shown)


libs/remix-ws-templates/src/script-templates/sindri/sindri.conf
```diff
@@ -0,0 +1,8 @@
+{
+    "$schema": "https://forge.sindri.app/api/v1/sindri-manifest-schema.json",
```
KPreisner, line 2:
```suggestion
    "$schema": "https://sindri.app/api/v1/sindri-manifest-schema.json",
```
Small tweak. We updated our domains several weeks ago from forge.sindri.app to just sindri.app and we still have some work to do with all of our docs and where this may have been pulled from. It's ok, forge.sindri.app redirects to sindri.app for now.

apps/circuit-compiler/src/app/services/circomPluginClient.ts
```diff
@@ -313,15 +313,15 @@ export class CircomPluginClient extends PluginClient {
...
           // Regular expression to match include statements and make deps imports uniform
           const includeRegex = /include "(.+?)";/g
-          const replacement = 'include "circomlib/circuits/$1";'
+          const replacement = 'include "/circomlib/circuits/$1";'
 
           dependencyContent = dependencyContent.replace(includeRegex, replacement)
         } else {
           if (!include.startsWith('circomlib') && !pathModule.isAbsolute(filePath) && !pathModule.isAbsolute(path)) {
           // if include is not absolute, resolve it using the parent path of the current file opened in editor
             const absIncludePath = pathModule.resolve('/' + filePath.slice(0, filePath.lastIndexOf('/')), '/' + path)
```
Aniket-Engg, line 322: `absIncludePath ` already has a `/` prepended here why do you know it here too: `/${absIncludePath}`

Aniket-Engg (commented):
After clicking on option from hamburger, toaster says `scripts added in scripts folder` but after analyzing this code, I got to know there are files `.sindriignore` and `sindri.json` were added in root of project

Aniket-Engg (approved):

PR #4524, opened 2024-02-09
add save icon

Aniket-Engg:


libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
```diff
@@ -1,22 +1,79 @@
...
+
+  useEffect(() => {
+    const fetchSavedContracts = async () => {
+      env.current = await props.plugin.call('blockchain', 'getProvider')
+      if(env.current && env.current === 'injected') {
+        const allSavedContracts = localStorage.getItem('savedContracts')
+        if (allSavedContracts) {
+          const savedContracts = JSON.parse(allSavedContracts)
+          const { network } = await props.plugin.call('blockchain', 'getCurrentNetworkStatus')
+          if (savedContracts[env.current] && savedContracts[env.current][network.id]) {
```
yann300, line 20: The saved contracts should probably be independent of the current provider (metamask, external provider, ..)

libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
```diff
@@ -1,22 +1,79 @@
...
 import { UniversalDappUI } from './universalDappUI'
 
 export function InstanceContainerUI(props: InstanceContainerProps) {
   const { instanceList } = props.instances
+  const env = useRef()
+
+  useEffect(() => {
+    const fetchSavedContracts = async () => {
+      env.current = await props.plugin.call('blockchain', 'getProvider')
+      if(env.current && env.current === 'injected') {
```
yann300, line 15: could this be for all providers which aren't a Remix VM?

libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
```diff
@@ -1,22 +1,79 @@
...
+        const allSavedContracts = localStorage.getItem('savedContracts')
+        if (allSavedContracts) {
+          const savedContracts = JSON.parse(allSavedContracts)
+          const { network } = await props.plugin.call('blockchain', 'getCurrentNetworkStatus')
+          if (savedContracts[env.current] && savedContracts[env.current][network.id]) {
+            const instances = savedContracts[env.current][network.id]
```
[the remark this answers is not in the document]
Aniket-Engg: As discussed, file path including workspace name will be displayed for saved contract

libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
Aniket-Engg: this is done

libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
Aniket-Engg: This should be discussed

yann300:
Saved contracts don't work while using the `AtAddress` feature

yann300:
when switching networks, only the saved contracts from the current network should be displayed.

Aniket-Engg:
> when switching networks, only the saved contracts from the current network should be displayed.

this is done

Aniket-Engg:
> 1. As we discussed previously in the initial call, "Currently, you have no saved contracts to interact with." It takes too much space. the one for deployed contracts was fine as long as there was nothing shown after that. now we have 2 huge fields there that attract too much unnecessary attention. Maybe try to remove the alert and leave a text warning or info at least saved once.
> 2. There are issues in the flow.
> 
> * First of all, once the user saves, we remove the TX from the second list. it is confusing and we destroy the order.
> * Second - once we remove it from the "saved" list, it doesn't come back to the "Deployed.." list but disappears.

point 2 is done

Aniket-Engg:
> Saved contracts don't work while using the `AtAddress` feature

I think this feature is also working

yann300 (approved):

PR #4526, opened 2024-02-10
Vyper plugin UI Updates

### LINKED ISSUE #4421: Vyper UI updates
- [ ] Vyper docs need to be in our repo - next release
- [x] Text suggesting to put the compiler version & EVM version in the .vy file like "Specify the compiler version & EVM version in the .vy file" - next release
- [x] New vyper logo should be added https://github.com/pentcle/vyper-brand - update fa-vyper2 with new logo & make webp version next release
- [x] the cursor should be an arrow on the active compiler type button and on the inactive compiler (remote or local) the should be cursor pointer - (the one with the finger) - next release

- [x] add copy abi & bytecode buttons like in sol compiler  - next release
- [ ] replace vyper-lang for cloning with this repo: https://github.com/pcaversaccio/snekmate - next release

- [x] error handling should be shown in cards like in the sol compiler - next release

- [ ] ctrl + s for compilation on a .vy file - after next release
- [ ] play button for compiling on a .vy file - after next release

- [ ] When cloning - the FE should have a spinner to indicate that the files are loading (OK this is not Vyper issue but a FE Cloning issue). But this should not be as much an issue with snekmate

Item for later: a select box of Vyper compilers and of EVM versions (this is not needed yet ) - coordinate with Ape about when the manifest can handle this.

joeizang:
fixes #4421

### DIFF
--- a/apps/vyper/src/app/app.tsx
+++ b/apps/vyper/src/app/app.tsx
@@ -78,7 +78,7 @@ const App = () => {
             placement="bottom"
             tooltipText="Clone Vyper examples. Switch to the File Explorer to see the examples."
           >
-            <Button data-id="add-repository" className="w-100 text-dark bg-light btn-outline-primary " onClick={() => remixClient.cloneVyperRepo()}>
+            <Button data-id="add-repository" className="w-100 text-dark bg-light btn-secondary" onClick={() => remixClient.cloneVyperRepo()}>
             Clone Vyper examples repository
             </Button>
           </CustomTooltip>


apps/vyper/src/app/app.tsx
```diff
@@ -94,27 +99,57 @@ const App = () => {
     setOutput(remixClient.compilerOutput)
   }
 
+  console.log('output', output)
```
yann300, line 102: this should probably be removed

yann300 (approved):

PR #4527, opened 2024-02-12
Improve Sindri Integration with the IDE

sangaline:
This includes a collection of changes building on the initial `fix_circom_ref` integration which are primarily focused on making the user experience smoother and requiring less manual integration on their part. I tried to balance those goals against not overly pushing the service on users or disrupting their workflows if they don't use the service. This was a bit of a challenge, and I very much understand if you think I didn't strike the right balance and want to pull back on any of these changes or implement them in a different way. Hopefully, this can at least serve as a starting point for discussion about what the integration should look like and how it will evolve moving forward. I'm also more familiar with the codebase now than I was previously, so that should make conversation and planning easier.

### Enable Schema Request Fetching in JSON Files

This allows the editor to fetch a JSON Schema definition when a `$schema` field is set in the file, and in turn use this schema for in-editor diagnostics. This can help users catch basic configuration issues with `sindri.json` locally before uploading the package for remote compilation.

image

### Configure Sindri API Key in Settings

Manually entering the API key into scripts is non-ideal for two reasons: 1) it makes it easy to accidentally introduce the API key into a git repo, publish it in a gist, or include it in a package upload to Sindri, and 2) it requires additional manual steps for each circuit. I added this to the global settings here and removed the need to update the scripts to specify the API key. This seemed like the path of least resistance, but I feel like this might be more appropriate as part of a future Sindri plugin instead of the global settings. I'm very open to feedback on both how we want things to work in the first release of an integration and where we want things to head in future versions.

image

### Add `run_compile.ts` and `run_prove.ts` Scripts

I added these scripts under the `scripts/sindri/` directory in an attempt to give more convenient script entry points for these actions, and to separate out some of the implementation details that users hopefully won't need to worry about or look at too closely (*e.g.* mapping the workspace files into a `File` array). I moved the heavier functions into `scripts/sindri/utils.ts`. The `run_compile.ts` should work generically across circuits, the `run_prove.ts` script requires manually setting the circuit inputs in the file and logs out a warning if you haven't done so. My hope is that these make it a little more intuitive how to get started with compiling and proving remotely.

### Add `sindri.json` and `.sindriignore` to Existing Circom Templates

In order for the circuit to compile in Sindri with the `circuits/*.circom` project structure, a `circuitPath` field in `sindri.json` is required. It's also helpful to set a unique and meaningful project name so that proofs can be generated without juggling circuit IDs. The project name is analogous to a DockerHub image name or GitHub repository name, and you can generate proofs by referencing `semaphore:latest` or `sempahore:v0.1.0` in place of the circuit ID. By pre-configuring  `sindri.json` with these values, the `run_compile.ts` and `run_prove.ts` scripts require less customization for a specific circuit and the circuit is able to compile remotely without manual modifications to `sindri.json`. Here's an example of the customizations for the semaphore circuit:

```json
{
  "$schema": "https://sindri.app/api/v1/sindri-manifest-schema.json",
  "name": "semaphore",
  "circuitPath": "./circuits/semaphore.circom",
  "circuitType": "circom",
  "curve": "bn254",
  "provingScheme": "groth16",
  "witnessCompiler": "wasm"
}
```

I had mixed feelings about doing this because I can see how it does add noise for users who aren't using the Sindri service. We do have in-progress plans to make the `sindri` CLI and `sindri.json` project configuration useful outside of our API service by introducing a unified abstraction over linting, compiling, and proving locally, but totally get it if you don't want to include these by default in the templates. Perhaps it would be possible to do something like make the inclusion of these files opt-in when you create the new workspace from the template?

image

It seemed like doing something like this might require deeper changes to the templating system. Another alternative might be to work it into "Add Sindri ZK scripts" such that it tries to auto-detect the Circom templates and customize what it pulls in. It feels a little hacky, but I think we could make it work relatively smoothly.

### Improve Automatic Project Packaging

The creation of the `File` array that's passed to the client library now uses `compileFolderToJson` to pull in all files in the workspace, and I smoothed out some rough edges around how the dependencies are pulled in. In particular, we don't support absolute import paths so I rewrote these to be relative to the project root. With these changes, any Circom package should now be packaged in a way that will compile successfully by default as long as `circuitPath` is set appropriately in `sindri.json`.


yann300:
Thanks for the PR!

A few questions:

- > but I feel like this might be more appropriate as part of a future Sindri plugin instead of the global settings 

Would you have plan to develop a plugin? 
Having the settings here would be ok for a start, but we should probably move it in a specified time.

- > Add sindri.json and .sindriignore to Existing Circom Templates

We prefer this to be an opt-in solution (like you demonstrated it in the screenshot).
But this modal should probably have a link to more information about the usage of sindri or (and this will be probably better) a learneth tutorial (we can also start the tutorial directly from remix). Here is a link which explain how to write a learneth tutorial: https://remix-learneth-plugin.readthedocs.io/en/latest/index.html

- could you add just the english translation? the rest should be added from crowdin. Here's a link to the remix UI crowdin: https://crowdin.com/project/remix-ui

yann300:
- where are you storing the `circuit_id` ?

ryestew:
For the translations, you only need to update the en/settings.json file - the rest of the languages are handled in our translation management platform- Crowdin:
https://crowdin.com/project/remix-ui

So please remove the other edits to the other languages' settings.json files.  

Once this PR is merged then we (you or our translators) can update the settings translation.

sangaline:
Thanks for the feedback, I like this latest version a lot more. It simultaneously makes the experience smoother for Sindri users while being less intrusive for non-Sindri users. There's more explanation in the inline responses below:


> Would you have plan to develop a plugin? Having the settings here would be ok for a start, but we should probably move it in a specified time.

Yeah, we're absolutely interested in this. We've seen a decent percentage of our users using Remix, and we're invested in making that experience as smooth as possible for them. I think a plugin opens up the door for us to build UI customizations that would make things easier and more intuitive for users. Also totally understand the global settings being a temporary measure and it making more sense to isolate things in a plugin. Do you have a specific timeline in mind? Things are going to be especially busy with ETH Denver coming up, but we would very much like to continue moving forward with development.

> We prefer this to be an opt-in solution (like you demonstrated it in the screenshot).

I removed all changes to the Circom templates in favor of adding more intelligence into the "Add Sindri ZK scripts" action. This produces an equivalent result in terms of the `sindri.json` customizations, but does it in a more generic way that should work well for any Circom circuit project. The new logic uses some heuristics to derive the circuit entrypoint and circuit name from the ambient workspace. I also looked into adding the checkbox when creating a new template, but it felt like it should only happen for the ZKP templates and I wasn't sure how to do this without getting kind of hacky in `remix-ui-workspace.tsx`. I'm open to doing that, but wanted to touch base first.

> a learneth tutorial (we can also start the tutorial directly from remix)

Nice, this looks like it would be great for getting users started. We're definitely interested in this as well, perhaps it's something we could do in conjunction with the move to a plugin? Another potential quick thing we could do to help users is add a `scripts/sindri/README.md` file when the scripts are added?

> So please remove the other edits to the other languages' settings.json files.

Done.

> where are you storing the circuit_id?

This currently isn't stored anywhere in this cut. When you generate a proof with Sindri, you can either pick an exact compilation by using the circuit ID or you can specify the most recent compilation for a given project and tag. We default to a tag of `latest` and the general flow is meant to be reminiscent of Docker image tags. With the changes I've made here, we can safely skip storing the circuit ID because we request a proof for `{projectName}:latest`. The project name is pulled from `sindri.json` in both cases, and I now derive a reasonable circuit name from the workspace name when you add the Sindri scripts.

yann300:
> Another potential quick thing we could do to help users is add a scripts/sindri/README.md file when the scripts are added?

yes that makes totally sense.

sangaline:
Just pushed up a README and I open it by default after clicking "Add Sindri ZK scripts." I think this will help a lot for people getting started, and I'm excited to build a Learneth tutorial as well.

yann300 (approved):

PR #4546, opened 2024-02-19
Fix search

yann300:
This fixes filtering transactions from the terminal (using the search box), this wasn't working before this PR.


Aniket-Engg (approved):

LianaHus:
this doesn't work for me

PR #4550, opened 2024-02-19
Improve gist

yann300:
with this PR:

 - loading a gist will put all the files and folder at the root (instead of in the folder `gist-{id}`)
 - saving a folder as a gist is still possible, the only way to update the gist would be to load it as a workspace.
 - I removed the `gist` type because it doesn't seem to be used at all now.


apps/remix-ide/src/app/tabs/locales/en/filePanel.json
```diff
@@ -62,10 +62,10 @@
   "filePanel.compileForNahmii": "Compile for Nahmii",
   "filePanel.createNewFile": "Create new file",
   "filePanel.createNewFolder": "Create new folder",
-  "filePanel.publishToGist": "Publish all files to GitHub gist",
+  "filePanel.publishToGist": "Publish workspace to GitHub gist",
   "filePanel.uploadFile": "Upload files",
   "filePanel.uploadFolder": "Upload folder",
-  "filePanel.updateGist": "Update the current [gist] explorer",
+  "filePanel.updateGist": "Publish workspace to GitHub gist",
```
LianaHus, line 68: is this not update anymore?

LianaHus (approved):

PR #4551, opened 2024-02-19
Terminalwrapper

bunsenstraat:


yann300:
seems this can be removed: `libs/remix-ui/terminal/src/lib/components/TerminalTitle.tsx`

yann300:
In desktop, the search box and other items are pushed back on the right side, I guess this is wanted?

yann300 (approved):

yann300:
see the dropdown list on the right, it always shows only 1 terminal..
 
remix

PR #4554, opened 2024-02-20
Solcoder/explain contract

STetsing:
- Added switch for AiI copilot. This enables the inline completion and `///` code generation
- Added tab button for Contract/.sol file summarization & explanation 
- right click on selection -> "Explain this code" using solcoder. get function explanation still works


yann300:
I feel the loading icon, left to the `AI running...` it too big.
@LianaHus ^

yann300:
it's not really clear what does the `AI copilot` ... and how it differentiate from the " solidity copilot " in the settings

yann300:
> Added switch for AiI copilot. This enables the inline completion and /// code generation

ah so the `solidity copilot` in the settings is meant to be replaced?

libs/remix-ui/editor/src/lib/suggestion-service/suggestion-service.ts
```diff
@@ -0,0 +1,89 @@
+import EventEmitter from 'events'
+
+export class SuggestionService {
```
yann300, line 3: I see you have another file named `suggestion-service.ts`..
This one or the other should probably be removed.

STetsing: > ah so the solidity copilot in the settings is meant to be replaced?
yes 

> I see you have another file named suggestion-service.ts..
This one or the other should probably be removed. 
I think it is till part of the original  browser model

yann300: ok, I think it's best to clean up what is no longer used, (copilot in the settings and the duplicate suggestion-service.ts), but perhaps you actually already planned that.

yann300:
remix
getting this when trying the code suggestion feature.

yann300:
when loading Remix, it shows directly `Solidity copilot deactivated!` , is this wanted?

yann300:
When hovering the action which activate the AI features, it says that the model is downloaded. There's no way to know that the model is actually downloaded. I actually get "Unexpected token '<' ,... " (see screenshot above). Does this error means taht the model wasn't downloaded correctly?

yann300:
How should we tell users which features come with enabling the AI? (msg in the tooltip, link to documentation, .. ?)

libs/remix-ui/editor/src/lib/remix-ui-editor.tsx
```diff
@@ -811,8 +834,13 @@ export const EditorUI = (props: EditorUIProps) => {
         currentFunction.current = functionImpl.name
         executeGptGenerateDocumentationAction.label = intl.formatMessage({id: 'editor.generateDocumentation2'}, {name: functionImpl.name})
         gptGenerateDocumentationAction = editor.addAction(executeGptGenerateDocumentationAction)
-        executegptExplainFunctionAction.label = intl.formatMessage({id: 'editor.explainFunction2'}, {name: functionImpl.name})
+        executegptExplainFunctionAction.label = intl.formatMessage({id: 'editor.explainFunction'}, {name: functionImpl.name})
```
yann300, line 837: This update the label to show only "Explain function". Should this keep using `editor.explainFunction2`?

libs/remix-ui/renderer/src/lib/renderer.tsx
```diff
@@ -101,7 +101,20 @@ export const Renderer = ({message, opt = {}, plugin}: RendererProps) => {
             <span className="ml-3 pt-1 py-1" >
               <CopyToClipboard content={messageText} className={` p-0 m-0 far fa-copy ${classList}`} direction={'top'} />
             </span>
-            <span className="border border-success text-success btn-sm" onClick={() => { askGtp() }}>ASK GPT</span>
+            <span
+              className="position-relative text-ai text-sm pl-0 pr-2"
+              style={{fontSize: "x-small", alignSelf: "end"}}
+            >
+              AI
```
yann300, line 108: remix
Are you sure about showing this like that. This looks pretty weird @LianaHus

STetsing: @yann300 

> getting this when trying the code suggestion feature

this error originates from the code completion server. This is returned when the server is down 

> when loading Remix, it shows directly Solidity copilot deactivated! , is this wanted?

Yes, this is desired. A log is also prompted whenever the user turns ON the copilot 

> When hovering the action which activate the AI features, it says that the model is downloaded. There's no way to know that the model is actually downloaded. I actually get "Unexpected token '<' ,... " (see screenshot above). Does this error means taht the model wasn't downloaded correctly?

Do you have an image with the "model downloaded"? Anything related to the inbrowser model should be discarded by now


> How should we tell users which features come with enabling the AI? (msg in the tooltip, link to documentation, .. ?) 

Maybe the link to the documentation will be more appropriate. Or get them to know it by discovering it intuitively

yann300:
could you add a matomo call for all the actions? (click on doc, explain code, explain contract, activate AI, deactivate AI, etc...)

yann300 (approved):

PR #4568, opened 2024-02-27
copilot ui and sync with settings

LianaHus:


STetsing (changes requested):
@LianaHus 
the ai-switch and the checkbox are still not synched and the completion model is downloaded automatically using saved state.

image

STetsing:
@LianaHus  furthermore, when the copilot is deactivated, the loading mechanism seems to be activated. 
image

STetsing (approved):

PR #4608, opened 2024-03-08
add dapp-draft icon in deployed contract

drafish:
add an edit icon in deployed contract to automatically input dapp parameters. 
edit-dapp 

@ryestew @yann300

### DIFF
--- a/libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
+++ b/libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
@@ -294,10 +294,11 @@ export function UniversalDappUI(props: UdappProps) {
       </div>
       <div className="udapp_cActionsWrapper" data-id="universalDappUiContractActionWrapper">
         <div className="udapp_contractActionsContainer">
-          <div className="d-flex" data-id="instanceContractBal">
+          <div className="d-flex justify-content-between" data-id="instanceContractBal">
             <label>
               <b><FormattedMessage id="udapp.balance" />:</b> {instanceBalance} ETH
             </label>
+            {props.exEnvironment === 'injected' && <i className="fas fa-edit btn btn-sm p-0" onClick={() => {props.editInstance(props.instance)}}></i>}
           </div>
           { props.isSavedContract && props.instance.savedOn ? (
             <div className="d-flex" data-id="instanceContractSavedOn">
--- a/libs/remix-ui/run-tab/src/lib/run-tab.tsx
+++ b/libs/remix-ui/run-tab/src/lib/run-tab.tsx
@@ -324,6 +324,10 @@ export function RunTabUI(props: RunTabProps) {
             runTransactions={executeTransactions}
             sendValue={runTab.sendValue}
             getFuncABIInputs={getFuncABIValues}
+            exEnvironment={runTab.selectExEnv}
+            editInstance={(instance) => {
+              plugin.call('dapp-draft', 'edit', {address: instance.address, abi: instance.contractData.abi, name: instance.name, network: runTab.networkName})
+            }}
           />
         </div>
       </div>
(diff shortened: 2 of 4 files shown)


apps/remix-ide/src/remixAppManager.js
```diff
@@ -390,7 +390,7 @@ class PluginLoader {
         localStorage.setItem('workspace', JSON.stringify(saved))
       },
       get: () => {
-        return JSON.parse(localStorage.getItem('workspace'))
+        return JSON.parse(localStorage.getItem('workspace') || '[]').filter(item => item !== 'dapp-draft')
```
yann300, line 393: could you explain thy this change?

drafish: I think I made a mistake here. I was not sure whether or not the dapp-draft plugin should be loaded in main panel, when I reload remix. 

I just checked other main panel plugins, they were loaded when I reload remix. So I guess it's fine for dapp-draft too.

I will recover this change soon. @yann300

libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -103,12 +103,20 @@ export const setupEvents = (plugin: RunTab, dispatch: React.Dispatch<any>) => {
     if (plugin.name === 'remixd') {
       dispatch(setRemixDActivated(true))
     }
+
+    if (plugin.name === 'dapp-draft') {
```
yann300, line 107: I m not sure why you need to do this.. when the user click on the icon to load dapp-draft, the plugin will be automatically activated.

libs/remix-ui/run-tab/src/lib/run-tab.tsx
```diff
@@ -324,6 +324,11 @@ export function RunTabUI(props: RunTabProps) {
             runTransactions={executeTransactions}
             sendValue={runTab.sendValue}
             getFuncABIInputs={getFuncABIValues}
+            dappdraftActivated={runTab.dappdraftActivated}
+            exEnvironment={runTab.selectExEnv}
+            editInstance={(instance) => {
+              plugin.call('dapp-draft', 'edit', {address: instance.address, abi: instance.contractData.abi, name: instance.name, network: runTab.networkName})
```
yann300, line 330: from the top of my head, I think just this will work.

libs/remix-ui/run-tab/src/lib/actions/events.ts
drafish: I'm not sure. I mean it's not a native plugin for now. Are you ok with the icon always show in deployed contract? I definitely don't need this if you're ok with that. @yann300

yann300: If this requires the plugin to be activated I fear no one will see it. So, let's have this always displayed for now. As long as the feature is free to use, and doesn't imply any subscription, I think we'll be fine for that.
@LianaHus might want to check the UI here though.

libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
```diff
@@ -294,10 +294,11 @@ export function UniversalDappUI(props: UdappProps) {
       </div>
       <div className="udapp_cActionsWrapper" data-id="universalDappUiContractActionWrapper">
         <div className="udapp_contractActionsContainer">
-          <div className="d-flex" data-id="instanceContractBal">
+          <div className="d-flex justify-content-between" data-id="instanceContractBal">
             <label>
               <b><FormattedMessage id="udapp.balance" />:</b> {instanceBalance} ETH
             </label>
+            {props.exEnvironment === 'injected' && <i className="fas fa-edit btn btn-sm p-0" onClick={() => {props.editInstance(props.instance)}}></i>}
```
yann300, line 301: I think this should also be displayed for all providers except VM providers. But what do you think?

drafish:
Can I put dapp-draft in featured plugins? @yann300 

image

libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
drafish: Yeah, I think so. But the dapp template only works for intected provider for now.  More providers will be adapted in further iterations. @yann300

yann300 (approved):

PR #4619, opened 2024-03-11
save & fetch contracts using FE

### LINKED ISSUE #4603: Add pinned contracts in workspace
By @bunsenstraat :
"the point is, on desktop we don't use indexeddb anymore, so that clears the possibility that this storage would be deleted over time or whatever reason."

Aniket-Engg:
fixes #4603


libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
```diff
@@ -17,14 +17,37 @@ export function InstanceContainerUI(props: InstanceContainerProps) {
...
+        // Clear existing saved instance state
+        await props.plugin.call('udapp', 'clearAllSavedInstances')
+        // Load contracts from FE
+        const isPinnedAvailable = await props.plugin.call('fileManager', 'exists', `.deploys/pinned-contracts/${chainId.current}`)
+        if (isPinnedAvailable) {
+          const list = await props.plugin.call('fileManager', 'readdir', `.deploys/pinned-contracts/${chainId.current}`)
+          const filePaths = Object.keys(list)
+          for (const file of filePaths) {
+            const pinnedContract = await props.plugin.call('fileManager', 'readFile', file)
+            const pinnedContractObj = JSON.parse(pinnedContract)
```
bunsenstraat, line 49: Maybe try catch, the json file might have been tampered with

libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
```diff
@@ -17,14 +17,37 @@ export function InstanceContainerUI(props: InstanceContainerProps) {
...
+            if (pinnedContractObj) await props.plugin.call('udapp', 'addSavedInstance', pinnedContractObj.address, pinnedContractObj.abi, pinnedContractObj.name, pinnedContractObj.pinnedAt, pinnedContractObj.filePath)
```
bunsenstraat, line 50: it would be nice if we could move this away from a react hook, basically you're looking at the API state props.plugin.REACT_API.selectExEnv, props.plugin.REACT_API.networkName in a hook but not doing anything to related to the UI except the ref enableSave which could be a state var and not a ref. it feels like this logic of saving and migrating data does not belong in a hook here. enableSave could be in a reducer if there is one or something...

libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
Aniket-Engg: done

bunsenstraat (approved):

libs/remix-ui/run-tab/src/lib/components/instanceContainerUI.tsx
Aniket-Engg: will do the refactoring in another PR

PR #4621, opened 2024-03-12
v0.45.0 release highlights

Aniket-Engg:


yann300 (approved):

vermouth22 (commented):
On Line 4 I would say: 
Remix VM state can now be saved

On Line 6 I would say: 
Saved contracts are now stored in workspaces as "Pinned Contracts"

PR #4628, opened 2024-03-14
feat: add import from swarm

Cafe137:


Aniket-Engg:
@ryestew Can you check this for the feasibility to add import from Swarm?

ryestew:
@Cafe137 - Could you join the Remix discord server or email remix@ethereum.org.  We'd like to have a call about this PR.  Thanks!

PR #4631, opened 2024-03-15
fix version

bunsenstraat:
1. we define as requested the default version in package.json 
3. during build we load that version as default soljson in assets 
4. 3. we use the package version to replace the static strings that were used


apps/remix-ide/webpack.config.js
```diff
@@ -16,9 +16,10 @@ const versionData = {
 
 const loadLocalSolJson = async () => {
   //execute apps/remix-ide/ci/downloadsoljson.sh
+  console.log('loading local soljson')
```
Aniket-Engg, line 19: Remove log

libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
```diff
@@ -6,10 +6,12 @@ import { Toaster } from '@remix-ui/toaster' // eslint-disable-line
 import { ModalDialog } from '@remix-ui/modal-dialog' // eslint-disable-line
 import { Renderer } from '@remix-ui/renderer' // eslint-disable-line
 import { baseURLBin, baseURLWasm, pathToURL } from '@remix-project/remix-solidity'
-
+import * as packageJson from '../../../../../package.json'
 import './css/style.css'
 import { iSolJsonBinData, iSolJsonBinDataBuild } from '@remix-project/remix-lib'
 
+const defaultPath = 'compiler_config.json'
```
Aniket-Engg, line 13: Where is this being used?

bunsenstraat: Didn't know I put that in here...

yann300 (approved):

Issue #4537, opened 2024-02-14
Vyper checklist


ryestew:
- [x] https://github.com/ethereum/remix-project/issues/4536
- [x] Make an accordion called Advanced configurations (just like solidity) move the radio buttons for remote & local compiler in there.

- [x] Fix Remote compiler Radio button label so it is clickable and make sure both have a cursor pointer (the finger icon)
- [x] Make sure that compiler version & EVM version are in the compilation details
- [x] Remove the Vyper word with the new logo at the top of the plugin
- [x] Add the link to the docs on the text "Specify the compiler version & EVM version in the .vy file" and put a period at the end of the text. Add the link to https://docs.vyperlang.org/en/stable/compiling-a-contract.html#setting-the-target-evm-version on `EVM` and the link (if they update the vyper docs) to the appropriate link for the compiler version on the text `compiler version`. Or these links could go to our Vyper compiler docs.
- [x] Add the Ask GPT button to the compiler error
- [x] Make play button work for compilation work
- [x] ctrl + s for compilation on a .vy file 
- [x]  replace vyper-lang for cloning with this repo: https://github.com/pcaversaccio/snekmate 
- [x] write vyper compiler docs

LianaHus:
- [ ] error number on the badge behaves differently from solidity. (number, update etc)
- [ ] error item's width should be a bit less
- [ ] "Specify the [compiler version](https://remix-ide.readthedocs.io/en/latest/vyper.html#specify-vyper-version) & [EVM version](http://docs.vyperlang.org/en/stable/compiling-a-contract.html#setting-the-target-evm-version) in the .vy file." makes no sense to user. @TODO discuss
- [ ] too much space around that warning ^^
- [ ] Documentation link takes to the plugins list where the documentation for the vyper is outdated @ryestew

LianaHus:
- [ ] Compilation details UI improvements

LianaHus:
- [ ] spinner for compile button like in Sol plugin

joeizang:
> * [ ]  spinner for compile button like in Sol plugin

Good idea

PR #4645, opened 2024-03-18
Vyper improvement patches.

joeizang:
address #4537 issues including:

- spinner for compile button like in Sol plugin
- Fix badge behaviour to match solidity compiler

### DIFF
--- a/apps/remix-ide-e2e/src/tests/sol2uml.test.ts
+++ b/apps/remix-ide-e2e/src/tests/sol2uml.test.ts
@@ -4,30 +4,30 @@ import init from '../helpers/init'
 
 module.exports = {
   '@disabled': true,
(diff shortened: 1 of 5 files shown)


ryestew (commented):
The badge update is for the icon panel right?  But I don't see any change there - although it does look fine. The spinner change is fine.

ryestew:
OK - this PR works - but it is leaving out 1 thing:
When switching between vyper files - if 1 has a compiler error and you switch to another file, the error badge on the icons panel doesn't clear - so the uncompiled file shows the error. A successful compilation will clear this error - but it should be removed when another file has the focus.

joeizang:
> or. A successful compilation will clear this error - but it should be removed when another file has the focus.

Ok

ryestew (approved):

PR #4646, opened 2024-03-19
enable contract pinning for VM

Aniket-Engg:
Allow to pin contract for VM environments


LianaHus:
pinned btn is still green instead of filled

LianaHus:
you need a bit more space after "Currently you have no pinned contracts to interact with.", also once you pin and unpit a contract you don't see that msg anymore. The other issue is that one and the one for unpinned have different styles. can't we make the similar? for example text-secondary or text-info.

yann300:
can we add (if it's not arleady there) a matomo check when one interact with a pinned instance.

LianaHus:
data-id deployAndRunClearInstances  p-2
data-id NoPinnedInstanceText  m-2
no item elements: alert alert-secondary

Aniket-Engg:
> secondary

done

Aniket-Engg:
> you need a bit more space after "Currently you have no pinned contracts to interact with.", also once you pin and unpit a contract you don't see that msg anymore. The other issue is that one and the one for unpinned have different styles. can't we make the similar? for example text-secondary or text-info.

done

Aniket-Engg:
> pinned btn is still green instead of filled

done

Aniket-Engg:
> can we add (if it's not arleady there) a matomo check when one interact with a pinned instance.

done

LianaHus:
with the way you log in matomo we can't filter by network. I would like to know why @yann300 want to add tracking and what exactly should be tracked.

Aniket-Engg:
> with the way you log in matomo we can't filter by network. I would like to know why @yann300 want to add tracking and what exactly should be tracked.

If purpose of adding matomo is to ensure that user interact with pinned contract, that is solved by matomo added in this PR.

There is already a matomo tracking which checks on which network contract is pinned

Issue #4611, opened 2024-03-10
Fix error while parsing witness input in circom


ioedeveloper:
When manually typing an array like value as a witness input, circom plugin crashes. It doesn't crash if the array value is pasted in the field.

PR #4648, opened 2024-03-19
Fix parsing bug when typing an array value as witness input.

ioedeveloper:
Fixes https://github.com/ethereum/remix-project/issues/4611


yann300 (approved):

apps/remix-ide/src/assets/list.json
```diff
@@ -1091,9 +1091,22 @@
         "bzzr://c604bdd6384bf73594cd0e5cfbe979048191549ebc88e70996346f3b744c0680",
         "dweb:/ipfs/QmW2SQbEhiz3n2qV5iL8WBgzapv6cXjkLStvTMpCZhvr2x"
       ]
+    },
```
yann300, line 1094: This is a desktop artifacts that should be removed from this PR.

PR #4656, opened 2024-03-22
Update downloadsoljson.sh

bunsenstraat:
do not use JQ in this script but grep and awk..
everyone has awk right?


Aniket-Engg (approved):
local build working for me without any extra installation

PR #4657, opened 2024-03-22
Learneth patch1

drafish:
1. remove redux and redux-saga
2. remove custom css classnames, but still remain some classnames that can't be removed
3. i18n for learneth
4. auto switch repo based on current language
5. active solidityUnitTest automatically

@ryestew @yann300


apps/learneth/src/actions/index.ts
```diff
@@ -0,0 +1,371 @@
...
+import axios from 'axios'
+import groupBy from 'lodash/groupBy'
+import pick from 'lodash/pick'
+import remixClient from '../remix-client'
+import {router} from '../App'
+
+let dispatch, state
+
+// const apiUrl = 'http://localhost:3001';
+const apiUrl = 'https://static.220.14.12.49.clients.your-server.de:3000'
```
yann300, line 11: whose address is that?

drafish: It's the backend service address that I got from the old learneth plugin. @yann300 

https://github.com/bunsenstraat/remix-learneth-plugin/blob/master/src/environments/environment.prod.ts#L4

bunsenstraat: we will change that now that the plugin is internal. I will update it later.

yann300: could this PR be merged in the meantime?

PR #4689, opened 2024-04-07
Tooltip internalization

STetsing:
- changed the AI documentation link
- removed all solcoder terminal logs
- edited tooltip languages for internalization


ryestew (changes requested):
Internationalization should not be updated in the language folder - I mean you only should do the english one and not for example es/remixUiTabs.json

So it was probably done correctly originally. 
please remove the updates for folders: es, fr, it, zh

PR #4705, opened 2024-04-09
Use gasLimit

yann300:
Currently, the gasLimit property that can be set from the run and deploy interface is no use.
This change allows to manually set the gasLimit.
It also provides a checkbox for using or no the gas estimate. If the checkbox is activated, the value from gas estimate is used, if not, the provided value is used.


Aniket-Engg (commented):
Functionality seems working fine but UI is broken
Screenshot 2024-04-17 at 4 36 31 PM

yann300:
> Functionality seems working fine but UI is broken Screenshot 2024-04-17 at 4 36 31 PM

yes this needs @LianaHus checks

Aniket-Engg (approved):

PR #4713, opened 2024-04-12
refactoring contract pinning

Aniket-Engg:


libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -94,22 +96,7 @@ export const setupEvents = (plugin: RunTab, dispatch: React.Dispatch<any>) => {
...
-        for (const file of filePaths) {
-          const pinnedContract = await plugin.call('fileManager', 'readFile', file)
-          const pinnedContractObj = JSON.parse(pinnedContract)
-          if (pinnedContractObj) addPinnedInstance(dispatch, pinnedContractObj)
-        }
-      } catch(err) {
-        console.log(err)
-      }
-    }
+    await migrateSavedContracts(plugin)
```
yann300, line 99: on the `setWorkspace` you don't load pinned contracts?

libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -179,6 +166,50 @@ export const setupEvents = (plugin: RunTab, dispatch: React.Dispatch<any>) => {
   }, 30000)  
 }
 
+const loadPinnedContracts = async (plugin, dispatch, dirName) => {
+  console.log('loadPinnedContracts---->')
```
yann300, line 170: should probably be removed.

libs/remix-ui/run-tab/src/lib/actions/events.ts
Aniket-Engg: yes, we load it once network is ready

libs/remix-ui/run-tab/src/lib/actions/events.ts
Aniket-Engg: removed

libs/remix-ui/run-tab/src/lib/actions/events.ts
yann300: yes but when the workspace change you still need to reload pinned contract no?

Aniket-Engg: when workspace is changed, network is also reset/updated

yann300 (approved):

PR #4723, opened 2024-04-15
Update networks.ts

coolestowl:
etherscan api replace goerli with holesky

### DIFF
--- a/apps/etherscan/src/app/utils/networks.ts
+++ b/apps/etherscan/src/app/utils/networks.ts
@@ -19,7 +19,7 @@ export const scanAPIurls = {
   8453: 'https://api.basescan.org/api',
 
   // all testnet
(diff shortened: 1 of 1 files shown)


yann300 (approved):

coolestowl:
I'm a little confused. In the setupMetamask step in the test, after typing the words when importing the wallet, it waits to click the confirm button until it times out. Guessing there is some error that is causing the button not to be activated, what should I do to make this pull request pass the test?

coolestowl:
sorry to interrupt, but could you please take a look at this? I tried several times and the same test always fails. @yann300

yann300:
issue with this test is that the CI doesn't seem to push the metamask env correctly. but other tests are fine.

coolestowl:
well, now that we know why, what can I do next to move forward with this pr ?

PR #4728, opened 2024-04-16
test sol-gpt chat. Inference with context

STetsing:


Aniket-Engg:
I was never meant to be merged. Experimental PR. cc: @STetsing

PR #4735, opened 2024-04-17
Integrate solidity scan in udapp

Aniket-Engg:


yann300:
Moda UX isn't really working ok, I don't want to interfer with any UX decision, but it seems to me that logging the result to the terminal is just fine.

Aniket-Engg:
> Modal UX isn't really working ok, I don't want to interfer with any UX decision, but it seems to me that logging the result to the terminal is just fine. Also you won't have the issue of having to design a modal..

Terminal will also not be better to show this data IMO. May be we can adjust the modal size.

LianaHus:
- For the user who doesn't know what solidity scan is, the tooltip says nothing. Can you expand there?
- The modal for permission makes me anxious because I still don't know what solidity scan is, and I should share my code now. I would 99% not do it with this data. To improve this we can add more info in the modal and include the link to the project.
- columns are too narrow. I assume this is because you are using our default modal and it has some restrictions. you can either create a custom one or alter the existing one to have 2-3 more views 
- there are some <p> tags in your rightmost column
- close is not closing the modal for me

Screenshot from 2024-06-04 08-24-41

Aniket-Engg:
> * For the user who doesn't know what solidity scan is, the tooltip says nothing. Can you expand there?

Updated! Please check
> * The modal for permission makes me anxious because I still don't know what solidity scan is, and I should share my code now. I would 99% not do it with this data. To improve this we can add more info in the modal and include the link to the project.

Updated! Please check

> * there are some  tags in your rightmost column

They are coming directly from Solidity scan. 
> * close is not closing the modal for me

It is working fine for me

LianaHus (approved):

PR #4742, opened 2024-04-18
fix pinning for injected provider

Aniket-Engg:


libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -14,6 +14,10 @@ import { shortenAddress } from "@remix-ui/helper"
 const _paq = window._paq = window._paq || []
 
 export const setupEvents = (plugin: RunTab, dispatch: React.Dispatch<any>) => {
+  const currentNetwork = {
```
yann300, line 17: could you add a comment in a code on why you need to do this?

libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -33,7 +37,6 @@ export const setupEvents = (plugin: RunTab, dispatch: React.Dispatch<any>) => {
     }
     setFinalContext(plugin, dispatch)
     fillAccountsList(plugin, dispatch)
-    await loadPinnedContracts(plugin, dispatch)
```
yann300, line 36: what happens if one change the network in metamask (or any injected provider) ?

yann300: > what happens if one change the network in metamask (or any injected provider) ?

I mean that the network get changed not by selecting a new env in udapp, but by selecting another network directly from metamask.

Aniket-Engg: both the cases are handled by this condition `if (currentNetwork.provider !== networkProvider() || (!isVM && currentNetwork.chainId !== network.id))` you can try it on netlify link

libs/remix-ui/run-tab/src/lib/actions/events.ts
Aniket-Engg: done

yann300 (approved):

PR #4747, opened 2024-04-19
Vyper-patch-007

joeizang:
Update e2e tests to account for new vyper repo (snekmate).

### DIFF
--- a/apps/remix-ide-e2e/src/commands/addFileSnekmate.ts
+++ b/apps/remix-ide-e2e/src/commands/addFileSnekmate.ts
@@ -1,0 +1,74 @@
+import EventEmitter from 'events'
+import { NightwatchBrowser, NightwatchContractContent } from 'nightwatch'
+
+class AddFileSnekmate extends EventEmitter {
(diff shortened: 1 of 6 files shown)


apps/vyper/src/app/app.tsx
```diff
@@ -119,7 +119,7 @@ const App = () => {
         <div className="px-3 pt-3 mb-3 w-100">
           <CustomTooltip placement="bottom" tooltipText="Clone Vyper examples. Switch to the File Explorer to see the examples.">
             <Button data-id="add-repository" className="w-100 btn btn-secondary" onClick={() => remixClient.cloneVyperRepo()}>
-              Clone Vyper examples repository
```
ryestew, line 122: I think we should generalize this to:
Clone a repo of Vyper examples
- I think that won't be too long.

ryestew (changes requested):

apps/vyper/src/app/app.tsx
joeizang: Done

ryestew (approved):

PR #4752, opened 2024-04-22
Multiselect test

joeizang:
Create e2e tests for multiselection of files in file explorer.

### DIFF
--- a/apps/remix-ide-e2e/src/tests/file_explorer_multiselect.test.ts
+++ b/apps/remix-ide-e2e/src/tests/file_explorer_multiselect.test.ts
@@ -10,14 +10,12 @@ module.exports = {
     const selectedElements = []
     browser
       .openFile('contracts')
-    browser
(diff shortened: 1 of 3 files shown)


apps/remix-ide-e2e/src/tests/file_explorer_multiselect.test.ts
```diff
@@ -0,0 +1,29 @@
...
+module.exports = {
+  before: function (browser: NightwatchBrowser, done: VoidFunction) {
+    init(browser, done)
+  },
+
+  'Should select multiple items in file explorer #group1': function (browser: NightwatchBrowser) {
+    const selectedElements = []
+    browser
+      .openFile('contracts')
+    browser
```
Aniket-Engg, line 13: I think you don't need to write `browser.` on each line

joeizang: You are right sort of. I still need an extra browser.

Aniket-Engg (approved):

PR #4754, opened 2024-04-22
Update to latest font-awesome kit

ryestew:
We also need to update the apps/remix-ide/src/index.html and all the other index.html files with the correct import css files and also all the apps index.html of the css files. For example apps/circuit-compiler/src/index.html

The apps like the circuit-compiler/index.html import this: 
<link rel="stylesheet" integrity="ha384-50oBUHEmvpQ+1lW4y57PTFmhCaXp0ML5d60M1M7uH2+nqUivzIebhndOJK28anvf"
		crossorigin="anonymous" href="https://use.fontawesome.com/releases/v5.8.1/css/all.css">

The kit is downloaded as the **latest version**.   The latest version currently is 6.5.2.  It looks like the apps need to be pegged to a specific version. 

I believe that the desktop needs the file all.css but this is not optimized and is probably too big for the online IDE:  see
https://docs.fontawesome.com/web/setup/host-yourself/webfonts
" And if you are using all.js or all.css, your custom icons will be included in those, though for performance reasons, we recommend adding just the files you are using in your project."

And on a smaller note, the kit has a file in the css folder called **fontawesome.min.css** and in our assets/css folder there is a file called: **font-awesome.min.css**. So it looks like a little cleanup of the necessary css files is needed.

- [ ] So I'd like @LianaHus, @joeizang, @bunsenstraat to weigh in to clear up the issue of all.css which only needed for the desktop app
- [ ] Someone else to take over this PR - or to work with me to get the correct links for index.html files.
- [ ] Someone else to take over this PR - or to work with me to get the minimum files needed from the downloaded kit - that is if we don't want to use all.css


joeizang:
@ryestew I think pr can be closed as fontawesome has been updated by yourself.

PR #4756, opened 2024-04-23
Dapp draft patch1

drafish:
Add two more parameters to edit method, so that dapp-draft could auto input the instructions from devdoc.

image

And fix one small issue by the way -- the edit icon wouldn't show up because the exEvironment value has been changed.

@ryestew @yann300


libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
```diff
@@ -297,7 +297,7 @@ export function UniversalDappUI(props: UdappProps) {
             <label>
               <b><FormattedMessage id="udapp.balance" />:</b> {instanceBalance} ETH
             </label>
-            {props.exEnvironment === 'injected' && <i className="fas fa-edit btn btn-sm p-0" onClick={() => {props.editInstance(props.instance)}}></i>}
+            {props.exEnvironment.indexOf('MetaMask') > -1 && <i className="fas fa-edit btn btn-sm p-0" onClick={() => {props.editInstance(props.instance)}}></i>}
```
yann300, line 300: why do you restrict it to metamask?

yann300: oho ok i see... but we have to find another solution...

yann300: can you rebase and remove that changes from this PR?

yann300:
merging as the metamask test always fails for external PR, and this isn't linked to that PR.

PR #4767, opened 2024-04-25
Add Multiselect dragdrop Feature

joeizang:


yann300:
After releasing in a drop zone, the modal doesn't sometimes show the right files name. 
Also when releasing it tries take in account the actual file that the mouse is hovering.

joeizang:
in the file_explorer_multiselect.test.ts file, all tests pass on google chrome but one test, Should drag and drop multiple files in file explorer to tests folder #group1, fails on firefox. Can you look at this test @yann300 ?

libs/remix-ui/workspace/src/lib/css/file-explorer.css
```diff
@@ -69,3 +69,7 @@ ul {
 .remixui_icons:hover {
     color: var(--text);
 }
+
+.remixui_selected {
```
ioedeveloper, line 73: Remove unused css class

joeizang: right. done.

libs/remix-ui/workspace/src/lib/components/flat-tree.tsx
```diff
@@ -116,6 +124,12 @@ export const FlatTree = (props: FlatTreeProps) => {
 
   const onDragEnd = (event: SyntheticEvent) => {
     setIsDragging(false)
+    document.querySelectorAll('li.remixui_selected').forEach(item => {
```
ioedeveloper, line 127: Should be improved to use react hooks (useRef, useState, etc).

ioedeveloper (approved):

libs/remix-ui/workspace/src/lib/utils/getEventTarget.ts
```diff
@@ -22,4 +22,35 @@ export const getEventTarget = async (e: any, useLabel: boolean = false) => {
...
+
+/**
+   *  When multiple files are selected in FileExplorer,
+   *  and these files are dragged to a target folder,
+   *  this function will build the profile of each selected item
+   *  in FileExplorer so they can be moved when dropped
+   * @param target - Initial target item in FileExplorer
+   * @returns - {DragStructure} Array of selected items
+   */
+export const buildMultiSelectedItemProfiles = (target: {
```
ioedeveloper, line 35: Same with this function

PR #4797, opened 2024-05-15
simple chat in terminal

STetsing:
`sol-gpt` now remembers previous output, so users can refine their question thereby getting accurate answers


apps/remix-ide/src/app/plugins/solcoderAI.tsx
```diff
@@ -62,24 +80,30 @@ export class SolCoder extends Plugin {
     this.call('layout', 'maximizeTerminal')
     let result
     try {
+      const main_prompt = this._build_solgpt_promt(prompt)
+      console.log(main_prompt.length)
```
Aniket-Engg, line 84: do you want to keep this log?

PR #4798, opened 2024-05-15
Pinned panel

ioedeveloper:
Closes https://github.com/ethereum/remix-project/issues/3787
### How it works
This PR introduces the functionality of being able to pin/dock a plugin to a fixed position on the right of the IDE.
- Each plugin (except the filePanel) have a new icon on the header as indicated in the image below.
Screenshot 2024-05-15 at 15 07 49

NOTE: We would need to discuss the plugins that should have this new icon to pin and unpin plugins.
- After clicking on the "pin" icon, the plugin should be moved to the right of the IDE.
Screenshot 2024-05-15 at 15 14 17

NOTE: For now, only internal plugins have the ability to maintain state when pinned from left to right and vice-versa. Iframe plugins will have to reload when it switches sides. Also, i have only implemented state persistence for udapp and search plugin, because they are the easiest to implement for now because of how state is managed in this two plugins. Implementing for other internal plugins will involve a lot of code changes, which is out of the scope of this PR.
- When an icon is pinned, it stays pinned until you unpin the plugin or deactivate the plugin. Even after reload pinned plugins are still pinned.
- I will add more when i have more to disclose. Have fun testing...


apps/remix-ide/src/app/tabs/locales/en/panel.json
```diff
@@ -4,6 +4,8 @@
   "panel.documentation": "Documentation",
   "panel.description": "Description",
   "panel.maintainedByRemix": "Maintained by Remix",
+  "panel.pinnedMsg": "Click to dock plugin to the right",
+  "panel.unPinnedMsg": "Click to return plugin to the side panel",
```
Aniket-Engg, line 8: may be better to say: `Click to return plugin to the left side panel`

Aniket-Engg (commented):
If a plugin is pinned to right and you try to pin another one, it breaks everything

ryestew (changes requested):
- [ ] When you pin a panel - it pops to the right - but the main panel and the terminal get a ugly scroll bar that is not aligned with each other.
https://github.com/ethereum/remix-project/pull/4798

apps/remix-ide/src/app/tabs/locales/en/panel.json
LianaHus: Agree. Maybe even same sentence but with right/left

apps/remix-ide/src/app/tabs/locales/zh/panel.json
```diff
@@ -6,5 +6,7 @@
   "panel.maintainedByRemix": "由 Remix 维护",
   "panel.pluginInfo": "插件信息",
   "panel.linkToDoc": "文档链接",
-  "panel.makeAnissue": "提交 issue"
+  "panel.makeAnissue": "提交 issue",
```
LianaHus, line 9: You don't need to add those in all language files. It will take English by default anyway. Just add the label "needs translation".

libs/remix-ui/app/src/lib/remix-app/style/remix-app.css
```diff
@@ -34,6 +34,12 @@ pre {
   width              : 320px;
   transition         : width 0.25s;
 }
+.pinnedpanel           {
+  display            : flex;
+  flex-direction     : row-reverse;
```
LianaHus, line 39: Please  use bootstrap classes for first two

LianaHus (changes requested):

LianaHus:
@ioedeveloper please exclude the plugins you've added from the plugin manager

joeizang (approved):
I think it LGTM. In the spirit of small prs and the development process being more iterative, I would approve this pr and have corrections and patches come in later "smaller" prs

PR #4802, opened 2024-05-21
bring back rocket icon for bridges

Aniket-Engg:


ryestew (commented):
Just note that the rocket only appears when L2 - optimism or arbitrum is selected.  But if you go to the Injected Provider and choose one of those chains, the rocket doesn't appear.  This could be a feature in the future.

ryestew (approved):

joeizang (approved):

PR #4814, opened 2024-05-23
Fix Contract Flattener Recursion bug

joeizang:
A user in the community was not able to flatten a contract with a large number of imports. The problem was a bug in the recursion utilized where the flattened contract would not compile because it would come in the wrong order. This has now been fixed with this patch.


Aniket-Engg (commented):
you are removing all pragmas . may be keep the pragma in original parent file

PR #4826, opened 2024-05-31
Learnethupdate

bunsenstraat:
- fixes loading data & files issues, now it loads dynamically much faster
- fixes intl issues
- fix remixplugin client init issues
- added a simple basic e2e test


ryestew (changes requested):
When loading from the Solidity Intro link on the home tab, I get this error:
image

ryestew:
Loading a tutorial in LearnEth does work - but loading from the buttons on the home tab does not work.

ryestew:
How does this PR fix the internationalization issues?  Either in this PR or a future one, LearnETh should match the language choice of Remix.

Issue #4824, opened 2024-05-30
side-panel improvements


LianaHus:
If we move the side panel to the right, the vertical icon should be highlighted somehow because it would be 'focused' always but will not look focused ( no blue border, no bright color)
My suggestion would be

- To make it look disabled (dark color)
- right click would still work
- improved tooltip
- left click will highlight for a second the title of side panel in right part as we currently do with verticaliconpanels icons on plugin activation

ioedeveloper:
We also agreed to have the blue left border appear on the right of the vertical icons panel for the plugins that are pinned to the right.

ioedeveloper:
>     * improved tooltip
> 
>     * left click will highlight for a second the title of side panel in right part as we currently do with verticaliconpanels icons on plugin activation

@LianaHus What should be written as an improved tooltip?

PR #4829, opened 2024-05-31
Pinned panel updates

ioedeveloper:
Closes https://github.com/ethereum/remix-project/issues/4824


ryestew:
The cursor should switch to a finger when over the icon to move it to the right or left side.  In a future version there should be a mouseover state of the icon - maybe in bold.  But the tooltip makes it clear that there will be an action happening.  So I will just leave this as a comment for this PR and not a required change.

ryestew (approved):
This has my approval - but see the note and someone else should approve this as well.

LianaHus (approved):

PR #4844, opened 2024-06-05
add quick-dapp plugin

drafish:
Try it here -- https://remix-quick-dapp.vercel.app/

Here is the explanation for the references in code.

https://remix-dapp.pages.dev is the domain of the dapp template. 

https://remix-dapp.pages.dev/manifest.json is the json file contains all the static files paths of dapp-template. It's generated through the build process automatically. 

Before the dapp deployment, quick-dapp plugin will get all the static files paths of dapp-template, and download them from the dapp-template domain.

https://vercel-proxy-bice-six.vercel.app is the proxy to resolve the cross-domain issue. Because the backend service of surge does not support cross-domain.

Here is the codebase of the dapp-template https://github.com/drafish/remix-dapp
Here is the codebase of the proxy https://github.com/drafish/vercel-proxy

They are not in the remix codebase.

@ryestew


yann300:
@drafish how would you recommend to add e2e test for the plugin and the site template?

apps/quick-dapp/src/components/MultipleContainers/components/Item/Item.tsx
```diff
@@ -0,0 +1,120 @@
+import React, { useEffect } from 'react';
+import type { DraggableSyntheticListeners } from '@dnd-kit/core';
+import type { Transform } from '@dnd-kit/utilities';
+import { Handle } from './components/Handle';
+
+const removeIcon = (
+  <svg
```
LianaHus, line 7: Please use 1 font awsome, 2. webp 3. png instead

apps/quick-dapp/src/components/MultipleContainers/components/Item/components/Handle/Handle.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const Handle = forwardRef<HTMLButtonElement, ActionProps>(
+  (props, ref) => {
+    return (
+      <Action
+        ref={ref}
+        cursor="grab"
+        data-cypress="draggable-handle"
+        {...props}
+      >
+        <svg viewBox="0 0 20 20" width="12">
```
LianaHus, line 14: no svg please. see above

apps/quick-dapp/src/components/MultipleContainers/components/Item/components/Remove/Remove.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function Remove(props: ActionProps) {
+  return (
+    <Action
+      {...props}
+      active={{
+        fill: 'rgba(255, 70, 70, 0.95)',
+        background: 'rgba(255, 70, 70, 0.1)',
+      }}
+    >
+      <svg width="8" viewBox="0 0 22 22" xmlns="http://www.w3.org/2000/svg">
```
LianaHus, line 14: no svg

apps/quick-dapp/src/remix-client.ts
```diff
@@ -0,0 +1,32 @@
...
+import { initInstance } from './actions';
+
+class RemixClient extends PluginClient {
+  constructor() {
+    super();
+    createClient(this);
+  }
+
+  edit({ address, abi, network, name, devdoc, methodIdentifiers }: any): void {
+    // console.log(
```
LianaHus, line 12: please rm

apps/quick-dapp/src/App.css
```diff
@@ -0,0 +1,146 @@
+/* You can add global styles to this file, and also import other style files */
```
LianaHus, line 1: Please use as little manual styling as possible.
- Bootstrap classes should be prioritized over custom styles
- Flexbox, instead of manual calculations, that allows to adjust dynamically based on the available space,

drafish:
> @drafish how would you recommend to add e2e test for the plugin and the site template?

For the plugin

1. Click the edit icon in the deployed contract, and see if the plugin is activated successfully. 
2. Then do the drag and drop thing, and delete some functions.
3. Then input some titles and instructions, and test the natspec by the way.
4. Select all the rest options, such as `share to`, `verified by etherscan`, `no terminal`, `themes`
5. Input email, password, subdomain, and click deploy. May need retry here, because the deployment may fail sometimes.

For the deployed dapp

7. open the dapp domain in a new browser tab, see if it's loaded successfully.
8. check if the order of the functions is correct, and the deleted functions are not shown.
9. check if the titles and instructions are correct
10. check `share to`, `verified by etherscan`, `no terminal`, `themes`
11. send some transactions. (I'm not sure it's testable. Is it possible to preset metamask in Selenium?)

**By the end of the e2e test**, we should delete the subdomain which has be deployed in the e2e test. 

For the dapp template, same as the e2e test for the deployed dapp.
1. open the dapp-template domain, see if it's loaded successfully.
2. check if the order of the functions is correct, and the deleted functions are not shown.
3. check if the titles and instructions are correct
4. check `share to`, `verified by etherscan`, `no terminal`, `themes`
5. send some transactions. (I'm not sure it's testable. Is it possible to preset metamask in Selenium?)

@yann300

yann300:
@LianaHus could you review the UI once more?

drafish:
I think there is a better way to do e2e test for the dapp-template. 

For now, the dapp-template is in an individual repository. I have to do the e2e test in that repository. 

But what if I put it into to remix codebase, just like any other iframe plugins under folder `apps` . So that quick-dapp plugin can load the dapp-template from the ide domain during the dapp deployment. Then I don't need to do e2e test specifically for the dapp-template, when I'v already do the e2e test for the deployed dapp on Surge.

And there is another advantage in this way. It will be easy to keep versions in sync between quick-dapp and dapp-template, because they are in the same repository.

The only question is, can I put dapp-template into remix codebase, even it's not a remix plugin? @yann300 @ryestew

yann300:
> the only question is, can I put dapp-template into remix codebase, even it's not a remix plugin

yes I think we could do that, but first let's merge this PR, sorry for the delay...
Note that we also planned to fork dapp-template to the git org remix-project-org, in the meantime.

drafish:
> > the only question is, can I put dapp-template into remix codebase, even it's not a remix plugin
> 
> yes I think we could do that, but first let's merge this PR, sorry for the delay... Note that we also planned to fork dapp-template to the git org remix-project-org, in the meantime.

Hi @yann300 I'v putted dapp-template into remix codebase. Should I submit it in this PR? Or wait until you merge this one, and make another PR?

apps/quick-dapp/src/components/LoadingScreen/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+const LoadingScreen: React.FC = () => {
+  const { appState } = useContext(AppContext);
+  const loading = appState.loading.screen;
+
+  return loading ? (
+    <div
+      className="w-100 h-100 position-fixed opacity-100 z-3"
+      style={{
+        top: 0,
+        backgroundColor: 'rgba(51, 51, 51, 0.8)',
```
LianaHus, line 14: please use bootstrap vars or classes

apps/quick-dapp/src/components/MultipleContainers/components/Item/components/Remove/Remove.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { Action, ActionProps } from '../Action';
+
+export function Remove(props: ActionProps) {
+  return (
+    <Action
+      {...props}
+      active={{
+        fill: 'rgba(255, 70, 70, 0.95)',
+        background: 'rgba(255, 70, 70, 0.1)',
```
LianaHus, line 11: please use bootstrap vars or classes

yann300 (approved):

PR #4868, opened 2024-06-06
Update tooltips and toaster for solidity and udapp

ioedeveloper:


Aniket-Engg (approved):

LianaHus:
would it be possible to update tooltip for optimisation run number? i just noticed it is still old stule.

LianaHus (approved):

PR #4884, opened 2024-06-13
add eth_signTypedData to remix-simulator

yann300:
This adds `eth_signTypedData` and `eth_signTypedData_v4` endpoint to remix-simulator.
both endpoints should return the same.
see https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md and
https://docs.metamask.io/wallet/reference/eth_signtypeddata_v4/


libs/remix-simulator/src/provider.ts
```diff
@@ -125,7 +124,12 @@ export class Provider {
   }
 
   send (payload, callback) {
-    return this.sendAsync(payload,callback)
+    return this.sendAsync(payload, callback)
```
Aniket-Engg, line 127: this should be using `await`

PR #4886, opened 2024-06-13
Compilation Details Patch

### LINKED ISSUE #4187: compilation details improvement story
- [x] When expanding a nested section by clicking on carrot, carrot disappears. Which make it less user friendly to collapse it again.
- [x] Also, the content opened is not aligned to its parent. See related screenshot in the doc at the bottom. All subfolders should be shifted to the right one level
- [x] Change title to tooltip
- [x] Copy should stop propagation
- [x] Remove the question mark and add tooltips instead
- [x] Use same font family and font size for subcategories
- [x] Move compilation details of vyper to show as a tab
- [x] Add contract name to the file title
- [x] add solidity/vyper logo to the tab title
- [x] mouse pointer when hover over expandable row

joeizang:
fixes #4187


Aniket-Engg (commented):
You have check all the items from issue https://github.com/ethereum/remix-project/issues/4187 but I think not everything is implemented in this like `Remove the question mark and add tooltips instead`, `Add contract name to the file title` etc

joeizang:
> You have check all the items from issue #4187 but I think not everything is implemented in this like `Remove the question mark and add tooltips instead`, `Add contract name to the file title` etc

Ok Lemme check again.

joeizang:
> You have check all the items from issue #4187 but I think not everything is implemented in this like `Remove the question mark and add tooltips instead`, `Add contract name to the file title` etc

I was unable to achieve placing the contract name as the name of the tab. TabProxy will need some refactoring, I think. The situation here is when showing the compile details, a file isn't being opened, which tabproxy can handle easily. I would say this can be fixed in another pr for another release. Everything else seems fine.

Aniket-Engg (approved):

Aniket-Engg:
> > You have check all the items from issue #4187 but I think not everything is implemented in this like `Remove the question mark and add tooltips instead`, `Add contract name to the file title` etc
> 
> I was unable to achieve placing the contract name as the name of the tab. TabProxy will need some refactoring, I think. The situation here is when showing the compile details, a file isn't being opened, which tabproxy can handle easily. I would say this can be fixed in another pr for another release. Everything else seems fine.

please create an issue to track it

PR #4887, opened 2024-06-13
add cookbook

yann300:


Aniket-Engg (approved):
Approved to test it by merging on Alpha.

If it breaks anything, should be fixed on priority

PR #4908, opened 2024-06-20
SolidityScan into solidity compiler

Aniket-Engg:
- Option to run classic static analysis and solidity scan now added in Solidity compiler plugin
- Result of Solidity scan appears in Remix terminal


LianaHus:
- The titles you use for btns sounds more like a tooltip to me. can we get rid of "run"?
- we need icons for newly added btns. SA has one, for Sol scan you can ask them maybe

Aniket-Engg:
> The titles you use for Btns sound more like a tooltip to me. can we get rid of "run"?

I disagree with this. That will make it less user friendly. Also other buttons are having `Publish`

Aniket-Engg:
> The first btns promises to run analytics, but it is enabling the plugin only. It should enable with await and run on the current compiled contract

No APIs are exposed in SSA plugin to run analysis like that. For now, I move users to SSA plugin to run analysis there

Aniket-Engg:
> we need icons for newly added bins. SA has one, for Sol scan, you can ask them maybe

Add icon for SSA, asked for solscan one

apps/remix-ide/src/app/tabs/locales/en/solidity.json
```diff
@@ -52,6 +52,19 @@
 
   "solidity._comment_contract-selection.tsx": "libs/remix-ui/solidity-compiler/src/lib/contract-selection.tsx",
   "solidity.publishOn": "Publish on",
+  "solidity.runStaticAnalysis": "Run Static Analysis",
```
yann300, line 55: I 'd propose `Run Remix Analysis`

Aniket-Engg: updated

yann300:
LGTM!

LianaHus:
- you have issues with dark icons in dark theme
- we may consider moving icons to the left as it is everywhere else in the ide
- only the last one has no icon. we could add

1. `<i class="fa-regular fa-memo-circle-info"></i>`
or 
3. `<i class="fa-regular fa-memo-pad"></i>`

LianaHus (approved):

PR #4954, opened 2024-07-03
status bar fixes

LianaHus:
bg-info to fix theme black
Screenshot from 2024-07-03 11-46-11


joeizang (changes requested):
This is what I see
image
and this is where the dragbar is:
image

The statusbar looks nice though, I like the color and how contrasty it looks

LianaHus:
ok thanks will fix

joeizang (approved):
LGTM

PR #4960, opened 2024-07-05
Status bar updates

### LINKED ISSUE #4949: Status bar updates
1. alert goes to the right end and has a warning bg.
2. on loading the popup with alert should be opened
3. add did you know tips to the center

see wireframe:
image

joeizang:
fixes #4949


LianaHus:
All changes look good.
But (sorry) I just discovered we have a UX issue with the copilot label.
It is a bit confusing why it is there. 
- if this is the same as the switch near the robot, maybe it should look the same. and why we should have two. 
And then it should be really disabled 
image

- If this is just a status then "Copilot is disabled" should be the label and not the tooltip and the tooltip should explain how to switch it on

LianaHus:
lightbulb maybe should be success, warning should be reserved for potential problems. info will do but the background is using blue

LianaHus:
I'm approving this but the tooltip should be discussed at ux meeting

LianaHus (approved):

PR #4973, opened 2024-07-08
expand plugin view

yann300:


LianaHus:
I'm afraid this is not the right way to do it.
We shouldn't expand it but rather assign individual widths as for debugger.
This approach leads to 
Screenshot from 2024-07-09 10-38-56

Also two large panels (right and link) don't work. so we either can restrict their move on the right side or find other solution. see below 
Screenshot from 2024-07-09 10-39-23

PR #4978, opened 2024-07-10
show lines for warnings where available, in Solidity Scan report

### LINKED ISSUE #4967: SolidityScan integration bugs
**Describe the bug**
The recent SolidityScan integration is missing a few valuable features. 
1. The gas bugs' line numbers are not shown. This can be highlighted in the code where the bugs are present so the users have a clear indication of where the bugs lie. 
2. The summary of the bugs is not shown. It just shows 6 warnings found. 
3. Add UTM for tracking "https://solidityscan.com/?utm_campaign=remix&utm_source=remix"
4. Add a trailing `/` on the API call.

**To Reproduce**
1. Go to remix and compile any code.
2. Click on run "SolidityScan"
Screenshot 2024-07-07 at 00 30 05
5. We will notice that the gas bugs are not highlighted and hence users will not get to know where the bugs are. 
6. We will also notice the warnings. It would have been more suitable if the exact bug count as per severity is mentioned. 
Screenshot 2024-07-07 at 00 31 12

**Expected behavior**
Since the API provides the option of the location of the bug it is beneficial to show it to the users. 

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser - chrome, firefox

Aniket-Engg:
Fixes #4967 

It shows warning lines for `gas` issues in the report 

Screenshot 2024-07-10 at 5 01 23 PM

and highlights same lines in editor
Screenshot 2024-07-12 at 11 11 22 AM

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/contract-selection.tsx
+++ b/libs/remix-ui/solidity-compiler/src/lib/contract-selection.tsx
@@ -313,8 +313,19 @@ export const ContractSelection = (props: ContractSelectionProps) => {
 
           const { data: scanData } = await axios.post('https://solidityscan.remixproject.org/downloadResult', { url })
           const scanReport: ScanReport = scanData.scan_report
-
           if (scanReport?.multi_file_scan_details?.length) {
+            for (const template of scanReport.multi_file_scan_details) {
+              if (template.metric_wise_aggregated_findings?.length) {
+                const { metric_wise_aggregated_findings } = template
+                let positions = []
+                for (const details of metric_wise_aggregated_findings) {
+                  const { findings } = details
+                  for (const f of findings)
+                    positions.push(`${f.line_nos_start[0]}:${f.line_nos_end[0]}`)
+                }
+                template.positions = JSON.stringify(positions)
+              }
+            }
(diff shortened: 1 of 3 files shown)


libs/remix-ui/solidity-compiler/src/lib/contract-selection.tsx
```diff
@@ -313,8 +313,29 @@ export const ContractSelection = (props: ContractSelectionProps) => {
...
+                  for (const f of findings){
+                    const position = {
+                      start: {
+                        line: f.line_nos_start[0] - 1,
+                      },
+                      end: {
+                        line: f.line_nos_end[0],
+                      }
+                    }
+                    await plugin.call('editor', 'highlight', position)
```
yann300, line 332: at which point does this get unhighlighted?

Aniket-Engg: I have just added a commit to discard highlighting on multiple actions

Aniket-Engg: highlighting is removed

LianaHus (approved):

PR #4979, opened 2024-07-10
Show log in github from FE

yann300:
- adds "Log in Github" to the File Explorer.
- the github user is synced through an app state
- tests are added to connect through the FE and disconnect


LianaHus:
I would suggest redesigning a bit:
Screenshot from 2024-07-11 10-38-48
Screenshot from 2024-07-11 10-38-25

LianaHus:
IMO "Log in to GitHub" is enough we do not need "Click to" in the tooltip

libs/remix-ui/git/src/components/panels/githubcredentials.tsx
```diff
@@ -52,6 +53,7 @@ export const GitHubCredentials = () => {
     setGithubToken(credentials.token || '')
     setGithubUsername(credentials.username || '')
     setGithubEmail(credentials.email || '')
+    props.plugin.emit('loggedInGithubChanged')
```
bunsenstraat, line 56: no passing around plugin in props please.

bunsenstraat: use plugin actions context

yann300:
little reminder, we need to remove the bootstrap CSS color class and put a cursor pointer to it.

Aniket-Engg (commented):
Screenshot 2024-08-27 at 7 07 38 PM


1. It should not say `Login in with`

Screenshot 2024-08-27 at 7 12 40 PM

2. It looks like manually providing Github credentials never works and show error dialog in each case.

bunsenstraat:
@Aniket-Engg @yann300 please chip in on the topic

Some answers and explanations:

> 1. It should not say `Login in with`
what should it say?

>2. It looks like manually providing Github credentials never works and show error dialog in each case.

That is not true in my tests and is also included in e2e, it works when you provide a valid token with the correct permissions. The point is this a lot of instructions to give how to create a valid token. So where do we put it? in a popup? or a panel? where can we explain this?
I would rather dismiss the whole manual procedure all together and _only_ allow login via github... but that would be too radical probably. 👯 

> 3. It is better to tell what permissions are expected with token. I couldn't make this work.
Yes it's technically not available. Github only supplies the scopes of permissions of a token in the API when you log in with github. Not when you provide tokens you create yourself. Nothing I can do about it. 

> 4. For logged in user, labels and details are not properly aligned, even if there are a great space available
It looks the same as in any other part of remix UI with input fields.

>5. Labels are inconsistent saying Git for username and email but GitHub for token
Yes on purpose. the username/email is a git setting, the token is only connected to github. the username/email is something you always need to do commits, and is git only. you can commit and do stuff if you have that filled in. the token is optional.

Aniket-Engg:
> what should it say? Sign in is better?

`Login in with GitHub` can be updated to just `Login with GitHub`

Aniket-Engg (approved):

STetsing:
Would be nice to 
- click on 'log in` and land to login setting
- access current setting when in explorer by clicking on the logged in profile

PR #4995, opened 2024-07-15
reduced completion timer

STetsing:
Reduced the timer amount for copilot suggestion. It might be a the root of causing the completions results to not being displayed


STetsing:
@yann300  can you check the timeout function, so we can discuss about it

yann300 (approved):

PR #5007, opened 2024-07-16
Dgituifix

bunsenstraat:
- the header was a bit broken, things would not line up and not be truncated correctly. I simplified in agreement with Liana
- the detached head warning wasn't displayed because it was on 1 line with the message
- I added required px-2 to the whole plugin
- some buttons where renamed
- I widen the plugin just like debugger to make more space to read
- somehow dgitApi wasn't activated in app.js anymore... don't know how that happened
- needed to add some pauses to the e2e because the animation causes rerendering which causes elements to be not clickable


joeizang (approved):
Changes seem ok

PR #5017, opened 2024-07-17
readd and fix learneth

yann300:
This readd the learneth sectionin the homepage and fix a bug.

To reproduce the bug:
 - from the home page, start a tutorial.
 - from the homepage, select another tutorial.
 - the learneth iframe view is crashing.

The fix is just about not adding a repo if it's already present.


Aniket-Engg:
As per the discussion, this will be done in a different way.

PR #5026, opened 2024-07-19
Track More Metrics for Circom with Matomo

ioedeveloper:
- Updated path for circom template scripts in other to uniquely identify scripts that are ran for circom.
- Updated e2e tests with new path to fix failing test.


libs/remix-ui/workspace/src/lib/actions/workspace.ts
```diff
@@ -208,6 +208,7 @@ export const createWorkspace = async (
     if (workspaceTemplateName === 'semaphore' || workspaceTemplateName === 'hashchecker' || workspaceTemplateName === 'rln') {
       const isCircomActive = await plugin.call('manager', 'isActive', 'circuit-compiler')
       if (!isCircomActive) await plugin.call('manager', 'activatePlugin', 'circuit-compiler')
+      _paq.push(['trackEvent', 'circuit-compiler', 'template', 'create', workspaceTemplateName])
```
Aniket-Engg, line 211: are we generally use 5 params with `_paq.push`

ioedeveloper: No, i think you can use as many

Aniket-Engg (approved):

joeizang (approved):

PR #5051, opened 2024-07-30
add matomo to git

bunsenstraat:


libs/remix-api/src/lib/plugins/matomo-api.ts
```diff
@@ -0,0 +1,10 @@
+import { IFilePanel } from '@remixproject/plugin-api'
+import { StatusEvents } from '@remixproject/plugin-utils'
+
+export interface IMatonmoApi {
```
Aniket-Engg, line 4: Seems there is a typo

libs/remix-api/src/lib/remix-api.ts
```diff
@@ -11,6 +11,7 @@ import { Plugin } from "@remixproject/engine"
 import { ISidePanelApi } from "./plugins/sidePanel-api"
 import { IPinnedPanelApi } from "./plugins/pinned-panel-api"
 import { ILayoutApi } from "./plugins/layout-api"
+import { IMatonmoApi } from "./plugins/matomo-api"
```
Aniket-Engg, line 14: consequently, here too

Aniket-Engg (approved):

PR #5057, opened 2024-08-01
update typescript to latest v5.5.4

Aniket-Engg:


joeizang:
I have set target to ES2022 based on recommendations [here](https://github.com/tsconfig/bases?tab=readme-ov-file#centralized-recommendations-for-tsconfig-bases) and [here](https://www.npmjs.com/package/@tsconfig/node20)

apps/remix-ide/src/app/plugins/remixd-handle.tsx
```diff
@@ -43,7 +43,7 @@ export class RemixdHandle extends WebsocketPlugin {
         await this.appManager.deactivatePlugin(plugin)
       }
     }
-    if (super.socket) super.deactivate()
+    if (this.socket) super.deactivate()
```
Aniket-Engg, line 46: should it be `this.deactivate()`

apps/doc-gen/tsconfig.json
```diff
@@ -2,6 +2,8 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "jsx": "react-jsx",
+    "target": "ES2022",
```
bunsenstraat, line 5: so only this plugin needs 2022? It's extended from base.

apps/remix-ide/src/app/plugins/remixd-handle.tsx
joeizang: true

apps/doc-gen/tsconfig.json
joeizang: We could make the base to target ES2022, but this would have to be agreed upon by us here.

apps/remix-ide/src/app/plugins/remixd-handle.tsx
joeizang: done

ioedeveloper (approved):

Issue #5056, opened 2024-08-01
Please add undecoded output to transaction details


aliveli186:
While there is input & decoded input there is only decoded output on the details page.
It would be nice to have undecoded output as well to be able to see how encoding of some data types is made on return.

![image](https://github.com/user-attachments/assets/614e6ade-ef91-4e7f-8815-9067664fcb46)

Deepakchowdavarapu:
I can work on this issue , please Assign it to me

LianaHus:
> I can work on this issue , please Assign it to me

you are welcome to work on this. please submit a PR.

shubhiscoding:
Attempted a solve at #5072

Aniket-Engg:
fixed by https://github.com/ethereum/remix-project/pull/5072

PR #5072, opened 2024-08-08
Added Output on terminal for undecoded Output

shubhiscoding:
Solved #5056

Like input shows unDecoded input similarly added output for unDecoded Output in terminal

Screenshot from 2024-08-08 23-45-11


shubhiscoding:
@Aniket-Engg would you like  to review it

Aniket-Engg (approved):
change set looks good

PR #5075, opened 2024-08-14
persist providers

yann300:
This PRs:

 - Reduce the list of default providers.
 - Persist the list to the local storage, so selected providers get pinned again upon reload.
 - Add the remaining provider for e2e purposes.


Aniket-Engg (approved):

Hounddogg85 (commented):
Good stuff does it actually provide and hit me up with any kind of work u have I'm D4W

PR #5083, opened 2024-08-16
address copy button added

ijlal321:
### **Feature Added: Copy Address Button in Accounts Dropdown**

**Reason:**  
Previously, to copy an address, users needed to select the address, copy it, and then reselect the previous address. The goal of this feature is to streamline the process by providing a button that allows users to directly copy the address, eliminating the extra steps.

**Implementation Details:**  
Originally this dropdown uses a html native select (which does not support any buttons inside option ). This is now replaced by react-select component. React-select allows a dropdown option to have buttons.

**Code Dependability:**  
This new select is made to look exactly as previous one. Most classes and components used are same as previous select. The copy button is imported from remix-copy. All fonts are working perfectly.


Aniket-Engg:
@ijlal321 linting is failing for the PR

ijlal321:
> @ijlal321 linting is failing for the PR

Got it, fixed it 👍

libs/remix-ui/run-tab/src/lib/components/account.tsx
```diff
@@ -196,6 +197,77 @@ export function AccountUI(props: AccountProps) {
     )
   }
 
+  const customStyles = {
```
LianaHus, line 200: please do not use custom styles

LianaHus:
please check out how the custom dropdown is implemented in **Environment** dropdown and do not use custom styles.

Aniket-Engg:
Closing because of inactivity.

PR #5084, opened 2024-08-17
ui bug fixed

ijlal321:
### Hi, there was a text overlap in run tab section. Here is before and after.

Screenshot 2024-08-17 222825

After changing line Height

image

## Note: Text overlap only happens when there is a contract deployed.

Let me know if need any further details or changes. Thanks


joeizang (changes requested):
Nice try. But you cannot make changes to the theme css file. That's like editing a bootstrap.min.css file. Can you find a better way?

ijlal321:
> Nice try. But you cannot make changes to the theme css file. That's like editing a bootstrap.min.css file. Can you find a better way?

Thanks for clarification @joeizang. I have updated code.
let me know if you need anything else 👍

joeizang (changes requested):
Much better. But I think there should be a bootstrap class that can do this.

ijlal321:
Hi @joeizang. 
So there is no way of doing this using Bootstrap, because I believe remix uses custom theme CSS files which are from Bootstrap4. 

lh (line height class need to fix this bug) was introduced in bootstrap5.

Ofc we can make changes to theme css files to include lh class,  but i think my last approach using CSS selector is much better.

Also I am new to all this, so I am extremely grateful for all the guidance you have provided in last messages. Thanks

Aniket-Engg:
As title is updated to `Deployed Contracts`. This is no more required. Thanks!

Issue #4028, opened 2023-08-24
mismatch between RemixVM flavor in Environment and badge under Contract


ryestew:
<img width="313" alt="image" src="https://github.com/ethereum/remix-project/assets/6050001/6246058d-c56f-4c72-951b-974cda9df47f">
Environment says Shanghai and the badge says Paris.

Aniket-Engg:
@ryestew AFAIK This is not a bug. EVM version is shown from the compiler point of view.

EVM tag shows EVM version used to compile that contract, so that user can accordingly selected the right VM environment.

PR #5086, opened 2024-08-18
wrong execution Environment Bug Fixed

ijlal321:
### The selected execution environment always shows vm-cannun. Now its fixed and it correctly shows selected execution environment. Here is before and after for better explanation.  

Screenshot 2024-08-18 152143

image

Let me know if need any further changes or details about this pr.

also this issue was also described in this issue #4028


Aniket-Engg:
This is not a bug. EVM version is shown from the compiler point of view.

EVM tag shows EVM version used to compile that contract, so that user can accordingly selected the right VM environment.

ijlal321:
> This is not a bug. EVM version is shown from the compiler point of view.
> 
> EVM tag shows EVM version used to compile that contract, so that user can accordingly selected the right VM environment.

Got it @Aniket-Engg . I was just following issue https://github.com/ethereum/remix-project/issues/4028 so didn't thought about it. Thanks for clarification.

PR #5087, opened 2024-08-19
change git plugin from dGit to Git

jvc-byte:
Updated the code to use the actual name Git instead of dGit


Aniket-Engg:
@jvc-byte build is failing for the PR

Aniket-Engg:
build is still failing and there are conflicts to fix

PR #5129, opened 2024-08-26
Workspace selection fixes

### LINKED ISSUE #4266: Issue with Localhost workspace selection
- [x] For localhost workspace, we do not show any workspace options but in the workspace dropdown, there is option to create workspace, which doesn't work successfully. Either we remove it or make it work
- [x] Also, in dropdown we still say `connect to Localhost` and on home tab we say `Access File System`
- [x] Switching to a workspace from Localhost seems messy
- [ ] Recent workspaces gets messy while selecting a workspace after Localhost 
- [ ] I think we can still show the options to add scripts and workflow files even if localhost workspace is selected

joeizang:
fixes #4266

### DIFF
--- a/libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
+++ b/libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
@@ -917,22 +917,6 @@ export function Workspace() {
     _paq.push(['trackEvent', 'Workspace', 'GIT', 'login'])
   }
 
-  const CreateDropDownMenuItem = () => {
-    return (
-      <>
-        <Dropdown.Item
-          onClick={() => {
-            createWorkspace()
-          }}
-        >
-          <span className="pl-3">
-            {' '}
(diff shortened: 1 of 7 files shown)


libs/remix-ui/workspace/src/lib/remix-ui-workspace.tsx
```diff
@@ -1001,54 +1097,27 @@ export function Workspace() {
...
+                          currentWorkspace !== LOCALHOST && <CreateDropDownMenuItem />
+                        }
                         <Dropdown.Item
                           onClick={() => {
                             switchWorkspace(LOCALHOST)
                           }}
                         >
                           {currentWorkspace === LOCALHOST ? (
-                            <span>&#10003; localhost </span>
+                            <span>&#10003; Connected to Local Filesystem </span>
```
Aniket-Engg, line 1112: Is this change discussed?

LianaHus:
could you please put a divider before the Connect in hamburger? and also please remove -create workspace- from the dropdown

Aniket-Engg (approved):

PR #5147, opened 2024-08-28
fix free function option in right click menu

### LINKED ISSUE #5134: Free Functions not in rt click menu
**Describe the bug**
When you rt-click on a free function, the option to run it should be in the popup menu.  I cant get it to work.

**To Reproduce**
1. see this medium post: https://medium.com/remix-ide/remix-ide-v0-32-0-release-3385255dccb2
2. using this function in a file but outside of a contract:

```function testRun() pure {
    console.log("test running a free function");
}
```

You used to see this popup menu but now the free function option is not there (when right clicking in a free function)
image

Aniket-Engg:
fixes #5134


yann300 (approved):

ryestew:
@yann300 - should a function inside a contract be able to be run as a Free Function?  This current PR works when the function is outside the contract - but should all functions inside the contract have this option?  Was this the case before?

yann300:
This wasn't the case before and shouldn't be.

PR #5158, opened 2024-09-02
Add partner plugin

yann300:
- skip some check for plugins developped by partners.
- update label template selectioni plugin


Aniket-Engg:
It seems filter is broken on master too

LianaHus (approved):

PR #5159, opened 2024-09-02
Short description for settings -> general section

Aniket-Engg:
Shorten the text under options and move more info under tooltip

This will give more exposure to other items below


Aniket-Engg:
> I would suggest
> 
> * to remove the first alert.
> * to use same I icon everywhere. solidity, udapp and here . maybe somewhere else we have
> * @yann300 since you were initially against short texts please review as well

I have updated the icons as per the comment @LianaHus

LianaHus (approved):

Issue #4610, opened 2024-03-10
Upgrade `web3.js v1` to `v4`


SantiagoDevRel:
Hi guys, hope u are doing amazing, first of all, congrats on having this amazing project which is the first step for many devs into blockchain development (including me!)

I'm the DevRel of [web3.js](https://github.com/web3/web3.js), looking at your repo I see you are using web3.js is some of your files (and some are `web3 v1`, I noticed this since the way to import v4 is with curly brackets `v4 = import { Web3 } from "web3"` instead of `v1= import Web3 from "web3"`

Not sure if there might be some things in your repo that will require to upgrade to v4 besides the import statement, anyway I just wanted to give you a heads up on this, **since web3 v1 will be deprecated on April 1st.** [Here is a guide to upgrade from v1 to v4](https://docs.web3js.org/guides/web3_upgrade_guide/x/)

Not sure if you are already in contact with someone from our team, anyway, I'm here to help or you can get support from our team if needed, so Remix can keep thriving and fulfill its purpose to grow the whole Ethereum dev ecosystem.

Aniket-Engg:
cc: @yann300

PR #5160, opened 2024-09-03
update web3.js import

yann300:
fix https://github.com/ethereum/remix-project/issues/4610


Aniket-Engg (commented):
change set looks fine but e2e are failing

Aniket-Engg:
FYI @yann300 changing the import for web3 fails to execute the script. May be we need to first update the web3 version in script runner side

Aniket-Engg (approved):

PR #5161, opened 2024-09-04
Template Explorer Fixes

### LINKED ISSUE #5088: Templates plugin
1. Descriptions should be all capitalized and wording should be reviewed 
2. Name change: "Workspace templates"
3. New item in hamburger on the first place "Create blank", with the toiltip "Create blank workspace"
The second one can become "Create using template", with the tooltip "Create workspace using template"
4. Logo and Icon should be discussed.

joeizang:
fixes  #5088

### DIFF
--- a/apps/remix-ide/src/app/plugins/templates-selection/templates-selection-plugin.tsx
+++ b/apps/remix-ide/src/app/plugins/templates-selection/templates-selection-plugin.tsx
@@ -147,7 +147,7 @@ export class TemplatesSelectionPlugin extends ViewPlugin {
           ['ERC721', 'secondary'],
           ['ERC1155', 'primary'],
         ]}
-        title='Template explorer'
+        title='Workspace Templates'
         description="Select a template to create a workspace or to add it to current workspace"
       >
         {


ryestew:
I'll make some text corrections - but 

1. Why are there 4 contracts for each of the OZ templates?  The descriptions are slightly different from each other but the content - on a brief look is the same. 
2. The modal that used to be there to choose additional dependencies (like transferable or burnable) is gone.  That is a loss.  Could returning that be part of this PR?  Or would that expand the scope of this PR too much?

ryestew:
See these issues to check if either are in the scope of this PR:
https://github.com/ethereum/remix-project/issues/5171
https://github.com/ethereum/remix-project/issues/5170

ryestew:
The descriptions of the templates need a bit editing - but do we need adjectives like "simple" or "standard"? (that's a writing style question)
And is the word "Interface"  correct?  These are describing conracts not  interfaces -  right? An interface is the scaffold of a contract.  (this is a technical question)

ryestew:
Once we figure out if we have too many contracts here - then I can make some edits to the text and I'll specify the exact text with capitalization.

LianaHus:
@joeizang can you fix the order/name  according to 3rd item in the issue?

LianaHus:
I don't get it. I'm talking about label not data-id or any other kind of keyword. Why it should ve relevant? If it is we are doing something wrong there

ryestew (approved):

PR #5166, opened 2024-09-07
Hometab: Learneth slider & Learneth promo

ryestew:
The code needs to be organized better.  The cursor needs to be an pointer in the slider.


LianaHus:
Please change the image before merge

joeizang (changes requested):
fix e2e's and the pr is golden.

joeizang (approved):

bunsenstraat (dismissed):
no need to include list.json
you can run
git checkout master -- apps/remix-ide/src/assets/list.json 
then add the change commit push

vermouth22:
The image could be a lot larger... like it is for Beta Tester panel. Also, the cursor is still showing on rollover of "Start Learning" button instead of a pointer

Aniket-Engg:
> no need to include list.json you can run git checkout master -- apps/remix-ide/src/assets/list.json then add the change commit push

list.json removed

Aniket-Engg:
> The image could be a lot larger... like it is for Beta Tester panel. Also, the cursor is still showing on rollover of "Start Learning" button instead of a pointer

Image looks quite similar to one in beta tester panel. I update the cursor to pointer too

Aniket-Engg (approved):

PR #5168, opened 2024-09-09
fix Remix copilot status in status bar

Aniket-Engg:
- Open Remix and enable Remix co pilot, status is fine
- Reload Remix IDE
- When Remix is loaded and no file is selected, status is always `disabled`. In terminal, it says `activated`

This is fixed by this PR.

This PR also updates the wording a bit to use `enable & disable`

Terminal log to show info about Remix co pilot status is removed


vermouth22 (changes requested):
I think we decided that the name should be standardized to "RemixAI Copilot" so that should be what it says everywhere. In any case, it should be consistent here in the toolip and in the status bar...
image

image

Also, see above screenshots... In Status Bar on open, it says the RemixAI Copilot is enabled "(enabled)" but in the tooltip it says

Aniket-Engg:
> I think we decided that the name should be standardized to "RemixAI Copilot" so that should be what it says everywhere. In any case, it should be consistent here in the toolip and in the status bar... image
> 
> image
> Also, see above screenshots... In Status Bar on open, it says the RemixAI Copilot is enabled "(enabled)" but in the tooltip it says "to use... select a .sol file" and the switch is off rather than on. So, it is confusing when status bar says "(enabled)" but the switch at top bar is in off position. And, then when you do select a .sol file you get this at top... image
> 
> ...and status bar at bottom now shows image

I have updated the text to `RemixAI` everywhere. The other behaviour where Switch and status bar is not in sync, I cannot reproduce that. Do you see it in this PR's preview link?

Issue #5180, opened 2024-09-12
Maximized side panel


LianaHus:
1. Learneth and dGit should use the same code methods to maximize
2. Maximized Side Panel is too wide for those two. For the debugger, it is fine because of 2 columns.
3. click on icon on vertical panel to close. open -> it is narrow

PR #5187, opened 2024-09-13
refactor width updates

yann300:
fix https://github.com/ethereum/remix-project/issues/5180


LianaHus:
Yann those two work perfectly fine bud debugger doesn't maximize now
when i just started the debugger it was maximized but when i switched to others and back to the debugger it is enhanced not maximized

LianaHus (approved):

PR #5211, opened 2024-09-21
auto switch repo based on current language

drafish:
Here are the build-in repos 
```js 
export const repoMap = {
  en: {
    name: 'ethereum/remix-workshops',
    branch: 'master',
  },
  zh: {
    name: 'ethereum/remix-workshops',
    branch: 'cn',
  },
}
``` 
Just add them into `repoMap`, if there are any other repos of other languages.

@ryestew You need to update the chinese repo, it's not the latest.


apps/learneth/src/redux/models/workshop.ts
```diff
@@ -3,16 +3,26 @@ import { toast } from 'react-toastify'
...
 const apiUrl = 'https://static.220.14.12.49.clients.your-server.de:3000'
 
+export const repoMap = {
+  en: {
+    name: 'ethereum/remix-workshops',
+    branch: 'master',
+  },
+  zh: {
+    name: 'ethereum/remix-workshops',
+    branch: 'cn',
```
ryestew, line 18: please change this to zh.  I just pushed that branch.  Could you also add the es branch?

ryestew (changes requested):

yann300 (approved):

PR #5237, opened 2024-09-30
EIP 712

yann300:
- removes old way of signing.
- add right click / Sign typed data (EIP712 compatible).


yann300:
- should add matomo
- check the 32 bytes length message.

Aniket-Engg (approved):

PR #5242, opened 2024-10-02
Fixmatomo

bunsenstraat:
- the modals where drawing focus to themselves if they were not visible causing the matomo to blur out
- added a preventBlur option to the modal to prevent blur if needed, this is only used on matomo now...
- there is now a timestamp set in localstorage which indicates when the user has reacted to the matomo dialog. if it's older than 6 months and the matomo setting is either false or not set it will ask for matomo permissions again. added this in e2e. 
- added an event in matomo to show this happening
- loader.js would always consider permission granted when any value 'settings/matomo-analytics' was set, even when false.
- the modal cancel method now accepts an enum telling the callback what kind of cancel it was, close, escape, enter, click etc etc allowing. this allows the modal handler not to set the storage when it's a real blur.
- added e2e
- sometimes matomo was spelled matamo :D fixed that

### DIFF
--- a/libs/remix-ui/app/src/lib/remix-app/components/modals/modal-wrapper.tsx
+++ b/libs/remix-ui/app/src/lib/remix-app/components/modals/modal-wrapper.tsx
@@ -142,7 +142,7 @@ const ModalWrapper = (props: ModalWrapperProps) => {
     props.handleHide()
   }
 
-  if(!props.id || props.id === '') return null
+  if (!props.id || props.id === '') return null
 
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/tabs/settings-tab.tsx
```diff
@@ -108,6 +108,7 @@ module.exports = class SettingsTab extends ViewPlugin {
   }
 
   updateMatomoAnalyticsChoice(isChecked) {
+    console.log('isChecked', isChecked)
```
LianaHus, line 111: Remove?

PR #5244, opened 2024-10-04
fix matomo bug

bunsenstraat:
Bug: matomo dialogs where not showing up in some conditions. esp on firefox it seemed to be an issue.

- the modals where drawing focus to themselves if they were not visible causing the matomo to blur out
- added a preventBlur option to the modal to prevent blur if needed, this is only used on matomo now...
- there is now a timestamp set in localstorage which indicates when the user has reacted to the matomo dialog. if it's older than 6 months and the matomo setting is either false or not set it will ask for matomo permissions again. added this in e2e. 
- added an event in matomo to show this happening
- loader.js would always consider permission granted when any value 'settings/matomo-analytics' was set, even when false.
- the modal cancel method now accepts an enum telling the callback what kind of cancel it was, close, escape, enter, click etc etc allowing. this allows the modal handler not to set the storage when it's a real blur.
- added e2e
- sometimes matomo was spelled matamo :D fixed that

### DIFF
--- a/libs/remix-ui/app/src/lib/remix-app/types/index.ts
+++ b/libs/remix-ui/app/src/lib/remix-app/types/index.ts
@@ -8,6 +8,15 @@ export const enum ModalTypes {
   forceChoice = 'forceChoice'
 }
 
+export const enum AppModalCancelTypes {
+  close = 'close',
+  cancel = 'cancel',
+  blur = 'blur',
+  escape = 'escape',
+  enter = 'enter',
+  click = 'click'
+}
+
 export const enum UsageTypes {
   Beginner = 1,
   Prototyper,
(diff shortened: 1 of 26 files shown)


libs/remix-ui/app/src/lib/remix-app/types/index.ts
```diff
@@ -8,6 +8,15 @@ export const enum ModalTypes {
   forceChoice = 'forceChoice'
 }
 
+export const enum AppModalCancelTypes {
+  close = 'close',
+  cancel = 'cancel',
+  blur = 'blur',
+  escape = 'escape',
+  enter = 'enter',
+  click = 'click'
```
yann300, line 17: is `click` a click on cancel? and in this case what does `cancel` means?

yann300:
closing as the master PR is merged.

PR #5267, opened 2024-10-09
Fix vyper

yann300:
- fix rendering error (using the same renderer as solidity)
- readd e2e tests
- remove parsing pragma

### DIFF
--- a/apps/remix-ide-e2e/src/tests/vyper_api.test.ts
+++ b/apps/remix-ide-e2e/src/tests/vyper_api.test.ts
@@ -114,6 +114,25 @@ module.exports = {
     }
   },
 
+  'Should lead to a compilation error #group1': function (browser: NightwatchBrowser) {
+    browser
+      .clickLaunchIcon('filePanel')
+      .switchWorkspace('default_workspace')
+      .addFile('test_error.vy', { content: wrongContract })
+      .clickLaunchIcon('vyper')
+      // @ts-ignore
+      .frame(0)
+      .waitForElementVisible('[data-id="compile"]')
+      .click('[data-id="compile"]')
+      .waitForElementVisible({
+        selector:'[data-id="test_error.vy"]',
+        timeout: 60000
+      })
+      .waitForElementContainsText('[data-id="test_error.vy"]', 'ERROR')
+      .waitForElementContainsText('[data-id="test_error.vy"]', 'StructureException:Unsupported syntax for module namespace')
+      .frameParent()
+  },
+
   'Compile test contract and deploy to remix VM #group1': function (browser: NightwatchBrowser) {
     let contractAddress
     browser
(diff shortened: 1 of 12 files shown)


ryestew:
I don't know if this is a problem of this deployed instance or not but given this contract with an error of a verision of vyper that does not exist:
```# I'm a comment!

# SPDX-License-Identifier: MIT
# pragma version >0.4.0

my_favorite_number: uint256
```
compile in Vyper
Hit the Ask REmixAI button and Remix will crash
I leave this here because I dunno if this is on live.
For other issues, it gives a response and doesn't crash.

Stephane says:
There might be something going on in your deploy, on live thi does not crash
theoretically Remix AI wont lead to a crash since that is Request wrapped up with some processing

Furthermore the fact that vyper is not detected by the AI is because the handle is not correctly implemented you will have to modify the askGpt function in the Vyper plugin app as follows:

```
// TODO: remove! no formatting required since already handled on server
      const file = await this.client.call('fileManager', 'getCurrentFile')  
      const content = await this.client.call('fileManager', 'readFile', file)
      const messageAI = intl.formatMessage({ id: 'solidity.openaigptMessage' }, { content, message })

      console.log('new prompt', messageAI)
      // await this.client.call('remixAI' as any, 'error_explaining', message)
      await this.client.call('remixAI' as any, 'chatPipe', 'error_explaining', messageAI)
    } catch (err) {
      console.error('unable to askGpt')
      console.error(err)
```

Aniket-Engg:
> I don't know if this is a problem of this deployed instance or not but given this contract with an error of a verision of vyper that does not exist:
> 
> ```
> 
> # SPDX-License-Identifier: MIT
> # pragma version >0.4.0
> 
> my_favorite_number: uint256
> ```
> 
> compile in Vyper Hit the Ask REmixAI button and Remix will crash I leave this here because I dunno if this is on live. For other issues, it gives a response and doesn't crash.
> 
> Stephane says: There might be something going on in your deploy, on live thi does not crash theoretically Remix AI wont lead to a crash since that is Request wrapped up with some processing
> 
> Furthermore the fact that vyper is not detected by the AI is because the handle is not correctly implemented you will have to modify the askGpt function in the Vyper plugin app as follows:
> 
> ```
> // TODO: remove! no formatting required since already handled on server
>       const file = await this.client.call('fileManager', 'getCurrentFile')  
>       const content = await this.client.call('fileManager', 'readFile', file)
>       const messageAI = intl.formatMessage({ id: 'solidity.openaigptMessage' }, { content, message })
> 
>       console.log('new prompt', messageAI)
>       // await this.client.call('remixAI' as any, 'error_explaining', message)
>       await this.client.call('remixAI' as any, 'chatPipe', 'error_explaining', messageAI)
>     } catch (err) {
>       console.error('unable to askGpt')
>       console.error(err)
> ```
> 
> He says,"by the way this seems to be an internal processing error thrown on the UI, not even the compiler error"
> 
> But other than this problem with the AI, the it is working fine for me. So either we should remove the RemixAI from the error cards or we should make it so it doesn't crash.

Ask RemixAI works fine for me for this PR.

Screenshot 2024-10-10 at 1 36 31 PM

ryestew (approved):

PR #5270, opened 2024-10-10
load pinned contracts on workspace space change

Aniket-Engg:


libs/remix-ui/run-tab/src/lib/actions/events.ts
```diff
@@ -45,6 +45,9 @@ export const setupEvents = (plugin: RunTab) => {
     }
     setFinalContext(plugin, dispatch)
     fillAccountsList(plugin, dispatch)
+    // 'contextChanged' is triggered on workspace change
+    // Load pinned contracts for updated workspace
+    if (context.startsWith('vm')) await loadPinnedContracts(plugin, dispatch, context)
```
yann300, line 50: why do you check only this? we also have pinned contracts for other providers right?

yann300: what happen if you are using an injected provider?

yann300 (approved):

PR #5278, opened 2024-10-11
i fixed the error on the hand burger menu which resets the name set for create blank

praisedisu:


joeizang:
e2es are failing.

Aniket-Engg (approved):
change set looks fine

PR #5292, opened 2024-10-15
Desktop git status & init fixes

bunsenstraat:
- 'init' should be called through the 'dgit' plugin, not directly on the git api
- the statusbar git component is now linked to the git status via the app state because it didn't cover use cases on desktop


libs/remix-ui/statusbar/src/lib/components/gitStatus.tsx
```diff
@@ -11,70 +12,32 @@ export interface GitStatusProps {
 }
 
 export default function GitStatus({ plugin, gitBranchName, setGitBranchName }: GitStatusProps) {
-  const [isLocalHost, setIsLocalHost] = useState(false)
+  const appContext = useContext(AppContext)
 
-  useEffect(() => {
```
yann300, line 16: you don't need all this anymore?

bunsenstraat: no... it's all context driven

yann300 (approved):

PR #5296, opened 2024-10-15
Text alignment settings

praisedisu:


praisedisu:
looks like vertical-align is a bootstrap property, well, I've changed the property and updated my code and it's working fine

praisedisu:
Alright then

PR #5306, opened 2024-10-22
optional logo in quick-dapp

### LINKED ISSUE #5291: QuickDapp Change UI for optional logo input
Currently if no logo is chosen then the user thinks that it is required to use the Remix logo.  Only when you hover on the Remix logo do you see the instructions.  But if they don't have a logo - they are stuck.

Right now the UI looks like this:

image

A rough sketch of the UI should be more like this:

image

drafish:
fix #5291


apps/quick-dapp/src/actions/index.ts
```diff
@@ -192,7 +193,9 @@ export const deploy = async (payload: any, callback: any) => {
     files[`dir/${path}`] = resp.data;
   }
 
-  files['dir/assets/logo.png'] = logo
+  if (logo) {
```
yann300, line 196: there is no default case anymore?

ryestew:
The logo for uploading a logo is too large:
image

It also does not say that it is optional.

apps/quick-dapp/src/actions/index.ts
drafish: That's it. No more default logo. @yann300

apps/quick-dapp/src/actions/index.ts
```diff
@@ -353,7 +356,7 @@ export const initInstance = async ({
       natSpec,
       solcVersion: getVersion(solcVersion),
       ...lowLevel,
-      logo: logo.data,
+      // logo: logo.data,
```
LianaHus, line 359: please remove if not needed

apps/quick-dapp/src/components/DeployPanel/index.tsx
```diff
@@ -51,6 +51,8 @@ function DeployPanel(): JSX.Element {
     <div className="col-3 d-inline-block">
       <h3 className="mb-3" data-id="quick-dapp-admin">QuickDapp <FormattedMessage id="quickDapp.admin" /></h3>
       <Button
+        size="sm"
```
LianaHus, line 54: please use "btn btn-sm" instead of custom style

LianaHus (changes requested):

Aniket-Engg:
closing as #5376 is merged

PR #5324, opened 2024-10-24
Script runner custom

bunsenstraat:
First phase of customized script runner
- adds a plugin to sidepanel to configure script runner being used
- options are dynamically loaded from  'https://remix-project-org.github.io/script-runner-generator'
- new configurations are easily deployed by making a PR to the script runner repo https://remix-project-org.github.io/script-runner-generator
- to allow the user to generate complete customized builds and dependencies there's a UI already in this PR but are turned off because server is not ready to handle it. it can be ignored for now.
- when a TS/JS is loaded in tabs it shows a icon to configure the runner
- it features 'smart' loading, when there's an error loading a runner you can reload a runner and it shows an error icon and tooltip
- when a custom runner is selected it will create a config file in .remix specifying the runner so on reloads or switching workspace it loads the correct runner
- added a fontawesome icon to the kit but can be easily replaced by anything else
- some tests have been added, a lot more testing is done on the script runner itself, separate pipeline

Next phases are 
- allowing complete customized builds to run...
- downloading runners in desktop for offline use


libs/remix-simulator/src/methods/accounts.ts
```diff
@@ -85,6 +86,10 @@ export class Web3Accounts {
     }
   }
 
+  eth_requestAccounts (_payload, cb) {
```
yann300, line 89: I think these changes are already in master...

bunsenstraat: No this was part of your PR for ethers 6, it's not on master

bunsenstraat: https://github.com/ethereum/remix-project/pull/5188/files#diff-d91b9465dbf7e11a3da4f3c022a8649e7e509c5ccca17641898910875277e982R89

Issue #5327, opened 2024-10-24
Remix IDE freezes when a contract call error occurs


0xcodefive:
**Describe the bug**
When I try to call a contract function that returns an error, Remix IDE execution freezes (from https://remix.ethereum.org/). It doesn't matter what the error is, even with the banal incorrect Owner during the call. Previously, during such a call, a pop-up message with the error out of gas popped up. And the returned error was written to the ide console if, for example, `revert` contains its description.

Log output to the ide console:
```
transact to ERC721_Update.setNewCaller pending ... 
```

Log output to the browser console:

```
web3_request_manager.ts:432 Uncaught (in promise) ContractExecutionError: Error happened while trying to execute a function inside a smart contract
    at Web3RequestManager._isReverted (web3_request_manager.ts:432:48)
    at Web3RequestManager._processJsonRpcResponse (web3_request_manager.ts:373:23)
    at rejectWithError (web3_request_manager.ts:233:12)
```

After that, the ide freezes for 42 blocks (if I'm not confused). At the same time, any other calls are queued, but nothing happens.

**Desktop**
 - Windows 10 and Windows 11
 - Google Chrome Version 130.0.6723.70 (Official build), (64 bit) and Version 130.0.6723.59 (Official build), (64 bit)
 - Remix IDE v0.56.0
 - Metamask Version 12.4.1, Version 12.4.2, Version 12.5.0

Aniket-Engg:
@0xcodefive Do you face this error while using remix desktop application?

0xcodefive:
This error is not observed in Firefox 115.16.1esr

0xcodefive:
@Aniket-Engg
An easy way to repeat the mistake:

1. Connect Metamask
2. In default_workspace, select the `_Storage.sol `contract
3. Click `deploy `and cancel the transaction in Metamask

After that, no other transactions that require confirmation in the Metamask will be performed until the timeout timer is triggered:

```
creation of Storage errored: Error occurred: The connected Ethereum Node did not respond within 750 seconds, please make sure your transaction was properly sent and you are connected to a healthy Node. Be aware that transaction might still be pending or mined!
	Transaction Hash: not available.
```

Console log:
```
Uncaught (in promise) Error: MetaMask Tx Signature: User denied transaction signature.
    at _callee3$ (3456.0.56.0.1729697541829.js:1:613683)
    at tryCatch (main.0.56.0.1729697541829.js:1:3663661)
[...]

PR #5337, opened 2024-10-26
show error in the terminal and don't switch away

yann300:
1) When a provider call throw an error, show the msg in the terminal and don't switch back the provider to the VM.
2) see https://github.com/ethereum/remix-project/pull/5337/files#diff-546f5a668ded4d080ec3ec093fc2ce73756f4250bf82ad89394e698ab56ea3f7L203 web3.js has some trouble dealing with function marked with `async`..

To reproduce the first issue:
 - create a file and paste the content (see at the bottom) .
 - open custom http provider from udapp, paste the URL that is in the content file.
 - run the script
 - it fails and it disconnect.
 - it should show the error and not diconnect
 
fix https://github.com/ethereum/remix-project/issues/5327
To reproduce the second issue: 
 - compile something
 - deploy with metamask on mainnet (where you are sure to not have any ether)
 - it silently fails.

script to paste:
```
// Importing necessary libraries from Ethers.js for interaction with Ethereum blockchain.
import { ethers } from "hardhat";

// https://scroll-rpc.publicnode.com
async function main() {
  // Setting up provider (RPC URL) to interact with your chosen Ethereum chain, 
  const [deployer] = await ethers.getSigners();
  
   try{
    let provider;
     if(!provider){
      provider=ethers.provider;   
       }
       
     const contractAddress = "0x2bC16Bf30435fd9B3A3E73Eb759176C77c28308D"; // Replace with your smart contract's address.
    
     // Retrieving all events of a specific kind from the blockchain
  let logs = await provider.getLogs({address:contractAddress, fromBlock: '0x332f23',toBlock: '0x384410', topics: ['0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef']});
   console.log("Got Logs ",logs)
 }catch(error){console.error(`Error ${error}`)}

}

main()
```

### DIFF
--- a/apps/remix-dapp/src/utils/txRunner.ts
+++ b/apps/remix-dapp/src/utils/txRunner.ts
@@ -267,7 +267,7 @@ export class TxRunner {
       };
     } catch (error: any) {
       console.log(
-        `Send transaction failed: ${error.message} . if you use an injected provider, please check it is properly unlocked. `
+        `Send transaction failed: ${error.message || error.error} . if you use an injected provider, please check it is properly unlocked. `
       );
       return { error };
     }
--- a/apps/remix-ide-e2e/src/commands/hideMetaMaskPopup.ts
+++ b/apps/remix-ide-e2e/src/commands/hideMetaMaskPopup.ts
@@ -5,7 +5,7 @@ class HideMetaMaskPopup extends EventEmitter {
   command(this: NightwatchBrowser) {
     browser
       .pause(5000)
-      .isVisible({
+      .isPresent({
         selector: 'button[data-testid="popover-close"]',
         locateStrategy: 'css selector',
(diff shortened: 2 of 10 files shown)


Aniket-Engg (commented):
For HTTP provider case, It shows error in toaster but not in terminal. Also, once it shows error on UI, in browser console, it floods the errors `provider not set`. See screenshot:
Screenshot 2024-10-28 at 10 56 03 PM


Screenshot 2024-10-28 at 10 56 41 PM

For injected metamask, it show error in terminal but as `[object Object]` . See screenshot:

Screenshot_2024-10-28_at_10 49 35_PM

bunsenstraat (approved):

PR #5348, opened 2024-10-29
matomo updates

LianaHus:


bunsenstraat (approved):

bunsenstraat (changes requested):
you also need to fix matamo e2e test to fit the new events

bunsenstraat (approved):

PR #5356, opened 2024-10-30
verification plugin: Add Routescan verifier

manuelwedler:
We would like to also add Routescan as a verifier to the verification plugin. It supports the Etherscan API, so the `RoutescanVerifier` can simply inherit from `EtherscanVerifier`. There are just some nuances that needed to add additional checks to the EtherscanVerifier.

Also added a fix for the Matomo integration which was called in a loop leading to events being fired multiple times.


Aniket-Engg (commented):
There is a single explorer URL for routescan. For the testnet, it is: https://testnet.routescan.io/

rest PR looks fine

Aniket-Engg (approved):

PR #5391, opened 2024-11-07
EvmMap Patch

joeizang:


ryestew (approved):
I tried it on Scroll Sepolia and it worked in with this PR and it didn't work on live.

PR #5401, opened 2024-11-13
make height adjustment

### LINKED ISSUE #5397: Height of RemixAI UI is not Responsive
The RemixAI ui is overheight so it can't close this ui.

image

joeizang:
fixes #5397


pr15il (changes requested):
I think `maxHeight` should be kept at `40rem` and `height` changed to `67%`

apps/remix-ide/src/app/components/popup-panel.tsx
```diff
@@ -89,10 +89,10 @@ export class PopupPanel extends AbstractPanel {
       <div
         className={`px-0 bg-light border-info ${appState?.showPopupPanel ? 'd-flex' : 'd-none'}`}
         style={{
-          maxHeight: '40rem',
+          maxHeight: '60%',
```
pr15il, line 92:
```suggestion
          maxHeight: '40rem',
```

apps/remix-ide/src/app/components/popup-panel.tsx
```diff
@@ -89,10 +89,10 @@ export class PopupPanel extends AbstractPanel {
...
           maxWidth: '25rem',
           width: 'max-content',
-          height: '40rem',
+          height: '60%',
```
pr15il, line 95:
```suggestion
          height: '67%',
```

pr15il (changes requested):

apps/remix-ide/src/app/components/popup-panel.tsx
LianaHus: I think we can even remove max-height

apps/remix-ide/src/app/components/popup-panel.tsx
joeizang: Ok will have a look

Aniket-Engg (approved):
height is responsive now

pr15il:
When will it be released if I may know?

Issue #5414, opened 2024-11-22
Remix Crashes: after closing a tab and then right clicking in the FE


ryestew:
**Describe the bug**
After closing a tab, Remix crashes

**To Reproduce**
1. In any workspace, choose a file to open it in the editor
2. Sometimes just closing the file tab will crash remix
3. sometimes after closing a tab, you need to right click on the file in the FE
4. after point 2 or 3  above, Remix crashes

PR #5415, opened 2024-11-22
Fix crash on right-click after closing tab

dominikrudzki:
#5414 The application crashes because when a file is opened, focus is set on it. After closing the file, the browserReducer sets the focusElement in the state to an empty array. When you right-click in the file explorer, a condition tries to read the key of the first element in the focusElement array, which is empty, causing the crash.


ryestew:
Thanks so much for your timely contribution. Unfortunately, one of the Remix team members made a PR just before you did and that one got merged.  https://github.com/ethereum/remix-project/pull/5417/

We really appreciated your contribution and the explanation of your update.  Thanks so much!  I do hope you can contribute again.

PR #5419, opened 2024-11-26
fix vyper file deployment

### LINKED ISSUE #5418: UDapp's dial up state is not working when deploying with injected provider
**Describe the bug**
given this Vyper file:
```
#pragma version >0.3.10

storedData: public(int128)

@deploy
def __init__(_x: int128):
  self.storedData = _x

@external
def set(_x: int128):
  self.storedData = _x
```

Compile and go to Deploy & Run and then try to deploy using injected provider.
In the dialed up state of the constructor, it won't deploy (the Deploy but becomes active with inputting the param - but it doesn't do anything).  

Then dial down the constructor, input the param, and hit the transact button and it will work.

This is only a problem with a Vyper file.  
If a Solidity file in the same workspace has been compiled, then fixes the Vyper file not being able to send a transaction from the dialed up state.

@bunsenstraat - could this have something to do with the script runner or Ethers?

Aniket-Engg:
`status` here `const status = await props.getCompilerDetails()` is undefined for any file other that a Solidity file

fixes #5418

### DIFF
--- a/libs/remix-ui/run-tab/src/lib/actions/events.ts
+++ b/libs/remix-ui/run-tab/src/lib/actions/events.ts
@@ -153,16 +153,12 @@ export const setupEvents = (plugin: RunTab) => {
   })
 
   plugin.on('fileManager', 'currentFileChanged', (currentFile: string) => {
-    if (/.(.abi)$/.exec(currentFile)) {
-      dispatch(setLoadType('abi'))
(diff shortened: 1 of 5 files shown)


LianaHus (approved):
code review

LianaHus (approved):

ioedeveloper (approved):

ryestew (approved):

PR #5421, opened 2024-11-26
refactoring of styles and ui fixes in Vyper, Solidity and themes

LianaHus:


libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
```diff
@@ -979,32 +979,47 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
...
-                  type="number"
-                  title={intl.formatMessage({ id: 'solidity.inputTitle2' })}
-                  onChange={(e) => onChangeRuns(e.target.value)}
-                  disabled={!state.optimize || state.useFileConfiguration}
-                />
+                <CustomTooltip
+                  placement="top"
+                  tooltipId="configfileOptimisationNumbeTooltip"
+                  tooltipClasses="text-nowrap"
+                  tooltipText={(state.optimize ? "Enable " : "Disable ") + "opcode-based optimizer for the generated bytecode and the Yul optimizer for the Yul code"}
```
Aniket-Engg, line 988: I think this condition should be reversed as: `state.optimize ? "Disable" : "Enable"`

as if `state.optimize` is true, it means it is already enabled and can be disabled

LianaHus: Updated. thanks

Aniket-Engg (approved):

PR #5481, opened 2024-12-10
enabled code completion by default

STetsing:
all user will be having code completion enabled by default for solidity, vyper and circom files


STetsing:
@ryestew pls take this into account for the next release by educating users in the release notes how to turn off completion

Aniket-Engg (approved):

STetsing:
@Aniket-Engg this need to be checked again, the copilot is not enabled by default

PR #5507, opened 2024-12-18
update Contributing.md

ryestew:
Please check if the explanation of adding a new language is correct.  I think there might be some problems.


ryestew (commented):
Check line 57 about adding French.  I believe languages are added only in CrowdIn - but if you are making a new plugin you'll need to set up the new language's folder.  @LianaHus - could you confirm this?

CONTRIBUTING.md
```diff
@@ -1,23 +1,36 @@
 # Contributing
 
-Everyone is very welcome to contribute on the codebase of Remix. Please join our [Discord](https://discord.gg/mh9hFCKkEq) in case of any queries.
+Everyone is welcome to contribute Remix's codebase. Please join our [Discord](https://discord.gg/mh9hFCKkEq) in case of any queries.
```
joeizang, line 3: I think the sentence should read "Everyone is welcome to contribute to Remix's codebase"

CONTRIBUTING.md
```diff
@@ -1,23 +1,36 @@
...
-Remix libraries work closely with [Remix IDE](https://remix.ethereum.org). Each library has a readme to explain its application.
+Remix libraries work closely with [Remix IDE](https://remix.ethereum.org). Each library has a README to explain its application.
 
-When you add code in any library, please ensure you add related unit tests.
+When you add code to a library, please add related unit tests.
 
 ## Coding style
 
-Please conform to [standard](https://standardjs.com/) for code styles.
+Please conforms to the [standard](https://standardjs.com/) for code styles.
```
joeizang, line 12: I feel this sentence should read a lot better than it does currently.

CONTRIBUTING.md
```diff
@@ -80,34 +93,21 @@ You can find the language's `code, name, localeName` in this link
...
 ```jsx
 <h6 className="pt-0 mb-1" data-id='sidePanelSwapitTitle'>
  <FormattedMessage id={plugin?.profile.name + '.displayName'} defaultMessage={plugin?.profile.displayName || plugin?.profile.name} />
 </h6>
 ```
-You can't be sure whether there is a match key in locale file or not. So it will be better to provide a `defaultMessage` prop.
-
-### Should I update the non-english locale json files?
-You probably will have this question when you are updating the english locale json files. 
+You can't be sure whether there is a match key in locale file or not. So its better to provide a `defaultMessage` prop.
```
joeizang, line 108: This might read better this way: "You can't be sure whether there is a matched key in the locale file or not..."

joeizang (changes requested):
I left a few comments regarding some sentences. But the javascript inside the index.js files isn't complete.

joeizang (approved):
this seems fine for now. It can be improved in another pr

PR #5511, opened 2024-12-18
fix typo groth16_trusted_setup.ts

sky-coderay:
Hope this helps streamline the project!


libs/remix-ws-templates/src/templates/rln/scripts/groth16/groth16_trusted_setup.ts
```diff
@@ -1,4 +1,4 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
+// eslint-disable-next-line @typescript-eslnt/no-var-requires
```
Aniket-Engg, line 1: not fixing typoe

PR #5526, opened 2024-12-20
fix empty sections in gridview

bunsenstraat:
fix empty sections in grids ( on filter too )
fixed some react key bugs


Aniket-Engg (commented):
Screenshot_2024-12-20_at_1 23 19_PM

Can you do it when filter is used?

yann300 (approved):

PR #5591, opened 2024-12-30
i18n to Contract Verification

### LINKED ISSUE #5522: Contract Verification plugin needs to be setup with i18n
The strings of Contract Verification - this should be done with a possible update in the instructions of Contributing.md for how to hook up the strings to the translation platform.

joeizang:
fixes #5522


apps/contract-verification/src/app/components/ConstructorArguments.tsx
```diff
@@ -102,7 +103,10 @@ export const ConstructorArguments: React.FC<ConstructorArgumentsProps> = ({ abiE
       <div className="d-flex py-1 align-items-center custom-control custom-checkbox">
         <input className="form-check-input custom-control-input" type="checkbox" id="toggleRawInputSwitch" checked={toggleRawInput} onChange={() => setToggleRawInput(!toggleRawInput)} />
         <label className="m-0 form-check-label custom-control-label" style={{ paddingTop: '2px' }} htmlFor="toggleRawInputSwitch">
-          Enter raw ABI-encoded constructor arguments
+          <FormattedMessage
```
ryestew, line 106: is the default message needed?  See Drafish's comment:
```
// There may have some un-translated content. Always fill in the gaps with EN JSON.
// No need for a defaultMessage prop when render a FormattedMessage component.
```

joeizang: I noted that when going through the docs, but without the defaultMessage, there are console errors for every FormattedMessage block added. I will look into it and see

joeizang: @ryestew I noted from the error I get when the FormattedMessage tags are created without a defaultMessage prop that it falls back to using the id prop as the default. See this link which is the documentation page of the package we are using [react-intl](https://formatjs.github.io/docs/react-intl/) (scroll to the bottom to see code examples). In the examples shown, defaultMessage prop is used to give a fallback. So we might need to update our docs.

This here also shows the fallback behaviour of the libarary [react-intl fallback behaviour](https://formatjs.github.io/docs/react-intl/components/#message-formatting-fallbacks)

ryestew (approved):
I believe the i18n is implemented correctly.

PR #5600, opened 2025-01-03
delete forked state environment from env explorer

Aniket-Engg:
This will allow to delete forked state provider from Environment Explorer

For forked state, it will show this button:

Screenshot 2025-01-07 at 3 03 44 PM

It will also delete the state file from workspace.


apps/remix-ide/src/blockchain/blockchain.tsx
```diff
@@ -687,6 +687,7 @@ export class Blockchain extends Plugin {
   }
 
   removeProvider(name) {
+    if (this.pinnedProviders.includes(name)) this.emit('shouldRemoveProviderFromUdapp', name, this.getProviderObjByName(name))
```
yann300, line 690: this is already emitted by `on('environmentExplorer', 'providerUnpinned'`. Could you explain why you need this here?

yann300 (approved):

ryestew:
the following question could arise, should we leave the delete & fork buttons after a VM has been forked - so you can fork a fork or delete it.  With this PR, at least if you can find the Environment Explorer, you can delete.  And as I use this functionality, it makes sense that the delete button should not be available to a fork like you can to delete the VM state.  
As for forking the fork...its probably also not so important.

ryestew (approved):
The functionality of this PR works for me.

LianaHus (approved):

PR #5601, opened 2025-01-03
Verification plugin: Add new icon and feature plugin

manuelwedler:
We wanted to give the plugin a prettier icon.
Besides this, it would be nice if we could feature the plugin on the home view of Remix, since it is still quite new.


ryestew (changes requested):

Aniket-Engg (approved):

Aniket-Engg:
Screenshot 2025-01-13 at 6 58 06 PM

Looks good

PR #5602, opened 2025-01-05
feat: add file path validation for improved security

crStiv:
- Add validatePath method to check file paths
- Protect against path traversal attacks
- Validate special characters and path length
- Add unit tests


Aniket-Engg:
build fails @crStiv

PR #5603, opened 2025-01-06
feat: improve error handling in TraceManager with TraceError class

DeVikingMark:
Description:
- Add TraceError class for better error handling and debugging
- Include step index in error objects for improved context
- Enhance error messages in buildCallPath, getCallStackAt and getStackAt methods
- Preserve error context instead of string conversion

Benefits:
- Better debugging experience with detailed error information
- Type safety with specialized error class
- Consistent error handling approach


libs/remix-debug/src/trace/traceManager.ts
```diff
@@ -167,9 +175,8 @@ export class TraceManager {
         stack.reverse()
         return stack
       }
-    } else {
-      throw new Error('no stack found')
     }
+    throw new TraceError('No stack data found for the given step index', stepIndex)
```
yann300, line 179: I think this needs to be in a `else` statement.

PR #5607, opened 2025-01-06
overrided the nlux clipboard button

STetsing:
On each code block the copy to clipboard is now available


Aniket-Engg (commented):
copy doesn't work for me on preview link

STetsing:
@Aniket-Engg copy only works atm only after the chat stream is fully rendered

Aniket-Engg:
> @Aniket-Engg copy only works atm only after the chat stream is fully rendered

Ok I see it now.

Aniket-Engg (approved):

PR #5613, opened 2025-01-07
Completion enabled default

STetsing:
fixes https://github.com/ethereum/remix-project/pull/5481


libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
```diff
@@ -88,6 +88,11 @@ export const TabsUI = (props: TabsUIProps) => {
 
   const getAI = async () => {
     try {
+      const init_state = await props.plugin.call('settings', 'getCopilotSetting')
+      if (init_state === undefined || init_state === null) {
+        await props.plugin.call('settings', 'updateCopilotChoice', ai_switch)
+        return ai_switch
+      }
       return await props.plugin.call('settings', 'getCopilotSetting')
```
Aniket-Engg, line 96: I think you can just return `init_state ` here

STetsing: An initial state is required for the very first time the user load the page.

Aniket-Engg (approved):

PR #5629, opened 2025-01-09
switch instead of check for gridview

LianaHus:
Check on Deploy and run -> Environment Explorer
Screenshot from 2025-01-08 13-39-24
Screenshot from 2025-01-08 13-38-08


joeizang (approved):
LGTM

PR #5633, opened 2025-01-09
activate sol testing when starting learneth

bunsenstraat:


ryestew (approved):
This PR fixes the problem and works for me.

PR #5639, opened 2025-01-10
Udapp disconnect fix

### LINKED ISSUE #5483: Switch to VM if Remix is disconnected from injected provider
Currently, the account list is empty, and the remix is in an ambiguous state.
When connection is lost Remix should switch to vm

joeizang:
fixes #5483

### DIFF
--- a/libs/remix-ui/run-tab/src/lib/run-tab.tsx
+++ b/libs/remix-ui/run-tab/src/lib/run-tab.tsx
@@ -235,6 +235,16 @@ export function RunTabUI(props: RunTabProps) {
     }
   }, [runTab.popup])
 
+  useEffect(() => {
+    if (runTab.selectExEnv.includes('injected') &&
+      Object.entries(runTab.accounts.loadedAccounts).length === 0 &&
+    runTab.accounts.selectedAccount.length > 0) {
+      // switch to vm-cancum because no account is loaded from injected provider
+      const context = plugin.blockchain.defaultPinnedProviders[0] // vm-cancun
+      setExecutionEnvironment({ context, fork: '' })
+    }
+  }, [runTab.accounts.loadedAccounts])
+
   const setCheckIpfs = (value: boolean) => {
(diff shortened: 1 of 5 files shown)


libs/remix-ui/run-tab/src/lib/run-tab.tsx
```diff
@@ -235,6 +235,16 @@ export function RunTabUI(props: RunTabProps) {
     }
   }, [runTab.popup])
 
+  useEffect(() => {
+    if (runTab.selectExEnv.includes('injected') &&
```
LianaHus, line 239: what about wallets?

libs/remix-ui/run-tab/src/lib/run-tab.tsx
```diff
@@ -235,6 +235,16 @@ export function RunTabUI(props: RunTabProps) {
...
+      Object.entries(runTab.accounts.loadedAccounts).length === 0 &&
+    runTab.accounts.selectedAccount.length > 0) {
+      // switch to vm-cancum because no account is loaded from injected provider
+      const context = plugin.blockchain.defaultPinnedProviders[0] // vm-cancun
```
LianaHus, line 243: can we load load the previously selected or it will be difficult to do?

joeizang: It would just require some more code to track any account that is selected and persist it outside the reach of a useEffect.

libs/remix-ui/run-tab/src/lib/run-tab.tsx
joeizang: I am not sure I understand what you mean by wallets.

joeizang:
The other issues with walletconnect will be addressed in a separate pr. See this issue https://github.com/ethereum/remix-project/issues/5666 @LianaHus

LianaHus (approved):

PR #5652, opened 2025-01-16
Scriptrunner as main plugin

bunsenstraat:
Scriptrunner should be moved from side panel to mainview bu ux team request. This is a refactoring that needed to start to work on the UI. (Liana)


apps/remix-ide/src/app/panels/tab-proxy.js
```diff
@@ -201,6 +203,7 @@ export default class TabProxy extends Plugin {
   }
 
   focus (name) {
+    console.log('focus', name)
```
LianaHus, line 206: do we need this?

apps/remix-ide/src/app/panels/tab-proxy.js
```diff
@@ -164,7 +164,9 @@ export default class TabProxy extends Plugin {
     })
 
     this.on('manager', 'pluginActivated', ({ name, location, displayName, icon, description }) => {
+      
       if (location === 'mainPanel') {
+        console.log('pluginActivated', name, location, displayName, icon, description)
```
LianaHus, line 169: and this?

LianaHus (approved):
concole.logs we will cleanup after review with Filip

PR #5653, opened 2025-01-16
fix typo  EtherscanVerifier.ts

futreall:
constructorArguements - constructorArguments


Aniket-Engg:
this is intentional as there is typo in required API param too

PR #5714, opened 2025-01-29
Workspace completion

STetsing:
The user workspace is indexed in a regular time period. More context file is used for creating suggestions


libs/remix-ai-core/src/agents/completionAgent.ts
```diff
@@ -1,23 +1,132 @@
...
+    this.props.on('filePanel', 'workspaceRenamed', () => { this.indexed = { isIndexed: false, reason:"workspaceRenamed" }});
+    this.props.on('filePanel', 'workspaceDeleted', () => { this.indexed = { isIndexed: false, reason:"workspaceDeleted" } });
+  }
 
-  constructor(codebasePath: string) {
-    // git or fs
-    this.codebase = this.loadCodebase(codebasePath);
+  async getDcocuments() {
+    const documents: Document[] = [];
+    const dirs = await this.props.call('fileManager', 'dirList', '/');
```
yann300, line 62: did you try to call `await plugin.call('fileManager', 'copyFolderToJson', '/')` if should returns a JSON value with all the files.

libs/remix-ui/remix-ai/src/lib/components/dynamicButtons.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <button key={index} onClick={button.onClick}>
+          {button.icon && <i className={button.icon}></i>}
+          {button.label}
+        </button>
+      ))}
+    </div>
+  );
+};
+
+export default DynamicButtons;
```
yann300, line 22: where is this used?

package.json
```diff
@@ -153,6 +153,7 @@
     "express-ws": "^5.0.2",
     "file-path-filter": "^3.0.2",
     "file-saver": "^2.0.5",
+    "flexsearch": "^0.7.43",
```
yann300, line 156: do you need that dependency?

libs/remix-ai-core/src/agents/completionAgent.ts
```diff
@@ -1,23 +1,132 @@
...
+      const searchResult = this.indexer.search(content)
+      const fcps = await this.processResults(searchResult, currentFile);
+      const resolvedFcps = await Promise.all(fcps);
+      return resolvedFcps;
+    } catch (error) {
+      return [];
+    }
   }
+
+  async processResults(results: any, currentFile: string) {
```
yann300, line 115: can you add a doc to this function to explain what is this about?

package.json
STetsing: Actually no, these are the bunch of alternative I tried out

libs/remix-ui/remix-ai/src/lib/components/dynamicButtons.tsx
STetsing: That slipped in from other work (adding context in chat)

PR #5727, opened 2025-02-01
fix: correct CheckTxStatus import paths

MarsonKotovi4:
I noticed that the imports for `CheckTxStatus` were slightly off in `RenderCall.tsx` and `RenderKnownTransactions.tsx`. 
This PR just fixes those paths so everything references the right file. No functional changes—just tidying up!


Aniket-Engg:
import is done as per the file name

PR #5747, opened 2025-02-10
Fix loading deps which use module name remapping

yann300:
- L273: check if the version contains `npm:` and replace the value if this is the case.
- resolving deps would only work before for dependencies that are specified in the package.json, not the sub-deps present in yarn or in package-lock.json.
So instead of looking at the deps L150 and then trying to resolve the version, we just infer the package name from the fetchUrl.


bunsenstraat:
can we cover these use cases with tests?

bunsenstraat (approved):

PR #5754, opened 2025-02-11
refactor & allow saving already forked state

yann300:
- refactor: set a `config` property instead of having all the properties flat.
- refactor the `Provider` type.
- add a new property: `isRpcForkedState`, `statePath` and `isVMStateForked`


Aniket-Engg (approved):
LGTM

PR #5758, opened 2025-02-11
fix metamask signing

yann300:
This fixes signing message with metamask.
I've got confused on one of my last PR, this causes to break metamask signing.
The reason is that even if metamask uses `personal.sign`, it doesn't not make use of the password property, what I was thinking to be a bug was actually correct..


Aniket-Engg (approved):
LGTM

PR #5778, opened 2025-02-17
Workspace Dropdown Fix

joeizang:
Fixes buggy artifact found in the workspace dropdown that enables users to create remixd connection.


ryestew:
This does fix the problem,but the dev who reported it, didn't know what that hamburger menu was for - so never clicked it.  He uses remixd - so without the hack he found, it would look like Remixd got discontinued.  

We need to have "Connect to Localhost" option as the first option in the Workspaces select box.

ryestew (approved):

PR #5779, opened 2025-02-17
fix(remixd): improve file system synchronization and folder detection

### LINKED ISSUE #5765: Remixd probs coordinating between FE and File System
**Describe the bug**
1. When coding in an editor and then saving it, Remix will ask if you want to update Remix. 
2. (Question 1.If you say no - these file diverge -where is the diverged file of the same name saved?)
3. If you programatically update the folder - for example with npx hardhat compile - which should generate an artifacts folder, Remix does not show this folder.

**To Reproduce**
1. Share a hardhat project folder via Remixd 
2. Check in Remix that there is no artifacts folder
3. Compile in HH and see the artifacts folder
4. In Remix the artifacts folder does not show up

VolodymyrBg:
Fixes #5765

This PR addresses two issues with remixd file system synchronization:

1. File divergence when saving and choosing not to update in Remix
- Added file polling to ensure reliable change detection
- Added write stabilization to prevent premature change events
- Improved initial state capture

2. New folders (like artifacts) not showing up in Remix
- Enabled directory creation event handling
- Added proper file creation event handling
- Added state checks to ensure events are processed at the right time

The changes improve the reliability of the file system synchronization between the local filesystem and Remix IDE.

### DIFF
--- a/libs/remixd/src/services/remixdClient.ts
+++ b/libs/remixd/src/services/remixdClient.ts
@@ -251,20 +251,28 @@ export class RemixdClient extends PluginClient {
     const absPath = utils.absolutePath('./', path)
 
     if (!isRealPath(absPath)) return
-    this.watcher = chokidar.watch(path, { depth: 2, ignorePermissionErrors: true })
+    this.watcher = chokidar.watch(path, {
+      depth: 2,
+      ignorePermissionErrors: true,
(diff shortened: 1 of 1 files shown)


VolodymyrBg:
@andremedeiros @axic @step21 @evertonfraga

PR #5808, opened 2025-02-25
Verification plugin: Disable verification when Etherscan is selected and no constructor arguments are provided

manuelwedler:
We wanted to improve the user feedback on the verification plugin a little. Etherscan is the only verifier that requires constructor arguments. Before this change, the user could hit the verify button even if no constructor arguments are provided. With this change, the verify button gets disabled if Etherscan is selected and no constructor arguments are provided.

### DIFF
--- a/apps/contract-verification/src/app/components/ConstructorArguments.tsx
+++ b/apps/contract-verification/src/app/components/ConstructorArguments.tsx
@@ -44,12 +44,16 @@ export const ConstructorArguments: React.FC<ConstructorArgumentsProps> = ({ abiE
 
   const constructorArgsInInitialState = useRef(true)
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


kuzdogan:
Should we be a bit more explicit by saying "Etherscan requires constructor arguments" ?

manuelwedler:
> Should we be a bit more explicit by saying "Etherscan requires constructor arguments" ?

What do you have in mind? Something like "Please provide the constructor arguments when verifying with Etherscan."?

kuzdogan:
@manuelwedler not like please do this but more Etherscan verification requires this.

How about: "Must provide constructor arguments if verifying on Etherscan"

Aniket-Engg (approved):
Functionality wise it looks fine

PR #5811, opened 2025-02-26
Address WalletConnect Udapp disconnection 

### LINKED ISSUE #5666: Injected Provider Account Selection
When Udapp plugin is connected to a wallet using the Injected Provider and if for any reason it disconnects, the environment dropdown still shows Injected provider but the accounts list is empty. This is true for walletconnect and metamask. This unwanted behaviour has been addressed in #5639 
Also there is no tracking of previously selected account in a more permanent way.

This issue would be considered addressed if:

- [ ]  For any disconnection, the fallback account should be to switch to latest vm for walletconnect.
- [ ] Previously selected accounts from the Account List should be tracked in a less ephemeral way.

joeizang:
fixes #5666


LianaHus:
the issue is fixed but we still have a tab in the background. can't we show this widget as a real modal?

Issue #5810, opened 2025-02-26
Issue with file switching and default Tabs


Aniket-Engg:
If I select a file from FE, it is opened in editor.

Select Home tab in editor.

click on same file in FE, it doesn't open the file in editor but stays to HOME tab

It is same if we use HOME or environment explorer or similar tabs

PR #5827, opened 2025-02-28
fix: issue with file switching and default Tabs

gap-editor:
solves #5810


libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
```diff
@@ -31,6 +31,7 @@ export interface Tab {
 export interface TabsUIApi {
   activateTab: (name: string) => void
   active: () => string
+  activateByTitle: (title: string) => void
```
LianaHus, line 34: we should work only with name as this is a key.

PR #5844, opened 2025-03-06
Fix Udapp Environment Label Dropdown bug

### LINKED ISSUE #5736: Environment Select box with L2-Optimism
**Describe the bug**
Mismatch between select box and the chain connected
Image

**To Reproduce**
1. Select L2 optimism
2. in metamask, change the chain
3. the chain is chained in the Environments but the selected item in the select box is still L2 - Optimism

**Expected behavior**
When there is a change in the chain in Metamask, the choice in the select box should just change to Injected Provider.

joeizang:
fixes #5736

- Creates a new stateful DropDownLabel component
- Updates label when chainId changes from metamask.


Aniket-Engg (commented):
Screenshot 2025-03-10 at 11 20 21 PM

rocket icon stays even on changing network from L2

Aniket-Engg (commented):
When I select Injected Provider - Metamask, Remix IDE breaks for me

PR #5845, opened 2025-03-06
fixing error feedback in script runner and vertical icons empty tooltip

### LINKED ISSUE #5700: env explorer icon has different colors on tab and plugin (check both themes)

### LINKED ISSUE #5805: Alignment of right click menu of vertical panel icons
should be left.
+ some margins

LianaHus:
- click on the row should work for radio button change
- error comes to the same line
- text changed according to wireframe. a bit shorter because of lack of space
- all items aligned on the script title line
- when script isn't loaded the old one is still chosen
-
- wrong color for mainview plugins icon fixed, change themes to check fixes #5700
- vertical icon bar. do not show empty context menu and align. fix #5805


joeizang (changes requested):
There is no context menu for icons on the vertical icons panel, is this the desired goal for the pr? Also can you add a description for the pr?

joeizang (approved):
LGTM

PR #5848, opened 2025-03-07
Add Matomo Tracking to FE & FE Context Menu

### LINKED ISSUE #5712: Right click in the FE
I'm not sure if the right click in the FE is recorded in matomo.  The specific items are collected, but the right click should also be.  I think many people don't know that they should click there.  So lets get a baseline.

joeizang:
fixes #5712

- Simply add matomo tracking to the right click context menu.
- Add matomo tracking to a few other FE actions


LianaHus:
are those changes for FE file menu items?
fileAction ?

joeizang:
file explorer menu? No. I think context menu actions only, if I am not mistaken.

PR #5858, opened 2025-03-11
Fix OpenZeppelin ERC20 Contract Path

Dahka2321:
This PR updates the terminal.test.ts file to correct the path for opening the OpenZeppelin ERC20 contract in the Remix IDE. The previous file path was incorrect, and this fix ensures that the contract is accessed from the proper GitHub repository structure.


Aniket-Engg:
e2e fails because of the change

PR #5943, opened 2025-04-09
sanitize

STetsing:
fixes `what the following code does  ?`


libs/remix-ui/remix-ai/src/lib/components/Default.tsx
```diff
@@ -9,6 +9,7 @@ import './color.css'
 import '@nlux/themes/unstyled.css';
 import copy from 'copy-to-clipboard'
 
+
```
Aniket-Engg, line 12: Remove this

Aniket-Engg (commented):
I think you need to add `dompurify ` as a direct dependency in package.json too.

Also fix the linting please

Aniket-Engg (approved):
LGTM

Issue #5718, opened 2025-01-30
AA integration with Gnosis Safe


Aniket-Engg:
**Objective**
Allow creation and usage of Gnosis safe smart account withIn Remix

**MVP Scope**
Develop a first version for Ethereum `Sepolia` test network

Aniket-Engg:
Next step will be to support Gnosis chiado testnet

PR #5945, opened 2025-04-09
proxy URL for bundler 

Aniket-Engg:
related to #5718


libs/remix-ui/run-tab/src/lib/actions/account.ts
```diff
@@ -111,22 +111,25 @@ export const createNewBlockchainAccount = async (plugin: RunTab, dispatch: React
 export const createSmartAccount = async (plugin: RunTab, dispatch: React.Dispatch<any>) => {
   const localStorageKey = 'smartAccounts'
   const PUBLIC_NODE_URL = "https://go.getblock.io/ee42d0a88f314707be11dd799b122cb9"
-  const PIMLICO_API_KEY =''
-  const BUNDLER_URL = `https://api.pimlico.io/v2/sepolia/rpc?apikey=${PIMLICO_API_KEY}`
+  const toAddress = "0xAFdAC33F6F134D46bAbE74d9125F3bf8e8AB3a44" // can be zero address too
```
yann300, line 114: can you put more information about which address is that?

yann300 (approved):

libs/remix-ui/run-tab/src/lib/actions/account.ts
Aniket-Engg: added

PR #5946, opened 2025-04-09
Workspaces Template Explorer OZ Options Restore

### LINKED ISSUE #5171: Modal to add dependencies to OZ contract templates is no longer present
**Describe the bug**
The modal to choose burnable mintable etc is gone.  It should be put back after the template is chosen.

joeizang:
fixes #5171

- Remove all possible combinations from template explorer(including info on cells) and leave only one per class
- all options the user can choose from will be added to the modal we currently use for the naming of workspace


Aniket-Engg (commented):
Screenshot 2025-04-14 at 8 47 41 PM

increase the margin between textbox and customize template

ryestew (changes requested):
- The checkbox to initialize as a git repo does not work. (It should works even if you are not signed in).

- There should be a little more vertical padding (or margin) between the name of the template and the text "Customize template" 
image


- The word "simple" should be removed from all the choices because you can add options in this modal
image

joeizang:
> * The checkbox to initialize as a git repo does not work. (It should works even if you are not signed in).
> * There should be a little more vertical padding (or margin) between the name of the template and the text "Customize template"
> 
> image
> * The word "simple" should be replaced with "customizable" for all the choices.
> 
> image

You cannot use the git feature without signing in @ryestew

Aniket-Engg (approved):
functionality wise looks fine to me

ryestew:
@joeizang - you can initialize a repo with Git without signing in.  You can do that now on live and in your PR, after you make a new workspace (where you can't click the initialize checkbox), you can click the initialize button at the top of the FE.

This is a small point - and it should not delay the release.  So if there is time, fix it now.

ryestew (approved):

ryestew:
The initialize checkbox is currently operating the way it is as it is in this PR - so we should merge this PR and decide how we want to use the initialize button separately.

PR #5978, opened 2025-04-28
Refactor AA code to support more chains

Aniket-Engg:
Related to https://github.com/ethereum/remix-project/issues/5718#issuecomment-2824245890


libs/remix-lib/src/execution/txRunnerWeb3.ts
```diff
@@ -203,8 +203,12 @@ export class TxRunnerWeb3 {
             callback(new Error('Gas estimation failed because of an unknown internal error. This may indicated that the transaction will fail.'))
             return
           }
-          if (tx.fromSmartAccount && tx.value === "0" && err && err.error && err.error.indexOf('insufficient funds for transfer') !== -1) {
-            // Do not show dialog for insufficient funds as smart account may be using paymaster
+          if (tx.fromSmartAccount && tx.value === "0" && (
```
yann300, line 206: I think we need to find a better way to check that...
below you are checking against `err.error` and not `err`..
should that be: `err && err.error..indexOf('gas required exceeds allowance (0)')` ?

libs/remix-lib/src/execution/txRunnerWeb3.ts
```diff
@@ -227,13 +231,26 @@ export class TxRunnerWeb3 {
     })
   }
 
-  async sendUserOp (tx) {
+  async sendUserOp (tx, chainId) {
+    // AA03: Add network name and public URL to support contract transactions using smart account
+    const aaSupportedNetworks = {
```
yann300, line 236: these constants should be put in an helper, so you don't duplicate it.

libs/remix-lib/src/execution/txRunnerWeb3.ts
Aniket-Engg: updated

libs/remix-lib/src/execution/txRunnerWeb3.ts
Aniket-Engg: done

yann300 (approved):

PR #6017, opened 2025-05-12
fixes compiler config file overriding

Aniket-Engg:
If `Compiler configuration` option is selected, compile config file is overwritten


yann300:
The issue persist while reloading Remix.
- enable `use config file`
- update the content of the file
- reload
- the file is overwritten

yann300 (approved):

PR #6026, opened 2025-05-15
do not unfold .deps folder while compiling and focus to the element w…

yann300:
- Do not unfold .deps folder while compiling.
 - Focus to the element when selected.


joeizang:
It works fine for me. Maybe when e2e's are fixed then I can approve.

joeizang (approved):
Behaviour LGTM

PR #6029, opened 2025-05-17
config file compiler

bunsenstraat:
The compiler state doesn't accurately represent the actual state, as the parameters are set by the compiler only when the user compiles. This fix allows the code parser to respond correctly to the changes in configuration and also adds the custom config file. It also responds to changes in that file. 

Can be tested by cloning remix-reward. the main file should show yellow lines when optimization is off. Now it will react properly to changes in the UI and also to changing custom config files.


apps/remix-ide/src/app/plugins/parser/code-parser.tsx
```diff
@@ -167,6 +167,24 @@ export class CodeParser extends Plugin {
       await this.handleChangeEvents()
     })
 
+    this.on('solidity', 'compilerQueryParamsUpdated', async (file) => {
```
yann300, line 170: This event doesn't have `file` as parameter. (I think same for the events below.

yann300 (approved):

PR #6060, opened 2025-05-26
Matomo & AI consent dialog

### LINKED ISSUE #5832: Add AI-copilot acceptance criteria to Matomo dialog
Rename Title
We should have two switches
- regular matomo data
- AI copilot requests (you can switch it on and off later from settings

Accept selected
Accept all
Decline all

Aniket-Engg:
fixes #5832

### DIFF
--- a/apps/remix-ide-e2e/src/tests/matomo.test.ts
+++ b/apps/remix-ide-e2e/src/tests/matomo.test.ts
@@ -110,6 +110,8 @@ module.exports = {
     },
     'change settings #group2': function (browser: NightwatchBrowser) {
         browser
+            .waitForElementVisible('*[id="remixTourSkipbtn"]')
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/tabs/settings-tab.tsx
```diff
@@ -36,6 +36,7 @@ export default class SettingsTab extends ViewPlugin {
   }
   element: HTMLDivElement
   public useMatomoAnalytics: any
+  public useMatomoPerfAnalytics: any
```
joeizang: Is there any way we can add basic types and avoid any going forward? Any should be reserved for completely unresolvable situations.

Aniket-Engg: updated

PR #6083, opened 2025-06-05
AI copilot sync fixed

Aniket-Engg:
When clicked in status bar, copilot is not toggled for editor. This PR fixes that

### DIFF
--- a/libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
+++ b/libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
@@ -86,7 +86,7 @@ export const TabsUI = (props: TabsUIProps) => {
       })
     }
   }, [tabsState.selectedIndex])
-  // Toggle the copilot in editor when clicked to update in status bar 
+  // Toggle the copilot in editor when clicked to update in status bar
   useEffect(() => {
     const run = async () => {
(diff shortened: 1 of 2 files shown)


libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
```diff
@@ -86,6 +86,15 @@ export const TabsUI = (props: TabsUIProps) => {
...
     }
   }, [tabsState.selectedIndex])
+  // Toggle the copilot in editor when clicked to update in status bar
+  useEffect(() => {
+    const run = async () => {
+      props.plugin.on('settings', 'copilotChoiceUpdated', async (isChecked) => {
+        setAI_switch(isChecked)
+      })
+    }
+    if (tabsState.currentExt === 'sol') run()
```
joeizang, line 96: Does this mean remix ai is enabled for solidity alone?

joeizang (approved):
LGTM

libs/remix-ui/tabs/src/lib/remix-ui-tabs.tsx
Aniket-Engg: Yes, I think it is same on live

Issue #6041, opened 2025-05-21
Loading code from URL improvement


yann300:
- When loading the code using the path, the other params are not forwarded. This should be the case.
- We should load the code to the root instead of under a sub folder (bc this makes the compilation sometimes  to fail).

PR #6093, opened 2025-06-17
Fix loading contracts from url

yann300:
fix https://github.com/ethereum/remix-project/issues/6041

- always save artifact in the root folder.
- fix remappings when compiling a contract from the `.debug` folder.
- pass the endpoint URL to the  etherscan resolver.

### DIFF
--- a/libs/remix-core-plugin/src/lib/compiler-fetch-and-compile.ts
+++ b/libs/remix-core-plugin/src/lib/compiler-fetch-and-compile.ts
@@ -176,6 +176,21 @@ export class FetchAndCompile extends Plugin {
       }
     }
     const { config, compilationTargets, version } = data
+    /*
+    * If the remappings are defined in the config, we need to update them to point to the targetPath
+    * it's beeing disabled for the moment.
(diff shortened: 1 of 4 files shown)


libs/remix-core-plugin/src/lib/helpers/fetch-blockscout.ts
```diff
@@ -21,7 +21,11 @@ export const fetchContractFromBlockscout = async (plugin, endpoint, contractAddr
     return null
   }
 
-  if (data.result[0].FileName === '') {
+  if (data.result[0].FileName === '.sol' && data.result[0].ContractName) {
```
ioedeveloper, line 24: Is this an error `data.result[0].FileName === '.sol'`? or do you meant `data.result[0].FileName.endsWith('.sol')`

yann300: This is intended, this API isn't really consistent and in my test case the filename returned by the API ia `.sol`

PR #6109, opened 2025-06-24
Fix vyper compilation from play button

### LINKED ISSUE #6104: on alpha: Vyper play button to compile not working
**Describe the bug**
Clicking the play button to compile a Vyper contract does not work - it makes a Network error.  Clicking the compile button does work.

Image

joeizang:
fixes #6104


ryestew (approved):
It works for me.

apps/vyper/src/app/utils/compiler.tsx
```diff
@@ -82,8 +82,12 @@ export async function compile(url: string, contract: Contract): Promise<VyperCom
       [contractName] : { content : contract.content }
     }
   }
-console.log('compilePackage', `${url}compile`)
-  let response = await axios.post(`${url}compile`, compilePackage )
+  let response
+  if (url.endsWith('/')) {
```
yann300: why are you checking the URL here an not below?

PR #6124, opened 2025-06-29
Fix broken link to Ethereum Contract ABI specification

MozirDmitriy:
Replaced the outdated link to the Ethereum Contract ABI specification (previously pointing to the old GitHub wiki) with the current official Solidity documentation link: https://docs.soliditylang.org/en/latest/abi-spec.html. This ensures that users and developers are directed to the most up-to-date and authoritative resource for ABI details.


Aniket-Engg:
As per the guidelines, Only updating a comment doesn't count as contribution.

PR #6141, opened 2025-07-07
Show network name and ticker

Aniket-Engg:


apps/remix-ide/src/blockchain/execution-context.js
```diff
@@ -88,24 +89,36 @@ export class ExecutionContext {
           else if (id === 5) name = 'Goerli'
           else if (id === 42) name = 'Kovan'
           else if (id === 11155111) name = 'Sepolia'
-          else name = 'Custom'
-  
+          else {
+            const response = await fetch('https://chainid.network/chains.json')
```
yann300: Would it make sense to store this in the local storage? And if the current chaind isn't in the stored artefact to fetch it again.

Aniket-Engg: done

apps/remix-ide/src/blockchain/execution-context.js
```diff
@@ -78,34 +78,51 @@ export class ExecutionContext {
...
+          else {
+            let networkDetails = localStorage.getItem('networkDetails')
+            if (!networkDetails) networkDetails = '{}'
+            networkDetails = JSON.parse(networkDetails)
+            if (networkDetails[id]) {
+              name = networkDetails[id].name
+              networkNativeCurrency = networkDetails[id].nativeCurrency
+            } else {
+              const response = await fetch('https://chainid.network/chains.json')
+              if (response.ok) {
```
yann300, line 97: Is the case where the nativeCurrency can't be resolved is handled? like defaulting to eth or something?

Aniket-Engg: that is done here: https://github.com/ethereum/remix-project/pull/6141/files#diff-edc29c7646590f31b7818ef955996b84c9d1b0ca9c874b03b070244bee9cc8cbR108

Aniket-Engg: if you want, I can do the same in this file too

yann300: I am saying that because line 83  the local variable is set to null.

Aniket-Engg: ok, I will update that here too

Aniket-Engg: done

PR #6144, opened 2025-07-09
Remix AI Assistant UI Patches

joeizang:


STetsing (changes requested):
@joeizang on the AI info tooltip, rather use this text

- Use `@workspace <text>` or `/w <text>` to manage or edit files within your workspace
- Use `@generate <text>` or `/g <text>`to generate contracts and scripts in a new workspace. In case the generation result does not compile, use the `/continue` or `/c` command to generate similar solutions until a compilable solution is suggested.
- Alternatively, you may type your question directly below

PR #6164, opened 2025-07-15
Ai local imports

STetsing:
- remove the `/c` or `/continue` command
- Adding first level imported libs in the compilation error prompt
- Added the local import when explaining an error with remix ai


libs/remix-ui/solidity-compiler/src/lib/api/compiler-api.ts
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { compile, helper, Source, CompilerInputOptions, compilerInputFactory, CompilerInput } from '@remix-project/remix-solidity'
 import { CompileTabLogic, parseContracts } from '@remix-ui/solidity-compiler' // eslint-disable-line
 import type { ConfigurationSettings, iSolJsonBinData } from '@remix-project/remix-lib'
+import { emit } from 'node:process';
```
yann300: this shouldn't be here.

yann300 (approved):

PR #6173, opened 2025-07-17
Add editor UI with compile/run actions, and AI assistant controls

hsy822:


libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
```diff
@@ -10,6 +10,7 @@ import * as packageJson from '../../../../../package.json'
 import './css/style.css'
 import { iSolJsonBinData, iSolJsonBinDataBuild } from '@remix-project/remix-lib'
 import { appPlatformTypes, platformContext } from '@remix-ui/app'
+import { CompileDropdown } from '../../../tabs/src/lib/components/CompileDropdown'
```
yann300, line 13: this should be imported using the `'@remix-ui/...` syntax.

PR #6175, opened 2025-07-17
Add editor UI with compile/run actions, and AI assistant controls

hsy822:


libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
```diff
@@ -10,6 +10,7 @@ import * as packageJson from '../../../../../package.json'
 import './css/style.css'
 import { iSolJsonBinData, iSolJsonBinDataBuild } from '@remix-project/remix-lib'
 import { appPlatformTypes, platformContext } from '@remix-ui/app'
+import { CompileDropdown } from '../../../tabs/src/lib/components/CompileDropdown'
```
yann300: you don't seem to use this ref in this file.

libs/remix-ui/tabs/src/lib/components/CompileDropdown.tsx
```diff
@@ -0,0 +1,222 @@
...
+    _paq.push(['trackEvent', 'solidityCompiler', 'solidityScan', 'initiateScan'])
+
+    const workspace = await plugin.call('filePanel', 'getCurrentWorkspace')
+    const fileName = `${workspace.name}/${compiledFileName}`
+    const filePath = `.workspaces/${compiledFileName}`
+    const file = await plugin.call('fileManager', 'readFile', filePath)
+
+    const urlResponse = await axios.post(`${endpointUrls.solidityScan}/uploadFile`, { file, fileName })
+
+    if (urlResponse.data.status === 'success') {
```
yann300: Is this code has been copied over from the solidity tab? If possible it's better to have a helper function for this.

yann300:
The check mark left to the compile appear before the compilation actually finishes..

libs/remix-ui/solidity-compiler/src/lib/solidity-compiler.tsx
hsy822: Resolved. Removed the unused CompileDropdown import.

libs/remix-ui/tabs/src/lib/components/CompileDropdown.tsx
hsy822: Extracted the duplicated Solidity Scan API call and WebSocket logic into a new helper function.

Issue #6234, opened 2025-08-06
Fixes to Bootstrap 5 Dark Theme


joeizang:
- [ ] Borders and horizontal line colors on most form elements shouldn't be white or light gray. See screenshots and images below. 

<img width="815" height="611" alt="Image" src="https://github.com/user-attachments/assets/9f89ee32-0bf0-4d7c-8c5a-b1299c399bc8" /> **OLD BORDERS AND HORIZONTAL LINES** This is what it looks like now.

<img width="298" height="514" alt="Image" src="https://github.com/user-attachments/assets/134b2be4-9c53-4b03-8e46-6476e170cd18" />

<img width="297" height="131" alt="Image" src="https://github.com/user-attachments/assets/6185f410-1579-4ada-ae96-9c12d7181818" />

- [ ] Contrast on badges should be improved. White on a light colored orange isn't the best in terms of contract and readability. See a comparison below 
**NEW LOOK WHICH SHOULD BE IMPROVED**

<img width="135" height="96" alt="Image" src="https://github.com/user-attachments/assets/892c8680-77e4-4a79-8aa4-bc97796df773" />

OLD IMAGE WHICH HAS CONTRAST THAT IS MUCH BETTER THAN IT IS NOW ON BS 5.

<img width="138" height="45" alt="Image" src="https://github.com/user-attachments/assets/f26dc70f-8e3e-4b20-bb51-daf8c3f893b5" />

- [ ] Contrast of primary color can be improved significantly. Same point as above but just for buttons. See images below

OLD BLACK THEME WHICH HAS BETTER CONTRAST TO WHAT IS IN THE BS 5 THEME NOW.

<img width="272" height="90" alt="Image" src="https://github.com/user-attachments/assets/4e7bbed3-430c-4615-8a4b-49614246f40c" />

<img width="125" height="111" alt="Image" src="https://github.com/user-attachments/assets/0621f101-14bd-4af7-b6c6-e0ec6a97a0f9" />

<img width="124" height="82" alt="Image" src="https://github.com/user-attachments/assets/fcc197e5-ec1b-41e9-9238-45da0b7dba5d" />

NEW BS 5 THEME THAT WILL NEED CONTRAST IMPROVEMENT

<img width="236" height="54" alt="Image" src="https://github.com/user-attachments/assets/0356a2dd-8c60-4539-b29e-9a9787b3bcc4" />

[...]

Issue #6230, opened 2025-08-05
Bootstrap 5 improvements


Aniket-Engg:
Improvements required from https://github.com/ethereum/remix-project/pull/6134

- [ ] Update Light theme colors as per discussion

Aniket-Engg:
Plugin manager is getting revamped through https://github.com/ethereum/remix-project/pull/6224

joeizang:
- The border above the terminal is the wrong type of gray.
- Text in the template explorer in not the correct size. See the Image below:
This is on live
<img width="493" height="390" alt="Screenshot 2025-08-11 at 11 12 56" src="https://github.com/user-attachments/assets/168814b2-0707-4fec-b54c-76fdd96b50c2" />

Just for comparison, here is the current image with BS 5:

<img width="474" height="395" alt="Screenshot 2025-08-11 at 11 13 09" src="https://github.com/user-attachments/assets/379df251-81e6-4b1a-b306-dd763b76c09e" />

- There should be some margins at the top and bottom of the search of the terminal. See Image:

<img width="316" height="72" alt="Screenshot 2025-08-11 at 11 15 09" src="https://github.com/user-attachments/assets/34e8b37a-0d93-4e75-9a07-490af9a9904b" />

- Some elements in template explorer are completely misaligned. See image:

<img width="253" height="206" alt="Screenshot 2025-08-11 at 11 11 13" src="https://github.com/user-attachments/assets/b6d84c38-ce85-4ecd-b86b-08c8ed23fccd" />

- Text of our buttons are of different colors which isn't right. See image:

<img width="130" height="47" alt="Screenshot 2025-08-11 at 09 34 22" src="https://github.com/user-attachments/assets/3135b0f0-eb5c-442f-9723-556c21f32bfa" />

<img width="254" height="77" alt="Screenshot 2025-08-11 at 09 34 58" src="https://github.com/user-attachments/assets/c6414d88-e44d-444f-aab2-751c2abf6fdd" />

- I am not sure about the dark text on a light button as it isn't consistent with our button text in the ide.

PR #6249, opened 2025-08-12
Bs5 dark theme fixes 2

Aniket-Engg:
fixes https://github.com/ethereum/remix-project/issues/6234
fixes https://github.com/ethereum/remix-project/issues/6230#issuecomment-3175203358


joeizang:
- Margins on the outside are still very small. From the image, you see that the input field is sitting right on the terminal.
image

- This caret should have the right color, see the image below:
image

- Border is the wrong color.
image

- The borders and button text colors need to be fixed.
image

- Margins between elements is off.
image

- Some buttons don't have the right text color:
image

- Width adjustment seems to be broken now. This is not the case on remix.ethereum.org
image

This is on the preview link:
image

- Error cards seem to be the wrong size

Screenshot 2025-08-12 at 15 51 50

joeizang (approved):

PR #6268, opened 2025-08-19
fix compile button

### LINKED ISSUE #6256: Editor Compile button improvements
- [x] Once a contract is compiled and changes are made to file, button text still stuck on `Compiled`
- [x] Submenu options usually work on successful compilation. If contract is having compilation errors, using any submenu like `Compile and ****` does nothing
- [x] After loading Remix, if a contract with errors is compiled, button text stuck on `Compiling....`

hsy822:
fixes #6256

### DIFF
--- a/apps/remix-ide-e2e/src/tests/compile_run_widget.test.ts
+++ b/apps/remix-ide-e2e/src/tests/compile_run_widget.test.ts
@@ -12,7 +12,7 @@ module.exports = {
     return []
   },
 
-  'Compile using the widget #group1': function (browser: NightwatchBrowser) {
+  'Compile using the widget #group1 #pr': function (browser: NightwatchBrowser) {
     browser
       .openFile('contracts/3_Ballot.sol')
       .click('[data-id="compile-action"]')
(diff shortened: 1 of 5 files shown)


Aniket-Engg (commented):
- If there are errors during compilation, clicking on `Open compiler configuration` does nothing
- If user clicks on `Compile` button and compilation fails, show no toaster but open Solidity Compiler plugin
- Clicking on `Compile and Run script` and compilation fails, should show toaster like `Compilation failed, skipping 'Run Script'` and open Solidity Compiler plugin . This should be the case for all `Compile and *****` options . All these toasters like `Could not find the publish button.` in case of failing compilation, should be handled accordingly. Toasters should be used minimal.
- For successful compilation, clicking on `Compile and run script` runs script twice
- Even if Solidity compiler is already in focus, clicking on `Run and Publish` items shows toaster `Switching to Solidity Compiler to publish...`
- Remove logs from browser console like `Ran Remix Solidity Compiler`, `Loaded 4 script files` etc.

PR #6283, opened 2025-08-22
fix plugin activation on home tab

Aniket-Engg:
Clicking on `Start Learning` for learneth & `Search for contracts` for cookbook doesn't activate plugin but shows error in browser console


joeizang (approved):
LGTM

PR #6320, opened 2025-08-29
Close Pinned Plugin Feature

### LINKED ISSUE #6305: remixAI Assistant removal
**Describe the bug**

There is no way to remove the `remixAI assistant` window. Please fix that. I should be able to close like github copilot.

joeizang:
fixes #6305 290825

### DIFF
--- a/libs/remix-ui/top-bar/src/lib/remix-ui-topbar.tsx
+++ b/libs/remix-ui/top-bar/src/lib/remix-ui-topbar.tsx
@@ -129,6 +129,27 @@ export function RemixUiTopbar () {
     }
   }, [global.fs.browser.workspaces, global.fs.browser.workspaces.length])
 
+  useEffect(() => {
+    plugin.on('theme', 'themeChanged', (theme) => {
+      setCurrentTheme(theme)
+    })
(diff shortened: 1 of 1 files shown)


joeizang:
> * ere are errors in browser console like `Invalid DOM property `fill-rule`. Did you mean `fillRule`?` , `Invalid DOM property `clip-rule`. Did you mean `clipRule`?` and `[TIMEOUT] Timeout for call pinView from sidePanel`

For point 1, I fixed the regression that was causing the console errors. The last point you made is also addressed by fixing this regression.

The plugin will still show on the right because it is still pinned. when you click the restore icon, it restores it back as pinned. We agreed on the last team call not to close or unpin any pinned plugin that gets closed. 

I moved the restore icon to the topbar because the tabs bar has very precise settings that would not be fixed easily.

Aniket-Engg (commented):
- On Solidity Compiler left & right, In browser console, I see: `compiler-container.tsx:380 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'selectorList')
    at compiler-container.tsx:380:49` That is not a case on live

PR #6331, opened 2025-09-02
Loginbutton

bunsenstraat:


joeizang:
- The login button looks like its disabled. See the image below
image

- Also the background of the button should not be this color. The previous color seems a better fit.

Everything else looks good to me.

joeizang (dismissed):
Left some comments

Aniket-Engg (approved):

PR #6349, opened 2025-09-05
fix path traversal in verification plugin

### LINKED ISSUE #6107: Path Travel Issue in Verified Contract Fetch Process in Contract Verification Tool
### Introduction
A path traversal Issue (../ file path issue) has been discovered in Remix Contract Verification Tool. This Issue allows attackers to shift verification paths to upper directories, thereby hiding specific real contract source files and modify the other files in remix. This results in a source code forgery Issue potentially enabling attackers to deceive users and auditors during contract verification and may impact the developer who use remix to develop their contract when they use Contract Verification Tool.
By constructing a specially crafted JSON file for source code verification, an attacker can exploit flaws in Remix's directory handling mechanism. This causes the actual source file `../../a.sol` to be overwritten and hidden while displaying the fake malicious source files in the directory. and other bad files would overwrite the users' source code in remix

## Example JSON
Below is an example JSON file demonstrating the Issue:

```json
{
    "language": "Solidity",
    "sources": {
        "a.sol": {
            "content": "// SPDX-License-Identifier: GPL-3.0

            pragma solidity >=0.7.0 <0.9.0;

            contract Owner {
                address private Hacker;
                function bad(address newOwner) public {
                    Hacker = address(0x0);
                }
            }"
        },
        "../../a.sol": {
            "content": "// SPDX-License-Identifier: GPL-3.0

            pragma solidity >=0.7.0 <0.9.0;

            contract Owner {
                address private owner;
                function changeOwner(address newOwner) public {
                    owner = newOwner;
                }
            }"
        }
    },
    "settings": {}
}
```

By submitting this crafted JSON file, the actual source file `../../a.sol` is concealed, while the fake file `a.sol` with malicious code is displayed in the directory. in the mean time, the other attack code would also overwrited the other file in remix.

Below are links illustrating the exploitation:

https://sepolia.etherscan.deth.net/address/0xb7e278f0116508c611f8a486666d3823e51c44dc

### Recommendations
* Implement strict path sanitization to prevent the use of `../` for directory traversal.

We also find same issue in [Forge](https://github.com/foundry-rs/foundry/issues/4896), Metasuite, Blockscan and [crytic-based ](https://github.com/crytic/crytic-compile/pull/425)tools like Slither. so this is a common issue for code fetcher tool but we also think this is a risky issue need to did some better pratice.

Thank you very much for providing such a powerful tool, it's a great convenience for us and hopefully this discovery will allow me to contribute something to the tool.

Aniket-Engg:
fixes #6107

### DIFF
--- a/apps/contract-verification/src/app/Verifiers/BlockscoutVerifier.ts
+++ b/apps/contract-verification/src/app/Verifiers/BlockscoutVerifier.ts
@@ -45,7 +45,7 @@ export class BlockscoutVerifier extends EtherscanVerifier {
     result.push({ content: blockscoutSource.SourceCode, path: targetFilePath })
 
     for (const additional of blockscoutSource.AdditionalSources ?? []) {
-      if(!additional.Filename.startsWith('..')) result.push({ content: additional.SourceCode, path: `${filePrefix}/${additional.Filename}` })
+      if (!additional.Filename.startsWith('..')) result.push({ content: additional.SourceCode, path: `${filePrefix}/${additional.Filename}` })
     }
 
     return { sourceFiles: result, targetFilePath }


apps/contract-verification/src/app/Verifiers/BlockscoutVerifier.ts
```diff
@@ -38,11 +38,11 @@ export class BlockscoutVerifier extends EtherscanVerifier {
     const result: SourceFile[] = []
     const filePrefix = `/${this.LOOKUP_STORE_DIR}/${chainId}/${contractAddress}`
 
-    const targetFilePath = `${filePrefix}/${blockscoutSource.FileName}`
+    const targetFilePath = `${filePrefix}/${blockscoutSource.FileName.startsWith('..') ? blockscoutSource.FileName.replace(/..\//g, '') : blockscoutSource.FileName}`
```
yann300: why ae you handling this part differently then for the other contract verifers?

apps/contract-verification/src/app/Verifiers/BlockscoutVerifier.ts
```diff
@@ -38,11 +38,12 @@ export class BlockscoutVerifier extends EtherscanVerifier {
     const result: SourceFile[] = []
     const filePrefix = `/${this.LOOKUP_STORE_DIR}/${chainId}/${contractAddress}`
 
-    const targetFilePath = `${filePrefix}/${blockscoutSource.FileName}`
+    let targetFilePath
+    if (!blockscoutSource.FileName.startsWith('..')) targetFilePath = `${filePrefix}/${blockscoutSource.FileName}`
```
yann300, line 42: maybe in that case we should use a default name.

Aniket-Engg: ok

Aniket-Engg: done

PR #6358, opened 2025-09-09
update questions

ryestew:


libs/remix-ui/remix-ai-assistant/src/lib/conversationStarters.ts
Aniket-Engg, line 1: It should be 'work with'

Aniket-Engg, line 1: Not sure if this is an expert level question

PR #6366, opened 2025-09-10
fix matomo remixassistant activities

### LINKED ISSUE #6362: Clean Matomo data.
The `handleActivity` has some issues:
 - it has only 2 parameters. Usually matomo events have 3 params. This can be fixed by just having `type` and `payload` as specific params.
 - we should remove the `promptSend` and `typing`.

STetsing:
fix #6362

### DIFF
--- a/apps/remix-ide/src/app/plugins/remix-ai-assistant.tsx
+++ b/apps/remix-ide/src/app/plugins/remix-ai-assistant.tsx
@@ -101,7 +101,8 @@ export class RemixAIAssistant extends ViewPlugin {
   }
 
   async handleActivity(type: string, payload: any) {
-    (window as any)._paq?.push(['trackEvent', 'remixai-assistant', `${type}-${payload}`])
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/plugins/remix-ai-assistant.tsx
```diff
@@ -101,7 +101,8 @@ export class RemixAIAssistant extends ViewPlugin {
   }
 
   async handleActivity(type: string, payload: any) {
-    (window as any)._paq?.push(['trackEvent', 'remixai-assistant', `${type}-${payload}`])
+    if ((type === 'promptSend') || (type === 'typing')) return
```
yann300: why don't you just remove the actual calls?

yann300: @STetsing I meant that you can remove the place where those calls are made.

yann300 (approved):

PR #6376, opened 2025-09-15
Add matomo keys

yann300:
Added keys for logging transaction hash and network id.


apps/remix-ide/src/blockchain/blockchain.tsx
```diff
@@ -797,8 +797,29 @@ export class Blockchain extends Plugin {
       (_) => this.executionContext.currentblockGasLimit()
     )
 
+    const logTransaction = (txhash, origin) => {
+      this.detectNetwork((error, network) => {
+        console.log(`transaction sent: ${txhash}`, network)
```
ioedeveloper, line 802: Do you want to keep this console log?

PR #6377, opened 2025-09-15
disable matomo necessary in settings

Aniket-Engg:
- Matomo necessary toggle is on and disabled. So it cannot be set to off position.
- Matomo with cookies option calls the method to disable cookies when disabled

### DIFF
--- a/libs/remix-ui/settings/src/lib/settings-section.tsx
+++ b/libs/remix-ui/settings/src/lib/settings-section.tsx
@@ -39,10 +39,8 @@ export const SettingsSectionUI: React.FC<SettingsSectionUIProps> = ({ plugin, se
   const handleToggle = (name: string) => {
     if (state[name]) {
       const newValue = !state[name].value
-
       dispatch({ type: 'SET_LOADING', payload: { name: name } })
       dispatch({ type: 'SET_VALUE', payload: { name: name, value: newValue } })
-      // _paq.push(['disableCookies'])
       if (!newValue && formUIData[name]) {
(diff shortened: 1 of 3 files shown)


libs/remix-ui/settings/src/lib/settings-section.tsx
```diff
@@ -39,17 +39,16 @@ export const SettingsSectionUI: React.FC<SettingsSectionUIProps> = ({ plugin, se
...
       dispatch({ type: 'SET_VALUE', payload: { name: name, value: newValue } })
-      // _paq.push(['disableCookies'])
       if (!newValue && formUIData[name]) {
         Object.keys(formUIData[name]).forEach((key) => {
           dispatch({ type: 'SET_VALUE', payload: { name: key, value: '' } })
         })
         dispatch({ type: 'SET_TOAST_MESSAGE', payload: { value: 'Credentials removed' } })
       }
       if (name === 'copilot/suggest/activate') plugin.emit('copilotChoiceUpdated', newValue)
+      if (name === 'matomo-perf-analytics') plugin.call('settings', 'updateMatomoPerfAnalyticsChoice', newValue)
```
yann300, line 51: maybe this should also emit an event, same as above.

Aniket-Engg: that is not required here as there is no such event and it is not being listened anywhere

yann300: it is just to be consistent with the line above `plugin.emit('copilotChoiceUpdated', newValue)`

PR #6390, opened 2025-09-19
Use textdecoder decoding code param payload

CoveMB:
**Title**
- Update UTF-8 base64 decoding for `code` URL param to use TextDecoder; add e2e coverage

**Summary**
- Replaces ad‑hoc percent-escape base64 decoding with a standards-based TextDecoder approach when loading code from the `?code=` URL parameter.
- Adds an e2e test that covers ASCII punctuation, URL-encoded characters, and Unicode (including emoji)

**Problem**
- Loading code from the `code` URL param could fail to correctly decode content containing:
  - ASCII punctuation and special URL-encoded characters.
- The prior implementation constructed a percent-escaped string from `atob()` bytes and then used `decodeURIComponent`, which is error-prone for arbitrary UTF‑8.

**Scope**
- `libs/remix-ui/workspace/src/lib/actions/workspace.ts`: Replace decoder implementation and update usage.
- `apps/remix-ide-e2e/src/tests/url.test.ts`: Add a new test covering special chars, punctuation, and persistence behavior.

**Changes**
- Rename and replace decoder:
  - Old: `decodePercentEscapedBase64(b64Payload)`
  - New: `decodeBase64(b64Payload)` using `TextDecoder('utf-8')`
- Updated code path where `params.code` is processed to use the new decoder.
- Added e2e test “Should load the code from URL & code params with ASCII punctua
tion and special character #group1” that:
  - Visits Remix with a `#code=` payload containing special characters and emoji
.
  - Ensures the decoded content matches expected substrings.
  - Verifies “code-sample” workspace is not persisted after returning to the app
 root.

**Implementation Details**
- Decoding flow now:
  - `decodeURIComponent(b64Payload)` → `atob()` → `Uint8Array.from()` → `new Tex
tDecoder().decode(bytes)`.
- This follows a robust, standards-based UTF‑8 decode path and avoids manual per
cent-encoding reconstruction.

**Backward Compatibility**
- No user-facing breaking changes expected.
- Internal function rename from `decodePercentEscapedBase64` to `decodeBase64` i
s contained within the same module with its usage updated. No external API chang
es anticipated.

**Note**
- The added e2e test was tested on it's own but the test suite is currently disabled
- Wizard side MR: https://github.com/OpenZeppelin/contracts-wizard/pull/665

### DIFF
--- a/apps/remix-ide-e2e/src/tests/url.test.ts
+++ b/apps/remix-ide-e2e/src/tests/url.test.ts
@@ -201,6 +201,42 @@ module.exports = {
       })
   },
 
+  'Should load the code from URL & code params with ASCII punctuation and special character #group1': function (browser: NightwatchBrowser) {
+    browser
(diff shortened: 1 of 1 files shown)


CoveMB:
Hello, let me know if the MR need anything else @Aniket-Engg

bedroomeyes555 (approved):

PR #6394, opened 2025-09-22
Add deploy and verify contract

### LINKED ISSUE #4490: Put verification button in Deploy & Run
There should be a button to verify a contract in Deploy & Run.

hsy822:
### **Changes**

* Added a **"Verify Contract on Explorers"** checkbox to the **Deploy & Run Transactions** tab.
* Removed the "Publish to IPFS" option from the UI; it is now integrated into the "Verify" feature and runs automatically.
* **Automatically selects only the supported verifiers** (e.g., Sourcify, Etherscan) for the current network, based on its entry in `https://chainid.network/chains.json`.
* **Note: For Etherscan verification, an API key must be pre-configured in the global Remix Settings tab.**
* The results of verification attempts (success, failure, pending) are saved as receipts to `localStorage`. This allows the feature to share state with the Contract Verification plugin and display the results in its **Receipts** tab.

### **How to Test**

1.  (Optional) Navigate to the `Settings` tab and add an Etherscan API key.
2.  Navigate to the `Deploy & Run Transactions` tab.
3.  Connect to an official test network, such as Sepolia.
4.  Select a compiled contract and check the new **"Verify Contract on Explorers"** checkbox.
5.  Click the **Deploy** button.
6.  In the terminal, confirm that after the successful deployment log, verification logs for each **supported** explorer are printed sequentially.
7.  Navigate to the `Contract Verification` plugin and click on the **Receipts** tab. Confirm that the new auto-verification record appears at the top of the list.

### **Related Issues**

* Fixes #4490

### DIFF
--- a/apps/contract-verification/src/app/ContractVerificationPluginClient.ts
+++ b/apps/contract-verification/src/app/ContractVerificationPluginClient.ts
@@ -119,32 +119,34 @@ export class ContractVerificationPluginClient extends PluginClient {
       const compilerAbstract: CompilerAbstract = compilationResult
       const chainSettings = mergeChainSettingsWithDefaults(chainId, userSettings)
 
+      const verificationPromises = []
+
       if (validConfiguration(chainSettings, 'Sourcify')) {
-        await this._verifyWithProvider('Sourcify', submittedContract, compilerAbstract, chainId, chainSettings)
+        verificationPromises.push(this._verifyWithProvider('Sourcify', submittedContract, compilerAbstract, chainId, chainSettings))
       }
 
       if (currentChain.explorers && currentChain.explorers.some(explorer => explorer.name.toLowerCase().includes('routescan'))) {
-        await this._verifyWithProvider('Routescan', submittedContract, compilerAbstract, chainId, chainSettings)
+        verificationPromises.push(this._verifyWithProvider('Routescan', submittedContract, compilerAbstract, chainId, chainSettings))
       }
 
-      if (currentChain.explorers && currentChain.explorers.some(explorer => explorer.name.toLowerCase().includes('blockscout'))) {
-        await this._verifyWithProvider('Blockscout', submittedContract, compilerAbstract, chainId, chainSettings)
+      if (currentChain.explorers && currentChain.explorers.some(explorer => explorer.url.includes('blockscout'))) {
+        verificationPromises.push(this._verifyWithProvider('Blockscout', submittedContract, compilerAbstract, chainId, chainSettings))
       }
 
-      if (currentChain.explorers && currentChain.explorers.some(explorer => explorer.name.toLowerCase().includes('etherscan'))) {
+      if (currentChain.explorers && currentChain.explorers.some(explorer => explorer.name.includes('etherscan'))) {
         if (etherscanApiKey) {
-          if (!chainSettings.verifiers.Etherscan) chainSettings.verifiers.Etherscan = {}
-          chainSettings.verifiers.Etherscan.apiKey = etherscanApiKey
-          await this._verifyWithProvider('Etherscan', submittedContract, compilerAbstract, chainId, chainSettings)
+          verificationPromises.push(this._verifyWithProvider('Etherscan', submittedContract, compilerAbstract, chainId, chainSettings))
         } else {
           await this.call('terminal', 'log', { type: 'warn', value: 'Etherscan verification skipped: API key not found in global Settings.' })
         }
       }
 
-      submittedContracts[submittedContract.id] = submittedContract
+      await Promise.all(verificationPromises)
 
+      submittedContracts[submittedContract.id] = submittedContract
       window.localStorage.setItem('contract-verification:submitted-contracts', JSON.stringify(submittedContracts))
       this.internalEvents.emit('submissionUpdated')
+
     } catch (error) {
       await this.call('terminal', 'log', { type: 'error', value: `An unexpected error occurred during verification: ${error.message}` })
     }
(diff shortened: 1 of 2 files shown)


apps/contract-verification/src/app/ContractVerificationPluginClient.ts
```diff
@@ -62,8 +64,91 @@ export class ContractVerificationPluginClient extends PluginClient {
...
+
+      const submittedContract: SubmittedContract = {
+        id: `${chainId}-${contractAddress}`, 
+        address: contractAddress,
+        chainId: chainId,
+        filePath: Object.keys(compilationResult.data.contracts).find(path => path in compilationResult.source.sources),
+        contractName: contractName,
+        abiEncodedConstructorArgs: constructorArgs,
+        date: new Date().toISOString(),
+        receipts: []
```
manuelwedler, line 116: It would be nice to have receipts for submittedContract and store it in local storage, such that it also shows up in the plugin's UI. See the `VerifyView` for reference. 

I also thought of refactoring the `VerifyView` once, since a lot of the logic from there could actually be reused for this feature.

Anyway, very nice that you tackled this feature!

libs/remix-ui/run-tab/src/lib/actions/deploy.ts
```diff
@@ -173,22 +174,71 @@ export const createInstance = async (
...
     plugin.compilersArtefacts.addResolvedContract(addressToString(address), data)
-    if (plugin.REACT_API.ipfsChecked) {
-      _paq.push(['trackEvent', 'udapp', 'DeployAndPublish', plugin.REACT_API.networkName])
-      publishToStorage('ipfs', selectedContract)
+
+    if (isVerifyChecked) {
+      _paq.push(['trackEvent', 'udapp', 'DeployAndVerify', plugin.REACT_API.networkName])
+
+      try {
+        await publishToStorage('ipfs', selectedContract)
```
yann300: Are we keeping this in order to make sure sourcify verifies it? If we can use the sourcify api to verify it that would be better, we won't have to rely on a IPFS endpoint.

libs/remix-ui/run-tab/src/lib/actions/deploy.ts
```diff
@@ -173,22 +174,71 @@ export const createInstance = async (
...
+
     addInstance(dispatch, { contractData: contractObject, address, name: contractObject.name })
     const data = await plugin.compilersArtefacts.getCompilerAbstract(contractObject.contract.file)
-
     plugin.compilersArtefacts.addResolvedContract(addressToString(address), data)
-    if (plugin.REACT_API.ipfsChecked) {
-      _paq.push(['trackEvent', 'udapp', 'DeployAndPublish', plugin.REACT_API.networkName])
-      publishToStorage('ipfs', selectedContract)
+
+    if (isVerifyChecked) {
```
yann300, line 184: In the UI, that should be checked by default.

yann300: But probably the setting should be remenbered when the page is loaded.

yann300:
If the network isn't supported we shouldn't log anything and best to disable the checkbox or maybe that's best to just display the checkbox only if the network is actually supported.

libs/remix-ui/run-tab/src/lib/actions/deploy.ts
hsy822: I misunderstood it as needing to run automatically on the backend when actually it was about removing IPFS. Removal completed.

libs/remix-ui/run-tab/src/lib/actions/deploy.ts
hsy822: Updated accordingly.

apps/contract-verification/src/app/ContractVerificationPluginClient.ts
```diff
@@ -62,8 +64,156 @@ export class ContractVerificationPluginClient extends PluginClient {
...
+    const verifierSettings = chainSettings.verifiers[providerName]
+    const verifier = getVerifier(providerName, verifierSettings)
+
+    try {
+      if (validConfiguration(chainSettings, providerName)) {
+
+        await this.call('terminal', 'log', { type: 'log', value: `Verifying with ${providerName}...` })
+
+        if (providerName === 'Etherscan' || providerName === 'Routescan' || providerName === 'Blockscout') {
+          await new Promise(resolve => setTimeout(resolve, 10000))
```
yann300, line 172: why do you need to wait here?

hsy822: There is a delay between the contract deployment and the time when the block explorer recognizes the deployed information. If the verification script runs too soon, the explorer may not yet have indexed the deployed contract, resulting in an “Unable to locate ContractCode” error.

apps/contract-verification/src/app/ContractVerificationPluginClient.ts
```diff
@@ -62,8 +64,156 @@ export class ContractVerificationPluginClient extends PluginClient {
...
+        abiEncodedConstructorArgs: constructorArgs,
+        date: new Date().toISOString(),
+        receipts: []
+      }
+
+      const compilerAbstract: CompilerAbstract = compilationResult
+      const chainSettings = mergeChainSettingsWithDefaults(chainId, userSettings)
+
+      if (validConfiguration(chainSettings, 'Sourcify')) {
+        await this._verifyWithProvider('Sourcify', submittedContract, compilerAbstract, chainId, chainSettings)
```
yann300: do you need to `await` here? these calls are all independent, it would make the verification much faster.

hsy822: This was done intentionally to ensure that the logs are output sequentially and to avoid potential race conditions when creating the `receipts` array. If performance is a priority, would it be better to use `await Promise.all` to process them concurrently?

yann300: > would it be better to use await Promise.all to process them concurrently?

yes that makes sense.

hsy822: updated.

PR #6411, opened 2025-09-24
Override matomo window object

ioedeveloper:


apps/remix-ide/src/app/components/preload.tsx
```diff
@@ -8,7 +8,12 @@ import { localStorageFS } from '../files/filesystems/localStorage'
 import { fileSystemUtility, migrationTestData } from '../files/filesystems/fileSystemUtility'
 import './styles/preload.css'
 import isElectron from 'is-electron'
+import { Matomo } from '../plugins/matomo'
```
yann300, line 11: wouldn't that slow down the loading? did you notice some latency?

ioedeveloper: No, it does not slow it down.

apps/remix-ide/src/app/components/preload.tsx
```diff
@@ -8,7 +8,12 @@ import { localStorageFS } from '../files/filesystems/localStorage'
...
+
 const _paq = (window._paq = window._paq || [])
+const matomo = new Matomo(_paq)
+
+window._paq = matomo
```
bunsenstraat, line 16: Redefining window._paq here may be tricky, we don't actually know if loader.js has completed its task async loading the tracker code into _paq, and if the window._paq init object has been properly handled by the matomo initializing script. the point of matomo async loading is prevent any delays that would be caused by sync loading the script. so at this point the _paq may just be an array of the init we set in loader, it may not be the tracker at all @yann300 @ioedeveloper

bunsenstraat (changes requested):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

PR #6414, opened 2025-09-26
Render videos in md

yann300:
Allow rendering Videos tag using markdown.


joeizang:
I added some css to enable and fix the aspect ratio. The logic is driven by this below
`56.25% → 16:9 ratio (since 9 ÷ 16 = 0.5625 = 56.25%)`. So the paddingBottom helps 
to achieve this. We could also achieve a 4:3 aspect ratio using the same concept which
means padding would be 70%

joeizang:
The code LGTM

PR #6423, opened 2025-10-01
Trackerfix

bunsenstraat:
Introducing the new MatomoManager class to address these issues
- track custom dimension of tracking modes anonymous and cookie
- effectively switch between modes without issues ( like cookies reappearing on load )
- proper init of the tracker when the user comes in at a blank state and needs to choose a tracker preference
- events are now queued up and sent out AFTER the user made a choice of the mode, not before
- disableBrowserFeatureDetection makes anonymous mode more anonymous
- all tracking calls run via the matomo plugin or the new trackercontext that wraps react
- the MatomoManager exposes a lot of methods to debug it and also to do all kinds of stuff we need
- MatomoManager exposes events we can use to display stuff 
- proper matomo API definitions in remix-api for plugin calls
- settings tab is simplified now with plugin calls to switch mode
- no more _push confusion with tracker settings
- removed _paq usage. no more direct usage of window._paq 
- loader.js is removed, we don't need it

TODO:
update electron use
write e2e


apps/circuit-compiler/src/app/services/circomPluginClient.ts
```diff
@@ -22,11 +23,23 @@ export class CircomPluginClient extends PluginClient {
   private lastCompiledFile: string = ''
   private compiler: typeof compilerV215 & typeof compilerV216 & typeof compilerV217 & typeof compilerV218
   public _paq = {
-    push: (args) => {
-      this.call('matomo' as any, 'track', args)
+    push: (args: any[]) => {
```
ioedeveloper: This function "push: (args...)" is not needed in the circom plugin, since _paq.push has been replaced with trackCircuitEvent throughout the class.

apps/learneth/src/redux/models/remixide.ts
```diff
@@ -46,11 +47,28 @@ const Model: ModelType = {
         },
       });
 
+      // Type-safe Matomo tracking helper
+      const trackLearnethEvent = (event: ReturnType<typeof LearnethEvents[keyof typeof LearnethEvents]>) => {
+        trackMatomoEvent(remixClient, event);
+      };
+
+      // Legacy _paq compatibility layer for existing learneth tracking calls
```
ioedeveloper: We do not need legacy support for _paq calls since we are replacing all of it.

apps/remix-ide/src/app.ts
```diff
@@ -217,36 +230,23 @@ class AppComponent {
     this.workspace = pluginLoader.get()
     if (pluginLoader.current === 'queryParams') {
       this.workspace.map((workspace) => {
-        _paq.push(['trackEvent', 'App', 'queryParams-activated', workspace])
+        this.track(AppEvents.queryParamsActivated(workspace))
```
ioedeveloper: I don't think we need event function calls (AppEvents, HomeTabEvent, etc) for enforcing strict typing for matomo tracking.

bunsenstraat: but functions take parameters right 

so witout it  ```trackMatomoEvent?.(HomeTabEvents.switchTo(lang))``` 

would become ```trackMatomoEvent?.({ 
  ...HomeTabEvents.switchTo, 
  name: lang 
})``` which seems impractical to me

and other events have two parameters, name and value.

so in this case we are safe 

```export const HomeTabEvents = {
  ...
  switchTo: (name?: string, value?: string | number): HomeTabEvent => ({
    category: 'hometab',
    action: 'switchTo',
    name,
    value,
    isClick: true 
  })
} as const;```


```export interface HomeTabEvent extends MatomoEventBase {
  category: 'hometab';
  action: 
    | 'header'
    | 'filesSection'
    | 'scamAlert'
    | 'switchTo'
    | 'titleCard'
    | 'recentWorkspacesCard'
    | 'featuredPluginsToggle'
    | 'featuredPluginsActionClick'
    | 'updatesActionClick'
    | 'homeGetStarted'
    | 'startLearnEthTutorial'
    | 'featuredSection';
}```

apps/learneth/src/redux/models/remixide.ts
bunsenstraat: OK! replaced it all

apps/circuit-compiler/src/app/services/circomPluginClient.ts
bunsenstraat: Yes, I replaced it now

PR #6424, opened 2025-10-01
mcp: deploy and interact

yann300:


STetsing (changes requested):
Call contract is not steady
- A transaction is susses full from terminal pov, but **undefined** in src code 
image
- ABI is not passed sometimes to the tool as parameter. Either infer the ABI from tool args or make sure the compilation result was passed as request payload.

STetsing (approved):

PR #6462, opened 2025-10-16
fix(verification): Align automatic verifier list with plugin

### LINKED ISSUE #6461: Add contract verifier list
When deploying a contract with the automatic verification feature, the list of verifiers does not match the list used in the Contract Verification Plugin. We should use the list from the Contract Verification Plugin to ensure consistency between the two.

hsy822:
This PR resolves the inconsistency where the list of verifiers used during automatic deployment did not match the list available in the Contract Verification Plugin.

To ensure consistency, the automatic verification process now uses the same list of verifiers as the plugin.

fixes #6461


Aniket-Engg (commented):
that old issue to show red cross for routescan even on successful verification came back

Aniket-Engg (approved):
Functionality wise looks fine

PR #6475, opened 2025-10-18
reload login

bunsenstraat:
GH login was not loaded on reloaded so now users need to click on login again each time. this bug was introduced last release. test now covers it.


Aniket-Engg (approved):
Changeset looks fine. As login with github doesn't work with netlify preview, it is difficult to check functionality

PR #6483, opened 2025-10-22
Dynamically update deploy button text based on verification checkbox state

hsy822:
Update the deploy button text dynamically when “Verify to explorers” is checked, the button now shows “Deploy & Verify” instead of “Deploy”.


Aniket-Engg (commented):
UI is a bit messed up

Screenshot 2025-10-22 at 2 23 47 PM

PR #6486, opened 2025-10-22
fix tracking events

bunsenstraat:


libs/remix-ui/run-tab/src/lib/actions/deploy.ts
```diff
@@ -176,7 +176,7 @@ export const createInstance = async (
     plugin.compilersArtefacts.addResolvedContract(addressToString(address), data)
 
     if (isVerifyChecked) {
-      trackMatomoEvent(plugin, { category: 'udapp', action: 'DeployAndPublish', name: plugin.REACT_API.networkName, isClick: true })
+      trackMatomoEvent(plugin, { category: 'udapp', action: 'DeployAndVerify', name: plugin.REACT_API.networkName, isClick: true })
```
ioedeveloper: Action should be "DeployAndPublish"

bunsenstraat: really ok ok then

ioedeveloper (approved):

PR #6492, opened 2025-10-24
patch verification

bunsenstraat:
fix race conditions 

this has to with this button reacting to compilation events

Screenshot 2025-10-27 at 08 56 59

in Desktop it would not happen because the listeners were only set when the plugin emits activation events but this event got lost if the react App component wasn't rendered yet. so having the app check if the plugin is already activated fixes that. 

So test by looking at the button reacting to compilation.


Aniket-Engg (approved):
change set looks fine

PR #6495, opened 2025-10-26
Learneth UI improvement

yann300:
- when a tutorial is marked as finished, label will be displayed.
- improvement MD rendering.
- create and switch to the "learneth workspace" instead of spamming the current workspace.


ryestew:
When I finish a tutorial I get the `finished` badge - but at one of the standard widths, the formatting does not look good:
image
Maybe some kind of checkmark icon would do the trick instead of spelling out `Finished`

PR #6504, opened 2025-10-30
Implement priority sorting logic for LearnEth tutorials

### LINKED ISSUE #6501: Ordering Tutorials in LearnEth
**Is your feature request related to a problem? Please describe.**
Currently, tutorials are sorted by values in the .yml file - first on the `level` and then alphabetically by the `name`. We need a better way of sorting the tutorials so that specific tutorials are ordered with a higher priority and the rest can be sorted alphabetically. The filters should be visible by default. 

**Describe the solution you'd like**
The filter list should be on by default not closed as in this image:

Image

So it should look like this:

Image

We need a new property in each tutorial's .yml file that specifies the **priority**. 
The first sorting will be on the `level`property - as it currently is.  The for the display order of tutorials with the same `level` we need a new property called `priority`. The `priority` property should be able to handle a 3 digit number so that the if a new tutorial is added it could easily after or before an existing one. 

If a tutorial doesn't have a priority, it goes at the end ordered alphabetically by the `name` property in the .yml file.
If 2 tutorials have the same priority, between them they are then ordered alphabetically by the `name` property in the .yml file.

**Describe alternatives you've considered**
Another approach is https://ethereum.org/developers/tutorials/
But Remix tutorials should come first in each `level`

The other issues is that we only have a column in LearnEth - so the ethereum.org approach won't work.  But we also need to make the tag system more robust.

hsy822:
fixes #6501

This PR implements the new sorting logic for LearnEth tutorials as requested and sets the filter panel to be open by default.

**Key Changes**
* **New Sorting Logic (`Home/index.tsx`):**
    1.  By `Level` (asc)
    2.  By `priority` (asc, lower numbers first)
    3.  By `name` (alphabetical)
    * Items without a `priority` are sorted last, alphabetically by name.
* **Filter Panel Default State (`Home/index.tsx`):**
    * Changed the default state of `showFilters` to `true`.
* **Mock Data for Testing (`workshop.ts`):**
    * Since the remote `.yml` files are not yet updated with the `priority` field, I have added mock data in `workshop.ts` to allow for immediate testing.
    * The original `loadRepo` function is commented out, and a temporary `loadRepo` is used to inject this mock data.
    * Once the real `.yml` files are updated and deployed, `workshop.ts` can be easily reverted (uncomment original function, delete mock data).

@ryestew Please review the new sorting logic and the default filter state.

### DIFF
--- a/apps/learneth/src/pages/Home/index.tsx
+++ b/apps/learneth/src/pages/Home/index.tsx
@@ -103,7 +103,7 @@ function HomePage(): JSX.Element {
           levelText,
           name: entity.name || '',
           subtitle: tags.join(', '),
-          preview: mdToPlain(entity.description?.content || '', 280),
+          preview: mdToPlain(entity.metadata?.data?.summary || '', 280),
           stepsLen: entity.steps?.length,
           duration: entity.metadata?.data?.duration,
           tags,


apps/learneth/src/redux/models/workshop.ts
```diff
@@ -41,7 +147,7 @@ const Model: ModelType = {
     *loadRepo({ payload }, { put, select }) {
```
yann300, line 41: That should be rollbacked.

hsy822: That’s right. This was written temporarily to use mock data, so please just check the priority logic. I’ll roll it back later.

ryestew:
I've update the remix-workshops so that each .yml file has a priority key.  So the dummy data can be removed here.

ryestew:
I am not sure its in the scope of this PR, but in the description of each tutorial, the title of the tutorial is prefixing the description and it should be removed.  See:
image

hsy822:
> I am not sure its in the scope of this PR, but in the description of each tutorial, the title of the tutorial is prefixing the description and it should be removed. See: image

@ryestew The description was originally read from the README file, but it has been changed to be fetched from the summary field in the .yml file.

apps/learneth/src/redux/models/workshop.ts
hsy822: The update has been completed.

PR #6517, opened 2025-11-07
hide topbar and pinned pannel for browser wallet mode

bunsenstraat:
Screenshot 2025-11-07 at 10 25 17

When running as browser wallet connector the app should not have a top bar and pinned panel

test with /?#activate=udapp,desktopClient&desktopClientPort=49588

### DIFF
--- a/apps/remix-ide/src/app/providers/foundry-provider.tsx
+++ b/apps/remix-ide/src/app/providers/foundry-provider.tsx
@@ -26,7 +26,7 @@ export class FoundryProvider extends AbstractProvider {
           <b>curl -L https://foundry.paradigm.xyz | bash</b>
         </div>
         <div className="p-1 ps-3">
-          <b>anvil</b>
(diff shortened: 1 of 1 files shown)


libs/remix-ui/app/src/lib/remix-app/remix-app.tsx
```diff
@@ -124,14 +124,22 @@ const RemixApp = (props: IRemixAppUi) => {
     props.app.localeModule.events.on('localeChanged', (nextLocale) => {
       setLocale(nextLocale)
     })
+    
 
-    props.app.pinnedPanel.events.on('pinnedPlugin', (profile, isClosed) => {
-      if (!isClosed) setHidePinnedPanel(false)
-    })
+    if (!props.app.desktopClientMode) {
+      props.app.pinnedPanel.events.on('pinnedPlugin', () => {
```
Aniket-Engg: event listener for this event is added twice

PR #6520, opened 2025-11-07
EVM Map Patch update

### LINKED ISSUE #6518: Celo chain is compatible with Prague but doesn't let me deploy it.
**Describe the bug**
I want to deploy to Celo mainnet, but I can't because evmmap.ts needs to get updated.

Image

**To Reproduce**
1. Compile with EVM prague
2. try to deploy to Celo mainnet
3. Click on deploy and get the warning

joeizang:
fixes #6518


ryestew:
This update has fixed the problem for me.  

The logic should be that deployment to all chains are allowed to use the default VM - (now at Prague).  If a chain contacts us, we can limit that that chain's EVM maximum fork - but it will be up to them to get back to us to tell us that they have made an update.  Although they will likely forget to contact us again to bring them up to date. 

for a little history: evmmap.tx was originally done for Celo Mainnet which was not up to the latest EVM.  So we limited it.  They have been up to date for some time now.

yann300:
why did you comment out everything?

ryestew:
@joeizang - can you respond to Yann?  The reason is because when a chain is listed it limits that chain to the specified EVM.  But we need to remove all the limitations and only add them when the chain contacts us. Probably for merging the commented lines can be removed - unless there is a reason to keep them commented out.  Also Mainnet could be removed from all of the EVMs because it will be always up to date with the latest (by definition).

PR #6534, opened 2025-11-10
Fix CLI framework integ && add mcp && update audio endpoint behavoir

yann300:
Clicking compile from the udapp plugin, will initiate a compilation in foundry.
to test it: 
 - run desktop.
 - switch to a foundry project.
 - click on compile.
 - the terminal should display messages from foundry.

Other updates:
 - improve how remix listen on compilation result from hardhat/foundry (e.g only emit compilationFinished for the current file.
 - Solidity plugin UI: selecting compile with hardhat/foundry will no longer trigger a Remix compilation.
 - Fix solidity scan with desktop.
 - Add mcp endpoint for foundry/hardhat.
 - When transcribing, append the text in the chat box.

### DIFF
--- a/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
+++ b/libs/remix-ui/solidity-compiler/src/lib/compiler-container.tsx
@@ -73,7 +73,7 @@ export const CompilerContainer = (props: CompilerContainerProps) => {
   const [hhCompilation, sethhCompilation] = useState(false)
   const [truffleCompilation, setTruffleCompilation] = useState(false)
   const [compilerContainer, dispatch] = useReducer(compilerReducer, compilerInitialState)
-  
+
   useEffect(() => {
     api.getAppParameter('hardhat-compilation').then((result) => {
       if (result) {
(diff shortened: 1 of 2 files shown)


apps/remixdesktop/src/plugins/hardhatPlugin.ts
```diff
@@ -51,170 +47,168 @@ class HardhatPluginClient extends ElectronBasePluginRemixdClient {
...
-          this.watcher.on('add', async () => await this.triggerProcessArtifact())
-          this.watcher.on('unlink', async () => await this.triggerProcessArtifact())
-          // process the artifact on activation
-          this.processArtifact()
-        } catch (e) {
-          console.log('listenOnHardhatCompilation', e)
+    }    
+      
+    async sync() {
+      console.log('syncing Hardhet with Remix...')
```
Aniket-Engg, line 143: typo in Hardhet

Aniket-Engg (commented):
- Even compile with foundry is not checked, editor compile button compiles with foundry and doesn't create artifacts folder
- Same for compiling using compile button, when checkbox is not checked, it shows Remix compilation result immediately and initiates foundry compilation which take 15-20 seconds, may be we can show some info to user for the meantime
- Running SolidityScan for a contract compiled for Remix shows this to me:
Screenshot 2025-12-16 at 8 07 56 PM

- Toaster is very quick on clicking compile with Foundry checkbox. Add a time there

Aniket-Engg (approved):
This works almost fine. Discussed improvments and workspace witching fix can come in other PR

PR #6548, opened 2025-11-13
Update EVM Map

### LINKED ISSUE #6546: The evmmap.ts file needs updating
**Describe the bug**
The file evmmap.ts is not updated correctly.  I cannot deploy to Sepolia without it switching the EVM to Petersburg.  I had tested this on Celo and Arbitrum but not Sepolia.   

Either there should not be any entries in this file - so everything is permitted to go to the latest EVM or some other variation of that nothing - so that every chain uses the latest EVM.  

**To Reproduce**
1. Deploy to Sepolia and you'll get the warning.

joeizang:
fixes #6546

### DIFF
--- a/apps/remix-ide/src/blockchain/blockchain.tsx
+++ b/apps/remix-ide/src/blockchain/blockchain.tsx
@@ -25,7 +25,6 @@ const profile = {
   displayName: 'Blockchain',
   description: 'Blockchain - Logic',
   methods: ['dumpState', 'getCode', 'getTransactionReceipt', 'addProvider', 'removeProvider', 'getCurrentFork', 'isSmartAccount', 'getAccounts', 'web3VM', 'web3', 'getProvider', 'getCurrentProvider', 'getCurrentNetworkStatus', 'getCurrentNetworkCurrency', 'getAllProviders', 'getPinnedProviders', 'changeExecutionContext', 'getProviderObject', 'runTx', 'getBalanceInEther', 'getCurrentProvider', 'deployContractAndLibraries', 'runOrCallContractMethod'],
-
   version: packageJson.version
 }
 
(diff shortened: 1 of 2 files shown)


libs/remix-ui/run-tab/src/lib/actions/evmmap.ts
```diff
@@ -117,7 +116,7 @@ export const evmMap: Map<HardFork, { chainId: ChainInfo[], minCompilerVersion: s
     chainId: [
       { id: 1, name: "Ethereum Mainnet" },
```
ryestew: This line stops it from mainnet working with Prague.

libs/remix-ui/run-tab/src/lib/actions/evmmap.ts
```diff
@@ -78,7 +77,7 @@ export const evmMap: Map<HardFork, { chainId: ChainInfo[], minCompilerVersion: s
...
```
ryestew: This line stops mainnet working with Prague.  Ethereum Mainnnet should not be in this document - because it is always up to the latest.

ryestew (changes requested):
Goerli's max EVM - which should be Cancun.  Of course its about to be depricated.  I haven't tested Goerli.  But I don't think you need to put Sepolia nor Ethereum mainnet anywhere in this file.

libs/remix-ui/run-tab/src/lib/actions/evmmap.ts
joeizang: ok I will make the changes

ryestew (approved):
Its working for me - so I approved it. But should Goerli only be on listed in the petersburg section and not in all of the EVM sections? Isn't the highest version a chain can deploy to the only essential entry - the rest are not important - right?  Or am I misunderstanding the logic of evmmap.ts?

joeizang:
> Its working for me - so I approved it. But should Goerli only be on listed in the petersburg section and not in all of the EVM sections? Isn't the highest version a chain can deploy to the only essential entry - the rest are not important - right? Or am I misunderstanding the logic of evmmap.ts?

I thought we said to remove everything except goerli and so I left goerli to the last known supported evm version which is cancun I believe. But if we should make changes then I am all for it!

PR #6553, opened 2025-11-13
fix updating provider

yann300:
to reproduce: 
- switch to metamask.
- wait for it to be loaded.
- in metamask switch another network.
- udapp shows "can't detect network"


ryestew:
This fixes the problem for me.

PR #6618, opened 2025-12-10
Minor fixes in editor and workspace modal

Aniket-Engg:
- FIx showing of gas estimates for .ts files in editor
- Fixed minor UI things in workspace modal

### DIFF
--- a/apps/remix-ide-e2e/src/tests/url.test.ts
+++ b/apps/remix-ide-e2e/src/tests/url.test.ts
@@ -352,26 +352,24 @@ module.exports = {
       })
   },
 
-  // 'Should load using compiler from link passed in remix URL #group3': function (browser: NightwatchBrowser) {
-  //   browser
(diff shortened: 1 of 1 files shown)


apps/remix-ide/src/app/plugins/parser/services/code-parser-gas-service.ts
```diff
@@ -40,6 +40,10 @@ export default class CodeParserGasService {
       return
     }
     this.plugin.currentFile = await this.plugin.call('fileManager', 'file')
+    if (!this.plugin.currentFile || !this.plugin.currentFile.endsWith('.sol')) {
```
yann300, line 43: this will fail if currentFile is null

Aniket-Engg: this method is about showing gas estimates. for no current file, why should we show them?

yann300: I mean if `this.plugin.currentFile` is null the execution will fail, but ok I guess it will never come to that point as you have `!this.plugin.currentFile`.

PR #6633, opened 2025-12-15
fixes settings plugin scroll

### LINKED ISSUE #6550: Settings tab doesn't scroll all the way to the bottom
**Describe the bug**
In the Settings tab, the scroll doesn't go all the way to the bottom of the panel especially when the terminal is fairly high.
The scroll only goes to visible height of the panel.  The only way to get the full scroll is to make the terminal quite small.

And the left side of the Settings panel has no scroll - so the images below don't show the Connected Services choice on the left - even though I keep making the terminal smaller.

Image

so I drag the terminal down a bit

Image

and finally more

Image

So to see Connected Services you need to drag the terminal way down.  It needs a scroll bar or the h4 text should be smaller.

Aniket-Engg:
fixes #6550


Aniket-Engg:
> inline styling should only be in cases when bootstrap classes are not available for what needs to be done. Please consider revisiting some of the styling so we are have bootstrap classes. Packaging the inline styles into css classes would be acceptable to an extent.

updated

PR #6637, opened 2025-12-16
remove HTML title prop

### LINKED ISSUE #6631: title still visible
Title on hover were discontinued long time ago in support of tooltip. Somehow they are still visible in Solidity Compiler plugin

Image

Aniket-Engg:
fixes #6631 

Added tooltip instead of title in mainnet modal and AI prompt


Aniket-Engg:
> The code changes look good to me but a small question is this: Some of the title props removed have not been replaced by customtooltips, are we satisfied with this?

If a title was added for checkbox with label or button with label, tooltip is as of now ignored for those in this PR

PR #6647, opened 2025-12-18
minimal tool selection based on lexical matching

STetsing:
Tools sent to the LLM are context based

### DIFF
--- a/apps/remix-ide/webpack.config.js
+++ b/apps/remix-ide/webpack.config.js
@@ -43,7 +43,7 @@ class EmitSoljsonPlugin {
             })
             if (RawSource) {
               // Match previous public path: assets/js/soljson.js
-              // compilation.emitAsset('assets/js/soljson.js', new RawSource(data))
+              compilation.emitAsset('assets/js/soljson.js', new RawSource(data))
             }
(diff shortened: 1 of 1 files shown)


scutuatua-crypto (approved):

libs/remix-ai-core/src/services/simpleToolSelector.ts
```diff
@@ -0,0 +1,264 @@
...
+  selectTools(
+    allTools: IMCPTool[],
+    userPrompt: string,
+    maxTools: number = 15
+  ): IMCPTool[] {
+    console.log('[SimpleToolSelector] Using keyword-based selection', allTools)
+    return this.selectToolsWithKeywords(allTools, userPrompt, maxTools)
+  }
+
+  private selectToolsWithKeywords(
```
yann300, line 78: that would probably deserves some unittest. Also what shall we do with newly added tools?

STetsing: newly added tools (all tools) will be subject to this algo.

yann300 (approved):

PR #6673, opened 2025-12-28
Enable manual saving

yann300:
Enable manual saving. Still need to check:
- what are the implication with `fileManager.saveCurrentFile()`.
- how that will affect the `'contentChanged'` event.

To test it, go in the settings and enable "manual saving".


apps/remix-ide/src/app/editor/editor.js
```diff
@@ -652,6 +666,7 @@ export default class Editor extends Plugin {
    * @param {string} path
    */
   discard (path) {
+    console.trace()
```
Aniket-Engg: remove

apps/remix-ide/src/app/editor/editor.js
```diff
@@ -417,9 +432,11 @@ export default class Editor extends Plugin {
   async handleTypeScriptDependenciesOf(path, content, readFile, exists) {
     const isJsOrTs = path.endsWith('.js') || path.endsWith('.jsx') || path.endsWith('.ts') || path.endsWith('.tsx')
     
+    /*
```
Aniket-Engg, line 435: can this commented code be removed?

PR #6696, opened 2026-01-07
Adding model names

STetsing:
- Adding model names
- Adding user logged in restrictions. Only mistral small is available for non logged in users

Users who are logged in have access to more models additionally to the default `mistral small free`
Non logged in users only have access to `mistral small free` and `ollama`

### DIFF
--- a/apps/remix-ide/webpack.config.js
+++ b/apps/remix-ide/webpack.config.js
@@ -43,7 +43,7 @@ class EmitSoljsonPlugin {
             })
             if (RawSource) {
               // Match previous public path: assets/js/soljson.js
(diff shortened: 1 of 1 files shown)


apps/remix-ide/webpack.config.js
```diff
@@ -43,7 +43,7 @@ class EmitSoljsonPlugin {
             })
             if (RawSource) {
               // Match previous public path: assets/js/soljson.js
-              compilation.emitAsset('assets/js/soljson.js', new RawSource(data))
+              // compilation.emitAsset('assets/js/soljson.js', new RawSource(data))
```
yann300: you would need to uncomment that.

PR #6699, opened 2026-01-07
Add OZ ERC20Permit Patch

### LINKED ISSUE #6458: Add OZ ERC20Permit dependency
When using the OZ wizard, the ERC20Permit is by default added.  This is option is not available at all in the Remix Template Explorer, but it should be.

joeizang:
fixes #6458


ryestew:
The ERC20 Permit import should be associated with the a checkbox - under the Features section.  See how it works in [wizard.openzepplin.com](https://wizard.openzeppelin.com/).  So it is easy to erase.

PR #6716, opened 2026-01-14
Improved Script runner UI

Aniket-Engg:
- Show log when script runner is loading
- Show log for file execution once script runner is loaded
- Show error log for script runner failure
- Set cursor to pointer for run script dropdown
- Updated UI to select script configuration

### DIFF
--- a/libs/remix-ui/tabs/src/lib/components/DropdownMenu.tsx
+++ b/libs/remix-ui/tabs/src/lib/components/DropdownMenu.tsx
@@ -86,7 +86,7 @@ const DropdownMenu: React.FC<DropdownMenuProps> = ({
               }}
               data-id={item.dataId}
               role="menuitem"
-              style={{cursor: "pointer"}}
+              style={{ cursor: "pointer" }}
             >
               {item.icon && <span className="custom-dropdown-item-icon">{item.icon}</span>}
               <span>{item.label}</span>


apps/remix-ide/src/app/plugins/script-runner-bridge.tsx
```diff
@@ -275,15 +272,17 @@ export class ScriptRunnerBridgePlugin extends Plugin {
   }
 
   async execute(script: string, filePath: string) {
-    this.call('terminal', 'log', { value: `running ${filePath} ...`, type: 'info' })
     if (!this.scriptRunnerProfileName || !this.engine.isRegistered(`${this.scriptRunnerProfileName}${this.activeConfig.name}`)) {
-      console.error('Script runner not loaded')
+      console.log('Script runner not loaded already, loading it...')
+      this.call('terminal', 'log', { value: `Loading ScriptRunner...`, type: 'log' })
```
yann300: should be `Loading runtime...`

apps/remix-ide/src/app/plugins/script-runner-bridge.tsx
```diff
@@ -292,7 +291,7 @@ export class ScriptRunnerBridgePlugin extends Plugin {
       await this.call(`${this.scriptRunnerProfileName}${this.activeConfig.name}`, 'execute', transformedScript, filePath)
 
     } catch (e) {
-      console.error('Error executing script', e)
+      console.error('Error in executing script: ', e)
```
joeizang: I think same here also. It should be "Error while executing script:"

Aniket-Engg: updated

PR #6740, opened 2026-01-21
fix opening AI assitant

Aniket-Engg:
- FIx click on `Build on AI` to open AI assistant
- Fixed opening right side panel on icon click


Aniket-Engg:
> This works now after dismissing default loaded AI panel. However, when reactivating using the AI icon, I still do not get hover state over the icon, and when reactivated, the AI icon does not show that the panel is loaded on right side. image

this should have been fixed, please review again

PR #6775, opened 2026-02-03
New Debugging layout

Aniket-Engg:


yann300:
- I think the design impl need some improvement, but that looks so far.
- What is the difference between the contract calls and transaction tab?

yann300:
- Also when one click on an execution trace entry, it should jump the execution there.

Aniket-Engg:
> * Also when one click on an execution trace entry, it should jump the execution there.

done

Aniket-Engg:
> What is the difference between the contract calls and transaction tab?

I think contract calls are listing tx under the deployed contracts, Transaction tab is listing each tx individually.

libs/remix-ui/debugger-ui/src/lib/debugger-ui.tsx
```diff
@@ -201,8 +255,94 @@ export const DebuggerUI = (props: DebuggerUIProps) => {
...
+    if (debuggerInstance && debuggerInstance.debugger && debuggerInstance.debugger.callTree) {
+      // Store the callTree instance for later use
+      setCallTreeInstance(debuggerInstance.debugger.callTree)
+
+      debuggerInstance.debugger.callTree.event.register('callTreeReady', () => {
+        try {
+          // Get the root scope with low-level scopes merged
+          const nojumpScopes = debuggerInstance.debugger.callTree.getScopesAsNestedJSON('nojump')
+          // Also get all scopes to access the original scope tree
+          const allScopes = debuggerInstance.debugger.callTree.getScopesAsNestedJSON('all')
```
yann300, line 259: I don't thinkk you need to allScopes tree. is that really needed?

Aniket-Engg: I think it is used only to find more appropriate entry step

PR #6789, opened 2026-02-09
Debugger improvement part 3

yann300:


libs/remix-core-plugin/src/lib/helpers/fetch-etherscan.ts
```diff
@@ -94,13 +95,30 @@ export const fetchContractFromEtherscan = async (plugin, endpoint: string | Netw
       }
     }
   }
-  let runs = 0
-  try {
-    runs = parseInt(data.result[0].Runs)
-  } catch (e) { }
+
+  /*
```
Aniket-Engg, line 99: remove commented code if not required anymore

libs/remix-core-plugin/src/lib/compiler-fetch-and-compile.ts
```diff
@@ -160,13 +160,14 @@ export class FetchAndCompile extends Plugin {
 
     targetPath = `${targetPath}/${network.id}/${contractAddress}`
     let data
+    /*
```
Aniket-Engg, line 163: here too

libs/remix-core-plugin/src/lib/helpers/fetch-etherscan.ts
yann300: I would like to keep that one, just as a reminder

libs/remix-core-plugin/src/lib/compiler-fetch-and-compile.ts
yann300: same here

PR #6815, opened 2026-02-17
fixed AI test

STetsing:


apps/remix-ide-e2e/src/tests/mcp_file_permissions.test.ts
```diff
@@ -253,16 +253,14 @@ const tests = {
     browser
       .refresh()
       .waitForElementVisible('*[data-id="remixIdeSidePanel"]', 10000)
-      .pause(1000)
-      // Clear config
+      .waitForElementVisible('*[data-id="treeViewLitreeViewItemremix.config.json"]', 5000)
       .execute(function () {
         localStorage.removeItem('remix.config.json');
         (window as any).getRemixAIPlugin.call('fileManager', 'remove', 'remix.config.json');
```
bunsenstraat: these are async too. if you use it a lot create a helper.

STetsing: no need to, it is only used in that file

PR #6835, opened 2026-02-24
add Slither mcp and further update.

yann300:
- Fix some issues with MCP file operation.
- Impl hasUnacceptedChanges to make sure AI isn't overwritting it.
- Add internal slither MCP: for the moment, the contract needs to be flattened before sending it to the backend.
- Fix deploy_contract MCP.


libs/remix-ai-core/src/remix-mcp-server/handlers/FileManagementHandler.ts
```diff
@@ -23,7 +23,18 @@ import { Plugin } from '@remixproject/engine';
  */
 export class FileReadHandler extends BaseToolHandler {
   name = 'file_read';
-  description = 'Read contents of a file';
+  description = `Read contents of a file
+  Input Interface:
```
STetsing: @yann300 rm the input

PR #6836, opened 2026-02-24
Weighted tools

STetsing:
Introducing weighted tools on user chat history


libs/remix-ai-core/src/inferencers/mcp/mcpInferencer.ts
```diff
@@ -833,6 +872,41 @@ ${toolsList}`,
...
+          content: [{
+            type: 'text',
+            text: 'Error: get_tool_schema requires a tool_name parameter (string)'
+          }],
+          isError: true
+        };
+      }
+
+      const allTools = await this.getAvailableToolsForLLM()
+      const tool = allTools.find(t => t.name === toolName);
```
yann300, line 888: I think toolName might not be defined at this stage. It is only defined in the `if` line 877.

PR #6869, opened 2026-03-03
Update hometab and open default file when loading a workspace.

yann300:
- show a first time user card && display only 1 card instead of 4 (randomly selected).
- open a default file when creating a workspace and switching to it.

### DIFF
--- a/libs/remix-ui/home-tab/src/lib/components/homeTabUpdates.tsx
+++ b/libs/remix-ui/home-tab/src/lib/components/homeTabUpdates.tsx
@@ -66,6 +66,7 @@ function HomeTabUpdates({ plugin }: HomeTabUpdatesProps) {
           setShowFirstTime(true)
           setSelectedUpdate(null)
           setIsLoading(false)
+          return
         } else {
           // Still include as option for returning users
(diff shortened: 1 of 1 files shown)


STetsing:
@yann300 I still have no file selected on new workspace.

libs/remix-ui/home-tab/src/lib/components/homeTabUpdates.tsx
```diff
@@ -48,10 +48,51 @@ function HomeTabUpdates({ plugin }: HomeTabUpdatesProps) {
       try {
         setIsLoading(true)
         const response = await axios.get(HOME_TAB_NEW_UPDATES)
-        setPluginList(response.data)
+        const updates = response.data
+
+        // Check if this is the first time visiting
+        const hasVisitedHomeBefore = localStorage.getItem('remix-home-visited')
+        const isFirstTime = !hasVisitedHomeBefore
```
STetsing, line 55: This does not trigger the home tab to show first time user features.
too many storage flags for the same purpose
- remixaiassistant_firstload_flag
- remix-home-visited
- remix_ai_first_time_user

PR #6899, opened 2026-03-11
fixing UI issues and adding copy btn

STetsing:
fixing UI issues and adding copy btn


yann300:
added in https://github.com/remix-project-org/remix-project/pull/6888

PR #6900, opened 2026-03-11
Fix Broken UI when Remix AI is Maximized

joeizang:
Remix AI breaks when maximized and a user toggles a sidePanel Plugin on the left. This pr fixes this by moving the AI Chat History to the right of the IDE when the AI Assistant plugin is maximized.

It also address some issues:

| # | File | Change | Eliminates |
|---|------|--------|------------|
| 1 | `PluginViewWrapper.tsx` + `remix-ai-assistant.tsx` | `getInitialState()` lazy initializer | Empty first frame |
| 2 | `remix-ai-assistant.tsx` | `_initializing` flag to batch renders | Intermediate re-renders during storage init |
| 3 | `remix-ui-remix-ai-assistant.tsx` | Synchronous theme fallback | CSS flash from null `data-theme` |
| 4 | `useModelAccess.ts` | Pre-seed model defaults | Empty model list on first render |

### DIFF
--- a/apps/remix-ide-e2e/src/commands/clickInstance.ts
+++ b/apps/remix-ide-e2e/src/commands/clickInstance.ts
@@ -8,8 +8,8 @@ class ClickInstance extends EventEmitter {
     this.api.waitForElementPresent({
       locateStrategy: 'css selector',
       selector,
-      timeout: 60000
-    }).waitForElementContainsText(selector, '', 60000).scrollAndClick(selector).perform(() => { this.emit('complete') })
+      timeout: 80000
+    }).waitForElementContainsText(selector, '', 80000).scrollAndClick(selector).perform(() => { this.emit('complete') })
     return this
   }
 }


apps/remix-ide/src/app/panels/layout.ts
```diff
@@ -94,8 +94,15 @@ export class Layout extends Plugin {
     })
     this.on('sidePanel', 'focusChanged', async (name) => {
       const current = await this.call('sidePanel', 'currentFocus')
-      if (this.enhanced[current]) {
+      const isMaxed = await this.call('rightSidePanel', 'isRightSidePanelMaximized')
+      if (current === 'dgit') {
```
yann300: why do you need to take care of dgit here?

joeizang: because when enhanced, dgit is too wide and affects remix ai when it is centered and maximized. Maybe I can make it for all enhanced plugins not just dgit?

PR #6904, opened 2026-03-12
feat: QuickDapp share panel & OG tags

hsy822:
- Inject OG meta tags (og:title, og:description, twitter:card) on IPFS deploy
- Add share panel after ENS registration with copy link and X post


libs/remix-ui/quick-dapp-v2/src/lib/components/DeployPanel/index.tsx
```diff
@@ -3,6 +3,7 @@ import React, { useContext, useState, useRef, useEffect } from 'react';
 import { Form, Button, Alert, Card, Collapse, Spinner } from 'react-bootstrap';
```
affiliatemarketcommissions-hue, line 3:
```suggestion
import { Form, Button, Alert, Card, Collapse, Spinner } from 'react_app_bootstrap';
```

PR #6916, opened 2026-03-16
Fix udapp

yann300:
- Fix using the current provider when creating a smart account.
- clear recorder when switching and creating a workspace.
- Add Unix date converter tool.


STetsing (approved):

yann300:
pushed a little update to that PR that we'll need to rollback soon.

PR #6922, opened 2026-03-17
UI & CSS Bug fixes

joeizang:
- fix a css bug that causes all the action button on the right of the topbar to crush each other
- fix hometab scroll missing even when its apparent that content can be scrolled into.
- Add new label to New chat button when ai assistant is maximized

| Finding | Root cause | Solution implemented |
| --- | --- | --- |
| Left side-panel dragbar appeared too far from the panel edge | The dragbar used guessed placement and width math instead of the panel’s actual rendered edge | Reworked the side dragbar to measure panel/container geometry with `getBoundingClientRect()` and place the handle on the true edge |
| Dragbar placement drifted after side-panel enhance/maximize/reset flows | The dragbar was not fully synchronized with layout-driven width changes and CSS transitions | Added explicit resync on width transitions and post-transition settling so the dragbar follows the animated panel edge |
| Enhance behavior for the left panel did not always update the dragbar correctly | `layout.ts` emitted `enhancesidepanel` without an explicit width coefficient, so the UI layer had to guess | Updated `layout.ts` to emit enhance events with an explicit coeff and added helpers to resolve enhanced panel sizing consistently |
| Left and right dragbars could react to the wrong panel sizing state | `remix-app` stored one shared `coeff` for both left and right panel dragbars | Split sizing state into `leftPanelCoeff` and `rightPanelCoeff`, and wired each dragbar to only its own side’s events |
| Side dragbar resize math was unreliable | The `react-draggable` stop handler was using the wrong callback argument shape and mixing coordinate spaces | Fixed drag callbacks to use actual drag data and compute widths from container-relative geometry |
| Right dragbar could misalign in layout variants | It previously depended on viewport-based math rather than the actual right panel boundary inside the app layout | Changed right-panel width and dragbar placement to derive from measured layout bounds instead of `window.innerWidth` assumptions |
| Terminal dragbar could miss the terminal edge or jump to the top of the viewport after show/hide or maximize/restore | The dragbar could sync during an unstable terminal layout state, keep a stale `y=0` position, and remain mounted while the terminal was maximized | Added terminal layout revision tracking in `main-panel`, resynced the dragbar after layout settling with multi-step scheduling, ignored hidden/minimized/maximized states during measurement, clamped the dragbar position to container bounds, and unmounted the dragbar while the terminal is maximized |
| Dragbar position could go stale after show/hide or other DOM changes | Sync relied too heavily on delayed timing and not enough on real DOM/layout signals | Added `ResizeObserver`, `MutationObserver`, and immediate layout sync hooks so dragbars follow real panel changes |
| The visible drag handle looked visually off-edge even when close | The handle thickness started from the boundary rather than being centered on it | Centered both drag handles on the measured edge and normalized the hit area to `6px` |
| Regression coverage was missing for geometry correctness | Existing tests mostly checked visibility, not panel-edge alignment | Added `apps/remix-ide-e2e/src/tests/dragbar_alignment.test.ts` to assert left-panel, enhanced-panel, right-panel, and terminal dragbar alignment using `getBoundingClientRect()` comparisons |

### DIFF
--- a/apps/remix-ide-e2e/src/tests/file_explorer_context_menu.test.ts
+++ b/apps/remix-ide-e2e/src/tests/file_explorer_context_menu.test.ts
@@ -42,6 +42,7 @@ module.exports = {
       .waitForElementVisible('*[data-id$="fileExplorerTreeItemInput"]')
       .sendKeys('*[data-id$="fileExplorerTreeItemInput"]', browser.Keys.CONTROL + 'v')
       .sendKeys('*[data-id$="fileExplorerTreeItemInput"]', browser.Keys.ENTER)
+      .pause(5000)
       .waitForElementVisible('*[data-id="treeViewLitreeViewItemtests/README.txt"]', 7000)
   },
   'Should copy file name and paste in another folder that has the same filename with new file button and it will contain a new file #group1 ': function (browser: NightwatchBrowser) {


joeizang:
image

For point 2, I added some space to the bottom. I hope this is good?

ryestew:
My comments from before:
- [x] width for login buttons is fixed
- [x] space at the bottom is fine

joeizang:
> Actually I am still having problems with the terminal dragbar. To reproduce:
> 
> 1. compile and deploy a contract - to automatically open up the terminal
> 2. maximize the AI chat so it covers the screen
> 3. minimize the AI chat
> 4. try to find the terminal dragbar

This has now been fixed

PR #6923, opened 2026-03-17
More fixes for udapp

ioedeveloper:


ryestew:
I made a custom fork of Base, there is no indication that this is successful. (We spoke about this already)
I assume it is fine, but then when I deploy Storage to this forked chain, I get the error, 
```
creation of Storage errored: Error occurred: EIP-1559 not enabled on Common.
EIP-1559 not enabled on Common
```
I am not sure this is important or not.

yann300:
> I made a custom fork of Base, there is no indication that this is successful. (We spoke about this already) I assume it is fine, but then when I deploy Storage to this forked chain, I get the error,
> 
> ```
> creation of Storage errored: Error occurred: EIP-1559 not enabled on Common.
> EIP-1559 not enabled on Common
> ```
> 
> I am not sure this is important or not.

This means the EVM version you selected is not the right one for that network. cna you create an issue please?

ryestew (approved):

PR #6930, opened 2026-03-18
All models UI

STetsing:
User can now select non beta model. He will be prompted to join the beta program


ryestew:
This is great!  My only comment is outside the scope of what this PR is doing.  I'll leave the comment - but the PR's changes are working for me.  

The small adjustment I'd like is that
the text below the MCP checkbox (which may not be at all part of this PR) is too tight on the right side of the modal:
image

This rewrite will save space and you won't need to adjust the css:
"Adds relevant context from the connected MCP servers"

yann300 (approved):

PR #6941, opened 2026-03-19
UI Fixes for Udapp

joeizang:
Fix various light theme bug in Udapp


libs/remix-ui/run-tab-transactions/src/lib/widgets/TransactionsPortraitView.tsx
```diff
@@ -245,7 +245,7 @@ function TransactionsPortraitView() {
                   <FormattedMessage id="udapp.cancel" defaultMessage="Cancel" />
                 </button>
                 <button
-                  className={`btn btn-sm btn-danger ${themeQuality === 'dark' ? 'text-white' : 'text-dark'} flex-fill`}
+                  className={`btn btn-sm btn-danger ${themeQuality === 'dark' ? 'text-white' : 'text-white'} flex-fill`}
```
Aniket-Engg: if in both cases, it's `text-white` why to put a condition for that.

PR #6954, opened 2026-03-24
fix: production build stability and add independent beta deploy workflow

bunsenstraat:
## Summary

This PR fixes the production build crash on CI and adds an independently triggerable beta deploy workflow.

### CI: Add deploy_beta workflow (.circleci/config.yml)
- Added run_deploy_beta pipeline parameter (boolean, default false)
- Added deploy_beta workflow that can be triggered independently without running the full test suite

### Fix: Production build crash (apps/remix-ide/project.json)
- Disabled source maps for production builds (sourceMap: false). The source-map generation was causing the Node VM to run out of memory on CI medium+ containers (4GB RAM) while the heap limit was set to 8GB - the OS OOM killer would terminate the process before V8 could throw a JS heap error.
- Removed invalid remix-debug from implicitDependencies. This was a library name, not an app - the webpack config tried to stat dist/apps/remix-debug which does not exist (the debugger app builds to dist/apps/debugger). The error was caught but logged noise.

### Build: Add verbose flag (package.json)
- Added --verbose to build:production script for better CI debugging output.

### DIFF
--- a/apps/remix-ide/src/assets/list.json
+++ b/apps/remix-ide/src/assets/list.json
@@ -1089,9 +1089,57 @@
       "urls": [
         "dweb:/ipfs/QmfLTRkP13TiEUHvB8vQMjTNZrqhQwJKf8wXLZNFfoiRN8"
       ]
+    },
+    {
+      "path": "soljson-v0.8.32+commit.ebbd65e5.js",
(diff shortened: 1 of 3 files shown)


Aniket-Engg (approved):
change set looks good

PR #6975, opened 2026-03-26
Fixed wrong contract name extraction

ioedeveloper:
Fixes https://github.com/remix-project-org/remix-project/issues/6974
Fixes https://github.com/remix-project-org/remix-project/issues/6973


ryestew:
I don't think this fixes the problem.  I used the deployed preview and compiled a basic ERC 20 contract and I still get the strange names like Bytes, Strings, Panic etc:
image

PR #6986, opened 2026-03-27
Remix Ai UI Updates

joeizang:
Updates ai plugin ui as discussed here https://discord.com/channels/725383452081389608/1072122681966739466/1486987988754763856 and here https://discord.com/channels/725383452081389608/1072122681966739466/1486995287263412225


ryestew:
1. The Mistral and the Audio buttons are split on 2 lines. 
image

 If we remove a word from both, they will fit on one line:
image

I think the word "prompt" is not necessary.  And I don't think we need the word "free" because at the moment everything is free.  But if the word "(free)" is used and only the word "prompt" from Audio is removed, the buttons will be on a single line.

The next 2 points are probably outside the scope of this PR:
2. We should fix the red stop button - the black square in the middle of the button is not centered.

3. The text input box for the prompts has no border and is the same color as the background - so it is slightly confusing that this is a text input box. @ninja-fire

ryestew:
image

The bottom padding should be reduced to be  the same size as the top padding.

joeizang:
> image
> The bottom padding should be reduced to be the same size as the top padding.

@ryestew fixed.

ryestew (approved):
It is working for me

PR #7110, opened 2026-04-21
Workspace Dropdown Submenu fix

joeizang:


libs/remix-ui/remix-ai-assistant/src/types/promptHistory.ts
```diff
@@ -0,0 +1,23 @@
+
+class HistoryCircularButffer<T> {
```
yann300, line 2: you should remove this I think.

PR #7136, opened 2026-04-24
tool selection

yann300:


libs/remix-ai-core/src/inferencers/deepagent/DeepAgentInferencer.ts
```diff
@@ -630,6 +709,109 @@ export class DeepAgentInferencer implements ICompletions, IGeneration {
     }
   }
 
+  /**
+   * Recreate agent with selected tools
+   */
+  private async recreateAgentWithTools(selectedTools: DynamicStructuredTool[]): Promise<void> {
```
STetsing: why not directly call initialize the inferencer?

STetsing (approved):

PR #7147, opened 2026-04-28
fix: default deployed contract functions to expanded state

### LINKED ISSUE #7132: [UI Suggestion] Stop auto-collapsing deployed contract functions — It's killing the debugging workflow
The current UI logic in Remix 2.1 that automatically collapses deployed contract functions is extremely frustrating. Every time I re-deploy or switch between contracts, I have to manually click to expand the instance to access my functions.

For developers working on complex contracts (like my current project involving multi-set logic and dozens of getters/setters), this adds hundreds of unnecessary clicks per day.

Why it’s a bad user experience:

Muscle Memory broken: We are used to "Deploy -> Scroll -> Click". Now it's "Deploy -> Click to Expand -> Scroll -> Click".

Efficiency Loss: When debugging, I need to see all my available entry points at once. Hiding them behind a collapse menu serves no purpose for an active session.

Redundant UI: While it might look "cleaner" for a screenshot, it's functionally regressive for actual coding and rapid prototyping.

Suggested Solution:

Provide a setting in "General Settings" to "Disable Auto-Collapse for Deployed Contracts".

Or, at least remember the expansion state of a pinned contract so it doesn't reset on every re-deployment.

tenderdeve:
## Summary
- Change the default state of deployed contracts from collapsed to expanded so developers can immediately interact with functions after deploying
- Previously, every deploy or contract switch collapsed the function list, requiring manual clicks to expand

**Files changed:**
- `DeployedContractItem.tsx` — `isExpanded` default: `false` → `true`
- `universalDappUI.tsx` — `toggleExpander` default: `true` → `false` (inverted logic, same effect)
- `clickInstance.ts` (E2E) — Updated to check visibility before toggling, so tests work correctly with the new default

Fixes #7132

## Test plan
- [ ] Deploy a contract — functions should be visible immediately without clicking
- [ ] Deploy multiple contracts — all should start expanded
- [ ] Clicking the contract header should still toggle collapse/expand
- [ ] E2E tests pass

### DIFF
--- a/libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
+++ b/libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
@@ -23,7 +23,7 @@ export function UniversalDappUI(props: UdappProps) {
   const intl = useIntl()
   const { trackMatomoEvent: baseTrackEvent } = useContext(TrackingContext)
   const trackMatomoEvent = <T extends UdappEvent = UdappEvent>(event: T) => baseTrackEvent?.<T>(event)
-  const [toggleExpander, setToggleExpander] = useState<boolean>(false)
+  const [toggleExpander, setToggleExpander] = useState<boolean>(true)
   const [contractABI, setContractABI] = useState<FuncABI[]>(null)
   const [address, setAddress] = useState<string>('')
   const [expandPath, setExpandPath] = useState<string[]>([])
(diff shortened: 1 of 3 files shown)


libs/remix-ui/run-tab/src/lib/components/universalDappUI.tsx
```diff
@@ -24,7 +24,7 @@ export function UniversalDappUI(props: UdappProps) {
   const intl = useIntl()
   const { trackMatomoEvent: baseTrackEvent } = useContext(TrackingContext)
   const trackMatomoEvent = <T extends UdappEvent = UdappEvent>(event: T) => baseTrackEvent?.<T>(event)
-  const [toggleExpander, setToggleExpander] = useState<boolean>(true)
+  const [toggleExpander, setToggleExpander] = useState<boolean>(false)
```
ioedeveloper: No need to make changes here. This file is deprecated and should be removed.

ioedeveloper (changes requested):

tenderdeve:
Good catch, reverted , left that file untouched now.

Aniket-Engg (approved):

PR #7154, opened 2026-04-29
Desktoplogin

bunsenstraat:


apps/remix-ide/src/app/plugins/auth-plugin.tsx
```diff
@@ -513,8 +513,108 @@ export class AuthPlugin extends Plugin {
...
+    if (!state) return
+
+    try {
+      const userJson = JSON.stringify(user)
+      const userBase64 = btoa(userJson)
+        .replace(/\+/g, '-')
+        .replace(/\//g, '_')
+        .replace(/=+$/, '')
+
+      const callbackUrl = `remix://auth/sso-callback?state=${encodeURIComponent(state)}&access_token=${encodeURIComponent(accessToken)}&refresh_token=${encodeURIComponent(refreshToken)}&user=${encodeURIComponent(userBase64)}`
```
yann300, line 537: wouldn't that mean the remix protocol need to be available in all cases, which might not be certain?

yann300 (approved):

PR #7190, opened 2026-05-08
Code analysis popover in editor

yann300:
only for starter pack


libs/remix-ai-core/src/storage/interfaces.ts
```diff
@@ -18,7 +18,7 @@ export interface TodoItem {
 
 export interface ChatMessage {
   id: string
-  role: 'user' | 'assistant'
+  role: 'user' | 'assistant' | 'editor_code_analysis'
```
yann300, line 21: you should not change that.  why do you need it?

PR #7201, opened 2026-05-10
Fix for clearing unpinned contracts

grace-tisher:
This fix is for the Deploy & Run bug described in #7033. An alert appears if there are unpinned contracts in the deployed contracts list when attempting to switch to a different environment.


ryestew (changes requested):
You are opening a browser modal - not a Remix modal.  To keep the design standardized, please use the Remix modal - @joeizang do you have a suggestion about how to do that?

joeizang:
@grace-tisher Hi. please use the IDE provided alert via a plugin call. Thanks @ryestew for tagging me here

PR #7221, opened 2026-05-14
Langchain user api key

STetsing:
BYO API key


apps/remix-ide/src/app/plugins/remixAI/DeepAgentManager.ts
```diff
@@ -17,6 +17,84 @@ export class DeepAgentManager {
...
+    } catch (error) {
+      console.warn('[DeepAgentManager] Failed to check API keys permission:', error)
+      return false
+    }
+  }
+
+  private getSettingFromStorage(key: string): string | boolean {
+    try {
+      const storageKey = 'config-v0.8:.remix.config'
+      const configData = localStorage.getItem(storageKey)
```
yann300: I think it's better to use a plugin call fro this, not using localstorage directly.

apps/remix-ide/src/app/plugins/remixAI/ModelManager.ts
```diff
@@ -29,6 +31,88 @@ export class ModelManager {
     this.deps = deps
   }
 
+  private async canUseOwnApiKeys(): Promise<boolean> {
```
yann300: why there seems to be a duplication between ModelManager and DeepAgentManager

apps/remix-ide/src/app/plugins/remixAIPlugin.tsx
```diff
@@ -195,17 +206,45 @@ export class RemixAIPlugin extends Plugin {
       try {
         console.log('[RemixAI Plugin] Initializing DeepAgent with mcpInferencer:', !!this.mcpInferencer);
         console.log('[RemixAI Plugin] Using model for DeepAgent:', this.selectedModel.provider, this.selectedModelId);
+
+        // Read user API keys from settings
+        let userApiKeys: IUserApiKeyConfig | undefined
+        try {
+          const config = Registry.getInstance().get('config').api
+          const useOwnKeys = config.get('settings/deepagent-api-keys-config') || false
```
yann300: I think the way api keys are loaded needs to be refactored.

apps/remix-ide/src/app/plugins/remixAI/ModelManager.ts
STetsing: this comes from the refactoring

yann300 (approved):

PR #7294, opened 2026-06-02
ENS contract naming integration

hsy822:
## Summary
Replace the frontend-only ENS PoC with server-backed contract ENS registration.

Users can register `{label}.{project}.remixcontract.eth` for deployed contracts. Forward resolution is handled by the backend, so users do not pay L1 gas. Ownable contracts can optionally set reverse resolution on the deployment chain.

## Changes
- Rewrite `EnsNaming.tsx` to use server API: preflight → job → polling → optional reverse
- Add reverse retry/status handling and read-only primary ENS lookup
- Hide ENS menu in VM environments


NickSneo:
Hey @hsy822 , nice work!

Small thing I noticed: @enscribe/enscribe got added to package.json, but I don't see it imported anywhere in the PR. Forward resolution runs through the server API, which is fine, but the reverse resolution in EnsNaming.tsx is all written out manually (hardcoded registrar/resolver addresses, separate setNameForAddr ABIs for L1 and L2, manual owner() check, etc).

We're on the Enscribe team and happy to help plug the lib in. It should cover the same logic with a lot less code and let you drop those hardcoded ENS addresses and ABIs. Want me and @abhijeetbhagat to leave a few concrete suggestions below?

abhijeetbhagat:
@hsy822 you can use the enscribe library to set the reverse record like this:
```
import { nameContract } from "@enscribe/enscribe";
import { createWalletClient, http } from "viem";
import { sepolia } from "viem/chains";

// Create a wallet client
const walletClient = createWalletClient({
  chain: sepolia,
  transport: http(),
  account: "0x...", // your account address
});

// Name a contract - the library handles all network configuration internally
const result = await nameContract({
  name: "mycontract.myname.eth",
  contractAddress: "0x1234567890123456789012345678901234567890",
  walletClient,
  chainName: "sepolia", // Library determines contracts and network logic
  enableMetrics: true, // Optional: enable metrics logging
});
```

refer to the basic example [here](https://github.com/enscribexyz/enscribe-ts#basic-example)

hsy822:
> Hey @hsy822 , nice work!
> 
> Small thing I noticed: @enscribe/enscribe got added to package.json, but I don't see it imported anywhere in the PR. Forward resolution runs through the server API, which is fine, but the reverse resolution in EnsNaming.tsx is all written out manually (hardcoded registrar/resolver addresses, separate setNameForAddr ABIs for L1 and L2, manual owner() check, etc).
> 
> We're on the Enscribe team and happy to help plug the lib in. It should cover the same logic with a lot less code and let you drop those hardcoded ENS addresses and ABIs. Want me and @abhijeetbhagat to leave a few concrete suggestions below?

Thanks @NickSneo, appreciate the kind words!

You're right that the package is added but not wired in. The reason is architectural: our flow splits Forward (server-side, server pays gas) and Reverse (client-side, user signs via MetaMask). nameContract() runs both through the provided wallet, which is broader than what we need at the reverse step, forward is already complete by that point.

The manual code also gives us fine-grained control over step-by-step terminal logging, L1/L2 ABI branching, simulation, and custom error messages.

Two questions:

1. Does Enscribe expose a reverse-only API (separate from nameContract())? That would fit our architecture well.
2. Could we use utility functions like getContractAddresses() or isOwnable() independently?
If there's a way to use Enscribe for just the reverse path, we'd be happy to wire it in.

hsy822:
> @hsy822 you can use the enscribe library to set the reverse record like this:
> 
> ```
> import { nameContract } from "@enscribe/enscribe";
> import { createWalletClient, http } from "viem";
> import { sepolia } from "viem/chains";
> 
> // Create a wallet client
> const walletClient = createWalletClient({
>   chain: sepolia,
>   transport: http(),
>   account: "0x...", // your account address
> });
> 
> // Name a contract - the library handles all network configuration internally
> const result = await nameContract({
>   name: "mycontract.myname.eth",
>   contractAddress: "0x1234567890123456789012345678901234567890",
>   walletClient,
>   chainName: "sepolia", // Library determines contracts and network logic
>   enableMetrics: true, // Optional: enable metrics logging
> });
> ```
> 
> refer to the basic example [here](https://github.com/enscribexyz/enscribe-ts#basic-example)

Thanks @abhijeetbhagat for the concrete example!

The main issue is that nameContract() runs subname creation and forward resolution through the provided walletClient. In our flow, by the time the user sees "Set Reverse", forward is already done server-side. That makes nameContract() broader than the reverse-only action we need at this point in the flow.

Does Enscribe expose a standalone reverse function, or a way to skip the forward steps? That would fit our architecture well.

Either way, utilities like getContractAddresses() and isOwnable() look useful. they could replace our hardcoded ENS addresses.

NickSneo:
> Thanks @abhijeetbhagat for the concrete example!
> 
> The main issue is that nameContract() runs subname creation and forward resolution through the provided walletClient. In our flow, by the time the user sees "Set Reverse", forward is already done server-side. That makes nameContract() broader than the reverse-only action we need at this point in the flow.
> 
> Does Enscribe expose a standalone reverse function, or a way to skip the forward steps? That would fit our architecture well.
> 
> Either way, utilities like getContractAddresses() and isOwnable() look useful. they could replace our hardcoded ENS addresses.

Currently no, but we are making those changes now to expose reverse-only flow. Will push it soon

hsy822:
> > Thanks @abhijeetbhagat for the concrete example!
> > The main issue is that nameContract() runs subname creation and forward resolution through the provided walletClient. In our flow, by the time the user sees "Set Reverse", forward is already done server-side. That makes nameContract() broader than the reverse-only action we need at this point in the flow.
> > Does Enscribe expose a standalone reverse function, or a way to skip the forward steps? That would fit our architecture well.
> > Either way, utilities like getContractAddresses() and isOwnable() look useful. they could replace our hardcoded ENS addresses.
> 
> Currently no, but we are making those changes now to expose reverse-only flow. Will push it soon

Thanks for adding `setReverseName()` in v0.1.3. it matches our reverse-only flow nicely.

I tried integrating it in Remix, but hit a browser bundling issue: `naming.js` imports `randomUUID` from Node `crypto`, while Remix resolves `crypto` to `crypto-browserify`, which does not export `randomUUID`.

I also saw a local Node ESM import issue around `@noble/hashes` / `@scure/bip32` exports. Would you be open to making the package more browser-bundler friendly, starting with replacing Node `crypto.randomUUID` usage with `globalThis.crypto?.randomUUID()` plus a fallback?

Once that’s resolved, we should be able to use `setReverseName()` and remove most of our manual reverse logic.

NickSneo:
> Thanks for adding `setReverseName()` in v0.1.3. it matches our reverse-only flow nicely.
> 
> I tried integrating it in Remix, but hit a browser bundling issue: `naming.js` imports `randomUUID` from Node `crypto`, while Remix resolves `crypto` to `crypto-browserify`, which does not export `randomUUID`.
> 
> I also saw a local Node ESM import issue around `@noble/hashes` / `@scure/bip32` exports. Would you be open to making the package more browser-bundler friendly, starting with replacing Node `crypto.randomUUID` usage with `globalThis.crypto?.randomUUID()` plus a fallback?
> 
> Once that’s resolved, we should be able to use `setReverseName()` and remove most of our manual reverse logic.

Sure, let me fix those and release with an updated version

libs/remix-ui/run-tab-deployed-contracts/src/lib/components/EnsNaming.tsx
```diff
@@ -0,0 +1,1019 @@
+import React, { useState, useEffect, useRef, useContext, useCallback } from 'react'
+import { DeployedContractsAppContext } from '../contexts'
+import { DeployedContract } from '../types'
+import { endpointUrls } from '@remix-endpoints-helper'
+import { createWalletClient, createPublicClient, custom, type Hex } from 'viem'
+
+const PARENT_NAME = 'remixcontract.eth'
```
ioedeveloper: Can you separate these constant variable declarations and types from the React component? Export them from a separate file.

libs/remix-ui/run-tab/src/lib/ens-contract-names-client.ts
```diff
@@ -0,0 +1,106 @@
+/**
```
ioedeveloper: I think it would be nice to have an ENS plugin (without UI), that manages all ens operations (contract naming) and interfaces with the backend.

NickSneo:
@hsy822 Hey, we have released a new version - https://www.npmjs.com/package/@enscribe/enscribe/v/0.1.4
Plz let me know if any more changes are required from our side

hsy822:
> @hsy822 Hey, we have released a new version - https://www.npmjs.com/package/@enscribe/enscribe/v/0.1.4 Plz let me know if any more changes are required from our side

I checked `@enscribe/enscribe@0.1.4.` The previous browser bundling blocker appears to have been resolved, and `setReverseName()` seems to align well with Remix’s current flow, where forward resolution is handled by the backend and reverse resolution by the user’s wallet.

However, this PR already includes a structural change that extracts the ENS logic into a headless plugin. Adding the Enscribe integration at the same time would likely make the scope too large.

I’ll proceed with the current refactoring approach in this PR, and integrate only `setReverseName()` into the reverse-only path in a separate follow-up PR. 

Thanks.

libs/remix-ui/run-tab-deployed-contracts/src/lib/components/EnsNaming.tsx
hsy822: Done. Shared constants/types/helpers were moved out of the React component into `ens-contract-names.ts`.

libs/remix-ui/run-tab/src/lib/ens-contract-names-client.ts
hsy822: Done. Added a headless `ensContractNames` plugin and moved ENS backend/wallet operations there. The UI now uses plugin calls/events.

ioedeveloper (approved):

PR #7314, opened 2026-06-03
tool UI usage names, fixed model selection after stop request

STetsing:
- tool UI usage names with correct paths
- fixed model selection after stop request]


yann300 (approved):

r2776454703-max (changes requested):
JxxgJ

PR #7340, opened 2026-06-09
Fix AI Model Responsiveness

joeizang:
Fixes UI bug where menu that shows the different model options in Remix AI Assistant will not show if the viewport is small. This fix should cover 13" screens and larger.


ryestew (approved):
works for me!

PR #7346, opened 2026-06-10
Hometab design

ninja-fire:


yann300:
you can remove "connect to local file system", and maybe replace it with a link to remix desktop

yann300 (approved):

PR #7362, opened 2026-06-12
Fix inline dapp beta test bugs

ioedeveloper:


joeizang (approved):
I think it looks fine

yann300 (approved):

PR #7399, opened 2026-06-16
UI topbar

ninja-fire:


ryestew (changes requested):
In the off state, the cloud should be white and not gray so that it is easier to see - at least it should be as bright as the current gray of the workspace name in the Workspace select box. But I think both should be brightened. Also, the cloud icon should go on the left of the Workspace select box not on the right.

ryestew (approved):

PR #7418, opened 2026-06-18
Query Thegraph

ioedeveloper:


apps/remix-ide/src/remixAppManager.ts
```diff
@@ -475,6 +477,17 @@ export class RemixAppManager extends BaseRemixAppManager {
       sticky: true,
       group: 7,
     })
+    this.call('filePanel', 'registerContextMenuItem', {
```
yann300, line 480: I see that menu item for all file types.

PR #7487, opened 2026-07-01
fix templates

yann300:
use wizard template instead of harcoded ones.
At the moment, compile a ERC721 fails because of that.


libs/remix-ui/template-explorer-modal/src/components/contract-wizard.tsx
```diff
@@ -81,9 +89,6 @@ export function ContractWizard () {
       dispatch({ type: ContractWizardAction.UPDATE_ERC20_PERMIT, payload: { ...strategy.contractOptions, permit: true } })
     }
   }
-  function updateTokenName(tokenName: string) {
```
Aniket-Engg, line 84: is this change needed?

Aniket-Engg (approved):

PR #7508, opened 2026-07-07
Verify circom groth16 proofs on zkverify.

ioedeveloper:


hsy822 (commented):
Tested the main groth16 zkVerify flow locally and it looks good to me.

libs/remix-ui/quick-dapp-v2/src/lib/utils/zkverify-runtime-config.ts
```diff
@@ -0,0 +1,177 @@
...
+    primeValue: zkCircuit.primeValue,
+    signalInputs: zkCircuit.signalInputs || [],
+    zkArtifacts: useRootPaths
+      ? {
+        wasmPath: 'circuit.wasm',
+        zkeyPath: 'circuit.zkey',
+        vkeyPath: 'verification_key.json'
+      }
+      : {
+        wasmPath: zkCircuit.zkArtifacts?.wasmPath || 'zk/circuit.wasm',
```
hsy822: I tested this flow and proof generation works after deploying the DApp to IPFS. However, it fails in QuickDApp Preview with `Failed to fetch WASM: 404`.

The Preview iframe resolves `zk/circuit.wasm` against the Remix web origin, but this file only exists in the Remix workspace and is not exposed as an HTTP resource. Should Preview load the artifacts through the Remix file manager/asset bridge? Otherwise, if Preview proof generation is intentionally unsupported, we should disable it and show an explicit message.

Image

libs/remix-ui/quick-dapp-v2/src/lib/components/DeployPanel/index.tsx
```diff
@@ -226,21 +228,67 @@ function DeployPanel(): JSX.Element {
...
+        const wasmPath = zkArtifacts?.wasmPath || `${defaultFolder}/circuit.wasm`;
+        const zkeyPath = zkArtifacts?.zkeyPath || `${defaultFolder}/circuit.zkey`;
+        const vkeyPath = zkArtifacts?.vkeyPath || `${defaultFolder}/verification_key.json`;
+
+        console.log('[IPFS Deploy] ZK artifacts paths:', { wasmPath, zkeyPath, vkeyPath, mode: activeDapp?.mode });
+
+        // Read and append circuit.wasm (at root level for IPFS compatibility)
+        try {
+          const wasmData = await plugin.call('fileManager', 'readFile', wasmPath, { encoding: null });
+          const wasmContent = wasmData instanceof Uint8Array ? new Uint8Array(wasmData) : new TextEncoder().encode(wasmData as string);
```
hsy822, line 417: When `zk/circuit.wasm` is open in the editor during IPFS deployment, the deployed DApp fails with the following error:

`WebAssembly.compile(): expected 114 bytes, fell off end @+154`

I suspect that the currently open WASM file is being read as the editor's text content and then re-encoded, resulting in binary corruption.

Image

hsy822 (approved):

PR #7530, opened 2026-07-14
fix(editor): avoid false busy state in code helper

### LINKED ISSUE #7494: Code Helper: RemixAI Assistant says it is busy but it is not
**Describe the bug**
Sometimes the RemixAI Assistant says it is busy - but it not.  This could be happening after I have previously hit the stop button in the AI Assistant.  Although - after refreshing Remix, its still says the RemixAI Assistant is busy.  But clicking the button "open in RemixAI Assistant" will work and the assistant will run the analysis.

Image

zzz100868:
The Code Helper treated any analysis request that took longer than five seconds as proof that RemixAI was busy. The request kept running in the background, which is why opening the same analysis in RemixAI could still work.

This removes the synthetic busy timeout and lets `basic_prompt` resolve or reject normally. Actual service failures continue through the existing error path.

Validation:
- bundled `tooltipPopOver.tsx` with esbuild to verify TSX syntax
- `git diff --check`

Closes #7494

### DIFF
--- a/libs/remix-ui/editor/src/lib/tooltipPopOver.tsx
+++ b/libs/remix-ui/editor/src/lib/tooltipPopOver.tsx
@@ -491,30 +491,7 @@ Risk: critical=severe, high=dangerous, medium=warning, low=minor, info=tip, perf
 For "relatedDocs", ONLY use URLs from these trusted domains: ${trustedUrls}
 Use empty array if no relevant trusted docs.`
 
-        // Wrap API call with timeout to detect if AI is busy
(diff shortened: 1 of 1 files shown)


Aniket-Engg:
@zzz100868 Sometimes AI assistant takes longer to process existing requests and in between, code analysis popover will keep showing spinner without any idea what is happening. Fix for attached issue should be some thing else not completely removing the AI busy message.

Aniket-Engg:
closing as https://github.com/remix-project-org/remix-project/pull/7544 fixed the issue.

PR #7542, opened 2026-07-16
fix(doc-viewer): improve line spacing for wrapped headings

### LINKED ISSUE #7493: MD Viewer: h1 with long titles are not well formatted
**Describe the bug**
When the title of a md file has a long name its line spacing is too narrow and the lines run into each other

Image

**To Reproduce**
1. Open a .md file with a long name
2. view the result

pengqima:
## Summary

- apply an explicit line height to level-one Markdown headings
- prevent long wrapped titles from overlapping in the document viewer
- keep the change scoped to rendered Markdown `h1` elements

## Root cause

The Remix theme uses a responsive font size for `h1` elements while also setting a fixed 24px line height. In narrow document viewer panels, long headings wrap onto multiple lines and the fixed line height causes those lines to overlap.

Using the existing Bootstrap `lh-sm` utility gives wrapped headings a line height relative to their font size.

## Testing

- verified that Markdown `h1` elements receive the `lh-sm` class
- ran `git diff --check`

Fixes #7493

### DIFF
--- a/apps/doc-viewer/src/app/App.tsx
+++ b/apps/doc-viewer/src/app/App.tsx
@@ -21,7 +21,13 @@ export default function App() {
     <>
       <div className="bg-light p-5">
         <button className="btn btn-sm border mb-2" onClick={edit}>EDIT</button>
-        <ReactMarkdown children={contents} remarkPlugins={[remarkGfm]} />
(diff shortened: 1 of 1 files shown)


ryestew (approved):
works for mw!

PR #7544, opened 2026-07-16
handle request stopping to show correct code analysis popover

### LINKED ISSUE #7494: Code Helper: RemixAI Assistant says it is busy but it is not

Aniket-Engg:
fixes #7494


ryestew (changes requested):
I am not getting the code helper at all in this PR.

yann300:
the fix is working.

PR #7595, opened 2026-08-03
fix(remix-ai): forward pending sync operations to the cloud backend

grandpig:
## Description

Pending chat history sync operations were stored in `ChatHistoryStorageManager.syncQueue`, while the S3 backend's `push()` processed a separate `S3ChatHistoryBackend.syncQueue`.

Because the operations were never forwarded to the cloud backend, `push()` could process an empty queue and return a successful result without uploading the pending local changes. The manager would then clear its own queue, dropping those operations.

This change:

- adds an optional `queueSync()` method to `IChatHistoryBackend`;
- forwards operations queued by `ChatHistoryStorageManager` to the cloud backend;
- adds a regression test verifying that a locally created conversation is
  queued in the cloud backend before `push()` runs.

## Root cause

`ChatHistoryStorageManager` and `S3ChatHistoryBackend` maintained separate sync queues, but there was no forwarding path between them.

## Testing

Added a unit test covering the following flow:

1. Initialize the storage manager with local and fake cloud backends.
2. Create a local conversation.
3. Verify that the cloud backend receives a `conversation/create` operation.
4. Verify that the operation is available when `push()` is called.

### DIFF
--- a/libs/remix-ai-core/src/storage/interfaces.ts
+++ b/libs/remix-ai-core/src/storage/interfaces.ts
@@ -172,6 +172,7 @@ export interface IChatHistoryBackend {
 
   // Sync operations (for cloud backends)
   supportsSync(): boolean
+  queueSync?(operation: SyncOperation): void
   push?(): Promise<SyncResult>
   pull?(): Promise<SyncResult>
   getLastSyncTime?(): Promise<number | null>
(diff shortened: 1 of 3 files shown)


grandpig:
@ryestew Hi! You reviewed the Remix AI chat-history work in #7133. This PR fixes the cloud sync queue handoff in the same chat-history storage area and includes a regression test for pending operations reaching the cloud backend. Would you mind taking a look when you have time? Thanks!

grandpig:
@ryestew Could you please take a look when you have a chance? Thanks!

PR #7598, opened 2026-08-04
fix(plugin-manager): persist permission changes on confirmation

### LINKED ISSUE #7597: Permission settings close button keeps unconfirmed changes
**Describe the bug**

Permission changes made in the Plugin Manager permissions dialog are persisted immediately. Clicking **Cancel** restores the permissions captured when the dialog was opened, but clicking the header **X** closes the dialog without invoking the rollback callback.

As a result, changes that were not explicitly confirmed remain persisted when the dialog is closed using the X button.

Steps to reproduce

1. Grant plugin A permission to call plugin B.
2. Open Plugin Manager → Permissions.
3.Toggle or delete the permission.
4.Close the dialog using the top-right X button.
5.Open the permissions dialog again.
6.Observe that the change is still present.

**Expected behavior**
Closing the permissions dialog using the X button should have the same rollback behavior as clicking Cancel.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows, Linux or MacOS]
 - Browser: [e.g. chrome, firefox]
 - Version: [e.g. 22]

**Additional context**

daixiheguu:
## Summary

Fixes https://github.com/remix-project-org/remix-project/issues/7597

- make the permissions dialog's header close action follow the cancellation path
- restore the permission state captured when the dialog was opened
- keep the existing OK behavior unchanged
- add coverage for closing the dialog after toggling or deleting permissions

## Problem

Permission edits are persisted to localStorage immediately. The footer Cancel button invokes `cancelFn`, which restores the original permissions, but the header X button only hides the modal.

This causes unconfirmed changes to remain persisted when users close the dialog using the X button.

### DIFF
--- a/apps/remix-ide-e2e/src/tests/pluginManager.test.ts
+++ b/apps/remix-ide-e2e/src/tests/pluginManager.test.ts
@@ -94,6 +94,102 @@ module.exports = {
         const finalCount = parseInt(result.value as string)
         browser.assert.equal(finalCount, initialAllCount, `Count should return to initial count after clearing filters.`)
       })
+  },
+
+  'Should only persist permission changes after confirmation #group1': function (browser: NightwatchBrowser) {
+    const permissionsButton = '[data-id="pluginManagerPermissionsButton"]'
+    const permissionCheckbox = '[data-id="permission-checkbox-fileManager-writeFile-e2ePlugin"]'
+    const removePermissionButton = '[data-id="pluginManagerSettingsRemovePermission-fileManager-writeFile-fileManager"]'
+
+    browser
+      .execute(function () {
+        localStorage.setItem(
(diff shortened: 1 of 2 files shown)


Aniket-Engg:
@daixiheguu with which plugin you reproduce this?

daixiheguu:
> @daixiheguu with which plugin you reproduce this?

Thanks for asking. I did not reproduce this through a specific third-party plugin.

I reproduced it in Remix Plugin Manager -> Permissions using an existing `plugins/permissions` entry for an external plugin. I then toggled or deleted the permission and closed the dialog with Cancel/X. The issue is in the permissions dialog's draft state and persistence behavior, so the regression test seeds a representative permission record instead of depending on a plugin-specific flow.

PR #7718, opened 2026-09-14
Fix ai feedback

yann300:
When deploying a contract in a test/live net, it should:
- send a prompt to remixAI if the email is verified and nothing in the other case.
- show a prompt placeholder instead of the full prompt (see the displayText property).
- fix: shows the default credit treshold for that feature in the settings.


Aniket-Engg (commented):
Image

- Deployment prompt formatting can be improved.

- Clicking on SAVE button in AI feedback section first time shows `credentials updated` which is incorrect. Once value is updated in textbox, and SAVE button is clicked, no toaster is shown

Aniket-Engg (commented):
Deployment prompt formatting fixed, other comment should also be handled.

Aniket-Engg (approved):

